PR #13, opened 2023-04-17
Direct floppy driver

Mellvik:
The main component of this PR is the new direct floppy driver - first cut. Based on the code that has been lingering in ELKS and now TLVC since the early nineties, this rewrite adapts the code to the present day environment which entails lots of changes in many different places. As of this PR the driver works well, boots, and needs more testing:
- Tested only with 1.44 and 1.2M floppies
- Supports track buffering (not cylinder)
- Has lots of debug stuff compiled in still, thus it's not very fast yet (but has the notable difference from the BIOS based driver that I/O operations can be interrupted at any time).
- No floppy-to-floppy testing yet
- Will fail (eventually) if run with the BIOS hd driver, track buffering and XMS buffers (this is permanent, a direct hd driver is forthcoming)
- No IOCTLs at this time, meaning that the `sys` command doesn't work yet.
- The config define is CONFIG_BLK_DEV_FD
- Formatting is not tested and not supported.

While developing the driver, a number of bugs have been encountered and fixed, most notably 
- the timers in the scheduler (`del_timer, add_timer`) had a bug causing the system to crash if deleting a non-existent timer. The timer code has been fixed and simplified.
- The DMA code has been updated and works. However, the code is about to be retired and replaced by a simple routine (from minix) which will reduce the code size and add speed. We don't need all the flexibility of the old DMA code given that this is the only driver that is likely to ever need DMA.

The block floppies have major device # 2, moving the `ssd` major to 7. The device names are `/dev/f0` and `/dev/f1` to distinguish them from the BIOS floppy devices. The update also prepares for the upcoming direct hd driver and includes updates to some commands, such as `df`, `dd`, `sys`- the latter still not working pending a geometry IOCTL in the floppy driver.

### DIFF
--- a/image/Make.devices
+++ b/image/Make.devices
@@ -116,24 +116,35 @@ endif
 
 ##############################################################################
 # Direct floppy support
+ifeq ($(CONFIG_BLK_DEV_FD), y)
 	$(MKDEV) /dev/f0 b 2 0
 	$(MKDEV) /dev/f0-1440 b 2 28
 	$(MKDEV) /dev/f1 b 2 1
 	$(MKDEV) /dev/f1-1440 b 2 29
+endif
 
 ##############################################################################
 # Direct PATA / IDE disks. (Not working)
 # Limit to 4 primary partitions per disk (MBR).
 
-#	$(MKDEV) /dev/dhda b 5  0    # Currently.
+ifeq ($(CONFIG_BLK_DEV_HD), y)
+	$(MKDEV) /dev/dhda b 5  0    # Currently.
 	$(MKDEV) /dev/dhda b 5  0 4  # Currently.
-#	$MKSET   0 15 $(MKDEV) /dev/dhda b 5	# Currently.
-#	$(MKDEV) /dev/dhdb b 5 64    # Currently.
-#	$(MKDEV) /dev/dhdb b 5 64 4  # Currently.
-#	$MKSET  64 15 $(MKDEV) /dev/dhdb b 5	# Currently.
+	$MKSET   0 15 $(MKDEV) /dev/dhda b 5	# Currently.
+	$(MKDEV) /dev/dhdb b 5 64    # Currently.
+	$(MKDEV) /dev/dhdb b 5 64 4  # Currently.
+	$MKSET  64 15 $(MKDEV) /dev/dhdb b 5	# Currently.
+endif
 
 #	$MKSET   0 63 $(MKDEV) /dev/dhda b 3	# Ought to be.
 #	$MKSET  64 63 $(MKDEV) /dev/dhdb b 3	# Ought to be.
+
+##############################################################################
+# Direct floppy drives (experimental)
+ifeq ($(CONFIG_BLK_DEV_FD_NOTUSED), y)
+	$(MKDEV) /dev/dfd0 b 3 0
+	$(MKDEV) /dev/dfd1 b 3 64
+endif
 
 ##############################################################################
 # Virtual consoles, pseudo-tty slaves and serial ports.
--- a/tlvc/arch/i86/drivers/net/ne2k-asm.S
+++ b/tlvc/arch/i86/drivers/net/ne2k-asm.S
@@ -472,11 +472,11 @@ ne2k_pack_get:
 #if 0
 	// -------------------------------------------------------------
 	// Packet size check not required since the NIC will not
-	// accept such packets per our initialization. Note, in order to handle
-	// erroneous packets, rx_get (BOUNDARY) pointer must be updated
-	// to point to the next packet.
+	// accept such packets per our initialization. Note, when handling
+	// erroneous packets manually, the rx_get (BOUNDARY) pointer must subsequently
+	// be updated do point to the next packet.
 	// If oversized packets still occur, it's a driver problem (most likely
-	// reading the wrong buffer page).
+	// reading the wrong buffer page from the NIC).
 	// -------------------------------------------------------------
 	or      %cx,%cx		// zero length
 	jz      npg_err2
--- a/tlvccmd/ash/bltin/history.c
+++ b/tlvccmd/ash/bltin/history.c
@@ -1,0 +1,18 @@
+/*
+ * The history command.  Lists the current history buffer - which may be empty.
+ * to the standard output.
+ *
+ * Copyright (C) 2023 by Helge Skrivervik.  All rights reserved.
+ * This file is part of ash, which is distributed under the terms specified
+ * by the Ash General Public License.  See the file named LICENSE.
+ */
+
+
+#include "bltin.h"
+
+void linenoiseHistoryList(void);
+
+void histcmd(int argc, char **argv) {
+	linenoiseHistoryList();
+	return;
+}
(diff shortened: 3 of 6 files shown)


tlvc/arch/i86/drivers/block/directfd.c
```diff
@@ -316,128 +353,154 @@ static unsigned char seek_track = 0;
...
+ *
+ * DOR (Data Output Register) is /MOTD/MOTC/MOTB/MOTA/DMA/(RST)/DR1/DR0/
+ */
+static void floppy_on(int nr)
 {
-    unsigned char mask = 0x10 << nr;
+    unsigned char mask = 0x10 << nr;	/* motor on select */
 
+    if (debug) printk("flpON");
+    *fl_timeout = 0;	/* Reset BIOS motor timeout counter, neccessary on some machines */
```
ghaerr, line 453: The kernel IRQ 0 hardware timer interrupt calls a saved copy of the original BIOS' interrupt vector every 5 clock ticks. The BIOS usually uses this interrupt to decrement floppy motor timeouts, amongst other things.

It would be interesting to remove that BIOS callback entirely and see whether that 1) affects anything workings of the direct FD and HD drivers, thus removing another BIOS dependency, as well as 2) eliminate the need for poking a zero in the Bios Data Area to reset the motor timeout counter.

Also: given that the declaration of `fl_timeout` is `long __far *`, using `*fl_timeout = 0` zeros not only the byte location of the BIOS motor timeout at 0x440, but also three other BDA variables at 0x441-0x443!

Mellvik:
Thank you @ghaerr,
It took me quite some time to figure this one out. I’ve been looking for a way to stop the motor timeout counter at 0x440, no success. So I ended up with this band aid solution. How can that BIOS callback be removed?

Thanks, and yes I notices the size bug in the poke a few days ago. It was included in the PR posted earlier today. BTW - out of curiosity I compared this method with using the pokeb() call. Turns out the memory footprint is the same, so I kept this one.

—M

ghaerr:
> How can that BIOS callback be removed?

The code is in `arch/i86/kernel/irqtab.S`, find this code and comment it out to quickly test:
```
//
//      IRQ 0 (timer) has to go on to the bios for some systems
//
        decw    bios_call_cnt   // Will call bios int?
        jne     do_eoi
        movw    $5,bios_call_cnt
        pushf
        lcall   *org_irq0
        jmp     was_trap        // EOI already sent by bios int
```
After testing, a few more lines would also be removed for a final version. For BIOS FD/HD, it would want to remain in, and out for direct FD/HD. It sounds like TLVC is at the point where one should select only BIOS or direct, never both. I am guessing the only reason BIOS is available is for compatibility testing with machines that might not yet work with direct FD/HD.

>  I compared this method with using the pokeb() call. Turns out the memory footprint is the same, so I kept this one.

Interesting that the amount of code it takes to push three parameters and call a function is the same as loading a far pointer and dereferencing... and certainly the far pointer dereference is faster. I have found that having at least a function wrapper around nonstandard C extensions (e.g. __far) allows for superior portability, especially when a compiler changes ("Our compiler won't ever change", famous last words I know :)

Mellvik:
Thanks @ghaerr,
I certainly didn’t catch that one. Removing it does the trick.

It’s on my list to track down and remove/ifdef low level bios dependencies (e.g. the drive parameter table stuff), but I want to get the system stable first. That said, in this case keeping it made the system unstable, so the argument is not a good one. My priority list just changed.

The BIOS block IO stuff may eventually go away, but like you say it's useful to keep around for comparison for a while.

C extensions: Right - then there is the third option: «Better call @tkchia» :-)

Code size (poke): Your comment made me curious, it does indeed sound too good to be true. And it is, but just barely: I was 2 bytes off :-)
     a28:       30 c0                   xor    %al,%al
     a2a:       50                      push   %ax
     a2b:       b8 40 00                mov    $0x40,%ax
     a2e:       50                      push   %ax
     a2f:       50                      push   %ax
     a30:       e8 fe ff                call   a31 <floppy_on+0x19>
                        a31: R_386_PC16 pokeb

Moot now, but interesting anyway.

-M

ghaerr:
> the timers in the scheduler (del_timer, add_timer) had a bug causing the system to crash if deleting a non-existent timer. The timer code has been fixed and simplified.

I know the timer code simplification was later reverted, but was there a bug when deleting a non-existent timer? For some reason, I thought I saw a fix somewhere like this but now can't find it. IIRC it was a null pointer de-reference?

Mellvik:
You remember (almost) correctly. I had to look it up.
The fix was simply initializing the static pointer `next_timer`:
```
static struct timer_list *next_timer = NULL;
```
The (rather elusive) problem happened only if `del_timer` was called before ANY ` add_timer` call. Which is exactly what the direct floppy driver did/does.

ghaerr:
> The fix was simply initializing the static pointer next_timer:

Could you do me a big favor and remove the initialization `= NULL` and see whether the problem persists?
If so, a much bigger problem is at hand, all uninitialized kernel data variables (static or not) are placed in the `.bss` data section, and cleared to 0 at the time of kernel startup. If that isn't working, we have a much bigger problem.

Here's the code in crt0.s that clears .bss:
```
/*
! Setup.S already initialized DS and ES (but not SS)
! In addition, registers contain:
!   BX, Text size
!   DI  Far text size
!   SI, Data size
!   DX, BSS size
*/
        mov     %bx,_endtext
        mov     %di,_endftext
        mov     %si,_enddata
        add     %dx,%si
        mov     %si,_endbss

// Start cleaning BSS. Still using setup.S stack

        mov     _enddata,%di    // start of BSS
        mov     %dx,%cx         // CX = BSS size
        xor     %ax,%ax
        shr     $1,%cx
        cld
        rep
        stosw

// End cleaning BSS
```

The only thing the initialization does in the struct time_list code is move the *location* of the stored pointer to `.data` rather than `.bss`. That is, it's address will be different, but the initialized value the same.

In summary, the above fix shouldn't need to happen and needs more investigation. *All* uninitialized kernel variables are set to 0. If you can find another case where this isn't true, that will be proof of problem. Otherwise, all that initialization of a variable to 0 does is increase the size of /linux in its .data section on disk.

Mellvik:
I'll do that - but bear with me while I try to catch this new rabbit and gasp for air every once in a while. I have a buffer pollution problem - probably self-inflicted in the sense that there is a typo somewhere in all the changes I made (and imported) over the past few days. Of course it only happens when booting off of a (slow) floppy. `git diff` is my friend...

I remember the `del_timer()` issue well, I went as far as to print the initialized value to make sure I was on the right track. I got random data. If initialization works as you describe - which is likely otherwise we'd likely have all kinds of other problems - I had or have a memory leak or similar. Interesting times.

ghaerr:
> I had or have a memory leak or similar. 

I'll wager on memory overrun. A quick way to get real on this is to look at the system.map file in arch/i86/boot/system.map. That will show the variables in front and back of the one getting trashed, which usually points you in the right direction. Since initializing the variable stuck it in a different section, it doesn't get trashed... but the one that took its place could still be. Look at the map file locations for all the direct FD driver variables (both .data and .bss).

> probably self-inflicted in the sense that there is a typo somewhere in all the changes I made (and imported) over the past few days. Of course it only happens when booting off of a (slow) floppy. 

Or this could be the same original problem, just trashing different variables since the codebase changed...

PR #17, opened 2023-06-18
XMS support in direct HD/FD drivers

Mellvik:
A relatively minor update in terms of code, major in terms of functionality.

### DIFF
--- a/tlvc/arch/i86/drivers/block/directfd.c
+++ b/tlvc/arch/i86/drivers/block/directfd.c
@@ -553,7 +553,11 @@ static void setup_DMA(void)
     unsigned int count;
 
     DEBUG("setupDMA ");
+#ifdef CONFIG_FS_XMS_BUFFER
+    dma_addr = LAST_DMA_ADDR + 1;	/* force use of bounce buffer */
+#else
     dma_addr = _MK_LINADDR(CURRENT->rq_seg, CURRENT->rq_buffer);
+#endif
     count = BLOCK_SIZE;
     if (command == FD_FORMAT) {
 	dma_addr = _MK_LINADDR(kernel_ds, tmp_floppy_area);
@@ -567,7 +571,11 @@ static void setup_DMA(void)
     } else if (dma_addr >= LAST_DMA_ADDR) {		/* Not likely */
 	dma_addr = _MK_LINADDR(kernel_ds, tmp_floppy_area); /* use bounce buffer */
 	if (command == FD_WRITE) {
+#ifdef CONFIG_FS_XMS_BUFFER
+	    xms_fmemcpyw(tmp_floppy_area, kernel_ds, CURRENT->rq_buffer, CURRENT->rq_seg, BLOCK_SIZE/2);
+#else
 	    fmemcpyw(tmp_floppy_area, kernel_ds, CURRENT->rq_buffer, CURRENT->rq_seg, BLOCK_SIZE/2);
+#endif
 	}
     }
     DEBUG("%d/%lx;", count, dma_addr);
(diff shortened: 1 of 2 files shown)


tlvc/arch/i86/drivers/block/directfd.c
```diff
@@ -567,7 +571,11 @@ static void setup_DMA(void)
     } else if (dma_addr >= LAST_DMA_ADDR) {		/* Not likely */
 	dma_addr = _MK_LINADDR(kernel_ds, tmp_floppy_area); /* use bounce buffer */
 	if (command == FD_WRITE) {
+#ifdef CONFIG_FS_XMS_BUFFER
+	    xms_fmemcpyw(tmp_floppy_area, kernel_ds, CURRENT->rq_buffer, CURRENT->rq_seg, BLOCK_SIZE/2);
+#else
 	    fmemcpyw(tmp_floppy_area, kernel_ds, CURRENT->rq_buffer, CURRENT->rq_seg, BLOCK_SIZE/2);
+#endif
```
ghaerr, line 578: An ifdef for `CONFIG_FS_XMS_BUFFER` isn't required here (or below) to differentiate between copying to/from XMS (using `xms_fmemcpy`) and `fmemcpyw` (to/from main memory) in any block driver. It is all handled automatically within `xms_fmemcpyw`, which can tell the difference between an XMS and main memory address, and will call `fmemcpy` if needed.

This is because the all buffer segment descriptors (including `CURRENT->rq_seg`) are declared of type `ramdesc_t` (an unsigned long), rather than `seg_t` (an unsigned short). A `ramdesc_t` (ram descriptor) is a 32-bit physical memory address when the upper 16 bits are nonzero, and an 8086 segment descriptor when the upper bits are 0. This format allows for passing a single address for buffers without having to specify XMS linear addresses or segment values separately.

A caveat of this is that when CONFIG_FS_XMS_BUFFER is defined, one can't assume that the lower 16 bits of a `CURRENT->rq_seg` is an 8086 segment value and shift left by 4 to obtain a physical address. The `ramdesc_t` values are interpreted within `xms_fmemcpyw` and the appropriate routine called.

ghaerr:
> there is also the debug printks which need to be different depending on the size of ramdesc_t.

One way to do this without ifdef's is to cast the `ramdesc_t` value to its largest size:
```
printk("Seg: %08lx\n", (unsigned long)req->rq_seg);
```
which will then work whether CONFIG_FS_XMS_BUFFER is set or not.

Mellvik:
Hey, that’s a good one. I'll definitely use that.

BTW - in the xms_??? routines, wouldn’t 
if (sizeof(ramdesc_t) == 4) 
be even simpler (and most likely being removed but the compiler anyway)?

-M

ghaerr:
> in the xms_??? routines, wouldn’t
if (sizeof(ramdesc_t) == 4)
be even simpler (and most likely being removed but the compiler anyway)?

Ah - I'm glad you asked that since I noticed in some code some assumptions might have been possibly incorrectly made when CONFIG_FS_XMS_BUFFERS is enabled: 1) Just because the XMS config option is enabled doesn't mean that XMS is active in the kernel - if the machine isn't a 386 (or CONFIG_FS_XMS_INT15 also enabled), then `sizeof(ramdesc_t)` will be 4 but XMS will be OFF. Also, 2) One still can't assume, even when XMS is configured and running, that a `ramdesc_t` is a linear descriptor; it may also be just a real mode segment. In fact, when XMS is active, the xms_fmemcpy routines are always passed both an XMS address and a real mode segment address, and `address >> 16` is used to figure out which way to transfer - XMS to main, or main to XMS. The `ramdesc_t` address thus can either be an unreal-mode linear address for directly addressing memory or a real mode segment address, depending on whether XMS is active as well as whether the buffer is in XMS or L1 cache.

Mellvik:
Well, now I’m glad I asked too.
There is at least one instance in the floppy driver that may be affected by this. Not dramatic, it will select to use a local bounce buffer when it shouldn’t - but still. I’ll have a look at it.

thanks.

-M

PR #19, opened 2023-07-25
Direct driver updates, RAW disk driver

Mellvik:
Another mega patch with lots of updates, mostly related to the new direct FD and HD drivers.

Notably, a new raw/char driver for disk devices has been added, another milestone towards the initial goal for TLVC. Some technicalities about the raw driver implementation:
- The driver itself is mostly a stub, calling code elsewhere. Most of the job is done in the block driver and in `ll_rw_blk.c`.
- The assumption that we're always reading or writing in `BLOCK_SIZE` quantities has been removed, which had some interesting ripple effects through the system, some of which may still linger.
- The major device numbers have been changed so that the char and block devices for the same physical device have matching major device numbers. This assumption eliminates the need for a mapping table. Not a big deal but simpler, less code.
- In order to avoid big changes at the lower levels of the IO system and also to avoid conflicts between concurrent block and raw IO on the same device, raw requests use the same request queues as block IO. A regular buffer header is used to pass metadata back and forth, without the buffer proper (replaced by the buffer provided by the user program).
- The raw disk IO needs more testing, but seems to work reliably (see also comments in the code)
- A similar stub driver for floppy is forthcoming, and will add minimal code to the system. Also coming is a renaming of the disk devices, incidentally back to what it was in ELKS until recently: `hda` for (direct) disks, `bda` for BIOS disks. The floppy device names will be kept, `fd` for BIOS floppy, `df` for direct floppy.
- The old CONFIG_BLK_DEV_CHAR configuration option turned out to be required for the system to compile but had no real meaning. It has been repurposed to do the real thing. Turning CONFIG_BLK_DEV_CHAR off now works and remove only the raw/char driver support.

Other updates:
- The `directhd` driver didn't do well when exposed to 'real' old IDE drives and has been significantly updated, among other things with delay loops required to give the older drives time to change their status settings. The experience revived the ambition to make the driver interrupt driven, thus eliminating most of the delay loops. The experience with old IDE drives triggered a new chapter in the Developer Notes wiki.
- The direct FD/HD drivers now work with XMS buffers.
- A number of configuration/conditional compile options have been tested and fixed. All combinations that make sense now compile.
- Cleanups: Quite a bit of inactive (`#if 0` type) code has been removed, the code related to floppy formatting (untested, unused) has been `#ifdef`ed to reduce bloat.

There are still race condition bugs in the system - becoming visible when booting and running off of floppy disk.


tlvc/arch/i86/drivers/block/directhd.c
```diff
@@ -431,47 +467,71 @@ int INITPROC directhd_init(void)
 	 * of them even have more than 16.384 cylinders
 	 *
 	 * This is some sort of bugfix, we will use same method as real Linux -
-	 * work with disk geometry set in current translation mode rather than
-	 * using physical drive info. Physical info might correspond to logical
-	 * info for some drives (those which don't support/use LBA mode).
+	 * work with disk geometry set in current translation mode (54-56) if valid (53)
+	 * rather than physical drive info. Old drives have only physical, which may
+	 * be misleading: The copnmfigured BIOS drive type's CHS mapping may be different.
+	 * E.g the Compaq drive type 17 (conner 42MB) reports 806/4/26 while the BIOS
```
ghaerr, line 473: Are you saying here that the "BIOS configured value" (assume from user or computer manufacturer configuration) should always override what the drive itself might report (if indeed it reports anything)? If both work, what is the reason that drive and BIOS CHS may be different? 

It would seem then for TLVC that a CHS specification (likely from /bootopts?) would just have to match the way that TLVC calculates CHS from block numbers. (?)

In the past, I have only noticed possibly a cylinder number being 1 greater through the IDE probing than the BIOS-reported value.

ghaerr:
Wow, you have been busy!! It looks like you're having a lot of fun writing this deep in the middle of the kernel block, buffer, driver and raw device code :) 

The idea of a raw device driver combined with multi-sector I/O is pretty neat, as it could provide for a way to greatly speed up certain copy operations. I thought to ask a few questions to better understand your design goal for this, as well as to point out some design considerations that went into the synchronous block driver and buffer design.

If the raw device driver could perform multi-sector I/O directly into an application buffer, copies could get very fast, compared to using sync or async I/O through the block device system. Of course, the primary reason the block drivers don't do multi-sector I/O is that there's no guarantee the next file block is contiguous with the one being read (although it could be, that's a different story for another time). The block drivers do attempt a two-sector contiguous read (for a 1K block) or sometimes a full-track read (when track caching is enabled).

Then we have the problem of exactly what physical memory addresses a disk sector can be read or written from. For the block drivers, this is always either an L1 (kernel data segment) or L2 (possibly main or XMS) memory. XMS memory can't ever be direct, and a copy must occur. Some older PCs don't allow disk I/O around a 64K segment overlap, and some supposedly require I/O only into the lower 64K physical (those systems must be configured to use L1 buffers only and then perform I/O into the fixed DMASEG buffer only). For both the L1 and L2 buffers (actual data buffer locations, not headers), care has been taken to allocate them aligned to a 1K-boundary (for both kernel data segment and main memory). Since the size of I/O is always 1K and the buffers are 1K-aligned, the 64k DMA (or other) wrap issue is solved. (I'm not sure whether this issue is only for FD and DMA, or whether its also an HD issue as well). Of course, the BIOS is doing all the I/O, we're just passing the ES:BX buffer address which is aligned properly.

I noticed that the direct HD driver is using a local buffer. Hopefully that can be redesigned to go directly to an L1 or L2 buffer, unless XMS is running (lets worry about XMS a bit later for this discussion). If the FD or HD driver has to use a local buffer, then there won't be a max speed improvement for using raw devices, since internal buffer copies will have to be made.

In order to use raw devices at full speed, the application, after opening the raw device, must pass a buffer that is 1K aligned, and also know what the max multi-sector count is. Then it could execute an (async) read directly into its buffer. Nice and fast! This method would still require that the current (raw) file offset is sector-aligned (otherwise requiring mandatory buffering of intra-sector contents).

Thus, IMO, it would be interesting to see whether the new raw device mechanism could be designed for maximum throughput first, then add considerations for buffering later, as an option. (These buffer options would be required for non-aligned user buffer, non-sector-aligned file offset, XMS buffer address, and possibly others. They could possibly all be handled with a single slow-path method, frankly using the previous `ll_rw_blk` raw driver that uses system buffers for I/O and handles non-aligned file and buffer offsets).

I have some thoughts as to how this could all be accomplished outside the block device/buffer system, only using an updated I/O request header, not going through `ll_rw_blk` at all. Also, the track or cylinder cache system could be rewritten to run as just another multi-sector I/O request (using aligned buffers) and thus be removed from being driver-specific. Essentially all this would live below the current buffer system and above the block driver, at the nr_request level.

I am curious as to your design goals and thoughts on this, and can continue commenting if desired, thank you!

Mellvik:
Thanks @ghaerr,I had forgotten about that.Very useful in a number of different places- saving both (execution) time and footprint.-M27. jul. 2023 kl. 19:31 skrev Gregory Haerr ***@***.***>:﻿
@ghaerr commented on this pull request.



In tlvc/arch/i86/drivers/block/directhd.c:
>      for (drive = 0; drive < 2/*MAX_ATA_DRIVES*/; drive++) {
 	if (!drive&1) reset_controller(drive/2);
 	/* send drive_ID command to drive */
 	out_hd(drive, 0, 0, 0, 0, ATA_DRIVE_ID);
 
 	port = io_ports[drive / 2];
 
-	/* wait */
-	while (WAITING(port));
+	/* wait -- if status is 0xff, there is no drive with this number */
+	mdelay(OLD_IDE_DELAY);

A system-independent way of busying looping to a precise number of milliseconds was added to the ELKS idequery driver, both to allow timing out on a nonexistent IDE device, as well as to work regardless of the CPU speed. While still a busy-loop (rather than process sleep), this method uses the volatile jiffies 1/100 second timer set by the hardware timer interrupt. Here's the code, which is also being used in serial port timeouts when a precise timer is needed:
/* return true if time a is after time b */
#define time_after(a,b)         (((long)(b) - (long)(a) < 0))

#define WAIT_READY      (3*HZ/100) /* 3/100 sec = 30msec - should be instantaneous */

while (1) {
	    unsigned long timeout = jiffies + WAIT_READY; /* 30 ms */
	    out_hd(drive, IDE_DRIVE_ID);
	    while ((STATUS(port) & 0x80) == 0x80) {    	/* wait 30ms until not busy */
		if (time_after(jiffies, timeout))
		    goto out;				/* timeout */
	    }

You might be able to use this method instead of mdelay, while also reading the status port and aborting if status == 0xff.

Mellvik:
@ghaerr, 
This is (ancient) history haunting us - seemingly again and again :-)

When the previous PR was merged, I had what I thought was a reasonably complete IDE/ATA driver. Then I fired up the 286 compaq with its original 40MB Conner drive, and nothing worked. Not even the ID probe. And suddenly I found myself in another deep rabbit hole, as chronicled in the Developer Notes Wiki. The journey is quite well described there, but a few additional points:


> 27. jul. 2023 kl. 19:41 skrev Gregory Haerr ***@***.***>:
> 
> 
> @ghaerr commented on this pull request.
> 
> In tlvc/arch/i86/drivers/block/directhd.c <https://github.com/Mellvik/TLVC/pull/19#discussion_r1276618161>:
> 
> > @@ -431,47 +467,71 @@ int INITPROC directhd_init(void)
>  	 * of them even have more than 16.384 cylinders
>  	 *
>  	 * This is some sort of bugfix, we will use same method as real Linux -
> -	 * work with disk geometry set in current translation mode rather than
> -	 * using physical drive info. Physical info might correspond to logical
> -	 * info for some drives (those which don't support/use LBA mode).
> +	 * work with disk geometry set in current translation mode (54-56) if valid (53)
> +	 * rather than physical drive info. Old drives have only physical, which may
> +	 * be misleading: The copnmfigured BIOS drive type's CHS mapping may be different.
> +	 * E.g the Compaq drive type 17 (conner 42MB) reports 806/4/26 while the BIOS
> Are you saying here that the "BIOS configured value" (assume from user or computer manufacturer configuration) should always override what the drive itself might report (if indeed it reports anything)? If both work, what is the reason that drive and BIOS CHS may be different?
> 
Depends. If sharing the drive with other OSes which use the BIOS values, those have to be used (unless the difference is in cylinders only, then it doesn’t matter). Otherwise use the physical values from the drive.
I’m assuming one of the reasons for Compaq et al to create IDE in the first place was to be able to play tricks with the Physical vs actual CHS values. Disk drive technology was evolving fast at the time and they had no way to update BIOSes after the systems left the factory. IDE allowed new drives with slightly different specs but similar capacity to use use the old drive table (this is the 80s). In the Compaq case disk type 17 originated as a 5.25in drive  with CHS 980/5/17 (all early drives had 17 sectors per track), the 3.5in Conner drive is 806/4/26.
> It would seem then for TLVC that a CHS specification (likely from /bootopts?) would just have to match the way that TLVC calculates CHS from block numbers. (?)
> 
Given the above, TLVC must either query the BIOS (bad) or have another mechanism (like bootopts) - or ignore compatibility and use physical values from the drive.
> In the past, I have only noticed possibly a cylinder number being 1 greater through the IDE probing than the BIOS-reported value.
> 
That's another historical thing, the spare cylinder of yesteryear (reserved for the bad sector table, spares, sometimes diagnostics). What you’re seeing (via the likes of TLVC’s IDEprobe) though is usually the actual CHS values, not the physical ones (I changed that in the TLVC IDE probe after being hit with the historical realities). They’re stored at different locations in the IDE ID block on the drive. There is a flag in the ID block indicating whether the actual values are valid or not.

—M

Mellvik:
Again thanks for taking the time to comment and ’think loud’ @ghaerr. Appreciated - and further commenting/discussions are welcome.

First, the primary goal is to provide a direct path to block devices - bypassing buffers  and other intermediate mechanisms useful primarily for file system access. Check out the Technical design notes wiki for more.

Like you say, there are many considerations to be made, and my intent is to keep it simple - resist the temptation to ‘do this one too’ because it’s so close (or easy whatever). 

More comments and observations below.

> 27. jul. 2023 kl. 20:25 skrev Gregory Haerr ***@***.***>:
> 
> 
> Wow, you have been busy!! It looks like you're having a lot of fun writing this deep in the middle of the kernel block, buffer, driver and raw device code :)
> 
> The idea of a raw device driver combined with multi-sector I/O is pretty neat, as it could provide for a way to greatly speed up certain copy operations. I thought to ask a few questions to better understand your design goal for this, as well as to point out some design considerations that went into the synchronous block driver and buffer design.
> 
> If the raw device driver could perform multi-sector I/O directly into an application buffer, copies could get very fast, compared to using sync or async I/O through the block device system.
> 
Yes, this is what the raw driver does now. When the raw floppy driver has been added, fixing fdtest is on the list to get some real numbers.
> Of course, the primary reason the block drivers don't do multi-sector I/O is that there's no guarantee the next file block is contiguous with the one being read (although it could be, that's a different story for another time). The block drivers do attempt a two-sector contiguous read (for a 1K block) or sometimes a full-track read (when track caching is enabled).
> 
> Then we have the problem of exactly what physical memory addresses a disk sector can be read or written from. For the block drivers, this is always either an L1 (kernel data segment) or L2 (possibly main or XMS) memory. XMS memory can't ever be direct, and a copy must occur. Some older PCs don't allow disk I/O around a 64K segment overlap, and some supposedly require I/O only into the lower 64K physical (those systems must be configured to use L1 buffers only and then perform I/O into the fixed DMASEG buffer only). For both the L1 and L2 buffers (actual data buffer locations, not headers), care has been taken to allocate them aligned to a 1K-boundary (for both kernel data segment and main memory). Since the size of I/O is always 1K and the buffers are 1K-aligned, the 64k DMA (or other) wrap issue is solved. (I'm not sure whether this issue is only for FD and DMA, or whether it’s also an HD issue as well). Of course, the BIOS is doing all the I/O, we're just passing the ES:BX buffer address which is aligned properly.
> 
I have a very pragmatic approach to this: Keep it simple and fix it if it breaks. Large transfers are attractive, but in practice, the largest IO raw request likes is coming from dd and will be 20k - and at no point ever close to 64k. Further: Disk IO is programmed IO which means there are no if any restrictions related to memory and addressing. For floppy, I haven’t decided yet whether to bypass the track buffer or not. Probably - at least in the fist cut – I’ll keep it just to avoid getting into trouble with DMA and 64k boundaries. That said, doing DMA anywhere in the first meg of memory (and let the DMA code handle boundaries) is no big deal. I have not been aware of restrictions in older systems, so I’ll make a note of that.
> I noticed that the direct HD driver is using a local buffer.
> 
It's not. 
> Hopefully that can be redesigned to go directly to an L1 or L2 buffer, unless XMS is running (let’s worry about XMS a bit later for this discussion). If the FD or HD driver has to use a local buffer, then there won't be a max speed improvement for using raw devices, since internal buffer copies will have to be made.
> 
The driver has a local buffer for debugging, conveniently available as a bounce buffer for XMS buffers. The general case is to transfer data directly to/from buffers - or, in the raw case, to/from user space.
> In order to use raw devices at full speed, the application, after opening the raw device, must pass a buffer that is 1K aligned, and also know what the max multi-sector count is. Then it could execute an (async) read directly into its buffer. Nice and fast!
> 
We’re doing this now, why would there have to be a 1K alignment? 
> This method would still require that the current (raw) file offset is sector-aligned (otherwise requiring mandatory buffering of intra-sector contents).
> 
This is how the raw driver works. It uses a regular buffer if required for partial start/end sectors, otherwise copies directly to from user space. The non-sector-aligned requests are probably very rare.
> Thus, IMO, it would be interesting to see whether the new raw device mechanism could be designed for maximum throughput first, then add considerations for buffering later, as an option. (These buffer options would be required for non-aligned user buffer, non-sector-aligned file offset, XMS buffer address, and possibly others. They could possibly all be handled with a single slow-path method, frankly using the previous ll_rw_blk raw driver that uses system buffers for I/O and handles non-aligned file and buffer offsets).
> 
You may be losing me here: From my point of view we’re doing this adequately (if not necessarily in the fastest possible way) already.
> I have some thoughts as to how this could all be accomplished outside the block device/buffer system, only using an updated I/O request header, not going through ll_rw_blk at all. Also, the track or cylinder cache system could be rewritten to run as just another multi-sector I/O request (using aligned buffers) and thus be removed from being driver-specific. Essentially all this would live below the current buffer system and above the block driver, at the nr_request level.
> 
You’ve probably lost me by now, but let me add a few things. First, IDE/ATA drives (except the pre-90 generation) have significant on-drive caches which are faster than anything we can provide on old machines, so low-level caching of disk IO is probably not very useful. Also, using DMA except for floppies is just too slow and PIO delivers a lot of flexibility even though it ties up the processor while ongoing. I do however, plan to add interrupts to the IDE driver to get rid of the wait loops.
Further, multisector IO is a device dependent thing. Floppies can do a full cylinder IO in one swoop, which means 36 sectors max (the direct driver does not support 2880 floppies), 18k bytes. All IDE drives support semi-multisector IO: one command, multiple responses (one interrupt per sector). Newer drives and CF cards support real multisector IO, one command, one response - up to 256 sectors per operation. The driver will choose the best possible variant - and soon (hopefully) be using interrupts to signal completions. I don’t think the difference between semi-multi and real multi IO is all that significant, but the interrupt load may be.
> I am curious as to your design goals and thoughts on this, and can continue commenting if desired, thank you!
> 
—M

ghaerr:
> as chronicled in the Developer Notes Wiki. 
> Check out the Technical design notes wiki for more.

Very good point about the on-drive caches on IDE/ATA drives, I hadn't thought about that. And super documentation on your journey, I'll read the Developer and Design Notes to keep up with what you've learned (I had previously looked and they weren't there yesterday :) Nice work!

> You may be losing me here: From my point of view we’re doing this adequately (if not necessarily in the fastest possible way) already.

It's probably me not fully understanding what you've done, I will look at the raw driver in more detail. Good points about the max read sizes, its easy to forget just how limited 16-bit real mode systems are: both `read` and `write` system calls can't pass even a full 64K buffer, and there are a number of issues within the system that might causes issues even at 32K being requested (e.g. length interpreted as negative number rather than unsigned, etc).

With regards to the nr_request system, the biggest concern I see is around allocating a kernel system buffer, (rather than, say, local or far heap memory) for I/O buffer space. Once a BH or EBH buffer header is used, the rest of the system (meaning the fs/buffer kernel buffer management code) may think the buffer is OK to use for an unrelated filesystem I/O request if the block numbers match. This could cause hard-to-find, time related filesystem errors, especially given that an I/O complete and/or kernel task switch can cause the buffer code to operate in much different conditions than it previously has had to. I had noticed that nr_request is still using a BH pointer for raw driver I/O, so I thought to bring this up. That basically ties into my point about being careful with using `ll_rw_blk` for raw driver I/O.

> We’re doing this now, why would there have to be a 1K alignment?

I must have misread the HD driver, but I'm only looking at each individual PR, not the finished product and there's lots of changes :) It seems the driver reads 256 words from an I/O address and copies directly, so this is no issue. The 1K alignment then would only become an issue for the raw FD driver coupled with DMA, where there is an limitation disallowing direct I/O across a 64K boundary. The application could pass a buffer which happens to cross a 64k boundary, if the DMA is done directly to it.

Mellvik:
>  Good points about the max read sizes, its easy to forget just how limited 16-bit real mode systems are: both read and write system calls can't pass even a full 64K buffer, and there are a number of issues within the system that might causes issues even at 32K being requested (e.g. length interpreted as negative number rather than unsigned, etc).
> 
Interesting point, that would be a great thing to test while at it, just make a (possibly temporary) change to dd and allocate 34 k of buffer space (system allowing) and see what happens.
> With regards to the nr_request system, the biggest concern I see is around allocating a kernel system buffer, (rather than, say, local or far heap memory) for I/O buffer space. Once a BH or EBH buffer header is used, the rest of the system (meaning the fs/buffer kernel buffer management code) may think the buffer is OK to use for an unrelated filesystem I/O request if the block numbers match. This could cause hard-to-find, time related filesystem errors, especially given that an I/O complete and/or kernel task switch can cause the buffer code to operate in much different conditions than it previously has had to. I had noticed that nr_request is still using a BH pointer for raw driver I/O, so I thought to bring this up. That basically ties into my point about being careful with using ll_rw_blk for raw driver I/O.
> 
OK, I read you now. And I have made some effort to ensure that the buffer is completely invalidated when released - and I’ll make a note to doublecheck on that. 

Thanks.

—M

Mellvik:
> 28. jul. 2023 kl. 18:06 skrev Gregory Haerr ***@***.***>:
> 
> 
> as chronicled in the Developer Notes Wiki.
> Check out the Technical design notes wiki for more.
> 
> Very good point about the on-drive caches on IDE/ATA drives, I hadn't thought about that. And super documentation on your journey, I'll read the Developer and Design Notes to keep up with what you've learned (I had previously looked and they weren't there yesterday :) Nice work!
> 

Memory lapse - I thought I had published these notes already. There is a lot more in the ’stash’.

Thanks

-M

ghaerr:
> that would be a great thing to test while at it, just make a (possibly temporary) change to dd and allocate 34 k of buffer space (system allowing) and see what happens.

I'd bet that filesystem read/write works fine, but there are many other drivers "underneath" that might not have been checked or coded so thoroughly (think tcpdev, UNIX or nano stacks, possible char drivers, etc).

However, I do know that `malloc` fails with 32K and over (negative length). So if the application dynamically allocates a buffer, that'll fail prior to the kernel. Fixing `malloc` has been on my list, but as you point out, there aren't as many applications as one may think that actually requiring large buffers so this hasn't been a big user problem. (And of course, with 1K max synchronous I/O, we've been way below that limit as far as any speed improvements for larger I/O requests!)

> And I have made some effort to ensure that the buffer is completely invalidated when released - and I’ll make a note to doublecheck on that.

Well, you have reported an inode-trashing problem (mv -> ls), and that's just what a buffer problem could look like. But we're talking about raw disk driver here, so that's not this problem. The issue isn't so much that the buffer be properly invalidated and/or released, but that the buffer system could grab that buffer and use it for writing, link it into the LRU list, etc, none of which you want. Strenuous testing of simultaneous raw disk I/O along with normal I/O will bring out any issues, of course. The existing stability problem will have to be solved first, before worrying about how the raw disk driver buffers might also interfere!

ghaerr:
Hello @Mellvik,

I've been meaning to comment more fully on the raw disk driver, and now that the deep dive into buffers seems to have solved the data corruption problem, I thought to mention an idea I had that IMO would make your raw disk driver much simpler.

Right now, both TLVC and ELKS have to perform I/O in 1K blocks, rather than sectors. The primary reason for this is the that request buffer uses rq_blocknr, rather than rq_sector. This was changed some time ago, as the only calls to the block drivers were for 1K blocks, and, with variable sector sizes supported (only for the PC-98 port), it was far easier to calculate the sector count at the driver level, rather than during the call to `ll_rw_blk`, and all I/O went through `ll_rw_blk` anyways.

I have been thinking of changing the I/O request layer in ELKS to use rq_sector and rq_nr_sectors (with the latter always being two for IBM PC block requests). Putting aside the PC-98 variable sector size issue, which TLVC doesn't care about, changing the request layer to always use a start sector and sector count would greatly simplify raw disk drivers, as `ll_blk_rw` could be entirely skipped, as well as the need for any buffer headers or even temp buffer space. (Intra-sector temp buffer could always use the single sector DMASEG). The raw disk code would just issue an I/O request for the sector directly and skip all the extra junk currently required to split blocks into sectors.

There are also some other very cool benefits of the minor recoding of HW drivers to use rq_sector/rq_nr_sectors rather than rq_blocknr. Take multi-sector I/O: the driver is setup to take the start sector and sector count directly from the request layer. That means that the track (or possibly cylinder) caching could move up a level and just ask for a track cache read directly from the FD or HD driver directly to a cache buffer, above the block device layer. Pretty cool, as new block devices wouldn't have to implement the cache code again each time.

For TLVC, this should be very straightforward, merely by removing rq_blocknr and replacing it with the appropriate rq_sector and rq_nr_sectors, and making the adjustments in each block driver. For ELKS, its a bit more complicated, as the lower level needs to report the sector size in use by major/minor number before calculating the start sector.

BTW, another addition I'm looking at is for the HW driver to report the max logical sector number, and to have that reported upwards and checked in make_request (see BDEV_SIZE_CHK). For the record, the last big problem where L1/L2 buffer addresses were being overwritten ended up getting a block number from an indirect block that was unmapped, remember? Well, that block number was way larger than the size of the device block, but since neither ELKS nor TLVC is checking max block/sector size, the error was occurring and the error was sometimes never reported as an I/O error, if the CHS was still within the overall HD size.

Mellvik:
Thanks @ghaerr,It seems (again) we're thinking along the same lines.The TLVC direct drivers are already doing this. The blocknr is actually sector number (it's an idea to change the name for clarity!), which as you allude to, makes a number of things easier.Raw for floppies is still pending, including how to handle the track buffer. I've decided to do the async hd driver first, inspired by our work on buffers!The max logical sector, no, I don't remember so I'll do a refresh on that. Sounds like a good idea!- M20. aug. 2023 kl. 19:47 skrev Gregory Haerr ***@***.***>:﻿
Hello @Mellvik,
I've been meaning to comment more fully on the raw disk driver, and now that the deep dive into buffers seems to have solved the data corruption problem, I thought to mention an idea I had that IMO would make your raw disk driver much simpler.
Right now, both TLVC and ELKS have to perform I/O in 1K blocks, rather than sectors. The primary reason for this is the that request buffer uses rq_blocknr, rather than rq_sector. This was changed some time ago, as the only calls to the block drivers were for 1K blocks, and, with variable sector sizes supported (only for the PC-98 port), it was far easier to calculate the sector count at the driver level, rather than during the call to ll_rw_blk, and all I/O went through ll_rw_blk anyways.
I have been thinking of changing the I/O request layer in ELKS to use rq_sector and rq_nr_sectors (with the latter always being two for IBM PC block requests). Putting aside the PC-98 variable sector size issue, which TLVC doesn't care about, changing the request layer to always use a start sector and sector count would greatly simplify raw disk drivers, as ll_blk_rw could be entirely skipped, as well as the need for any buffer headers or even temp buffer space. (Intra-sector temp buffer could always use the single sector DMASEG). The raw disk code would just issue an I/O request for the sector directly and skip all the extra junk currently required to split blocks into sectors.
There are also some other very cool benefits of the minor recoding of HW drivers to use rq_sector/rq_nr_sectors rather than rq_blocknr. Take multi-sector I/O: the driver is setup to take the start sector and sector count directly from the request layer. That means that the track (or possibly cylinder) caching could move up a level and just ask for a track cache read directly from the FD or HD driver directly to a cache buffer, above the block device layer. Pretty cool, as new block devices wouldn't have to implement the cache code again each time.
For TLVC, this should be very straightforward, merely by removing rq_blocknr and replacing it with the appropriate rq_sector and rq_nr_sectors, and making the adjustments in each block driver. For ELKS, its a bit more complicated, as the lower level needs to report the sector size in use by major/minor number before calculating the start sector.
BTW, another addition I'm looking at is for the HW driver to report the max logical sector number, and to have that reported upwards and checked in make_request (see BDEV_SIZE_CHK). For the record, the last big problem where L1/L2 buffer addresses were being overwritten ended up getting a block number from an indirect block that was unmapped, remember? Well, that block number was way larger than the size of the device block, but since neither ELKS nor TLVC is checking max block/sector size, the error was occurring and the error was sometimes never reported as an I/O error, if the CHS was still within the overall HD size.

ghaerr:
> The TLVC direct drivers are already doing this. The blocknr is actually sector number 

Oh, I see. I was looking at old code from this PR. In that code, it shows the raw driver stuffing the sector number in the `b_count` field and ORing it with 0x80 to indicate so. Sounds like your current use of rq_blocknr as the sector number is a much better idea. When I rename rq_blocknr, it will immediately flag via compiler errors all the other block drivers that use it, allowing their associated code to be easily updated to correct for the new sector value being twice as large as the old block number. (At least I hope so, I haven't looked at the cache code lately which might be a more complicated!?) One more reason to get the cache code moved out of bioshd.c!

I also see the raw driver calling `ll_blk_rw`. What I was saying is that by adding rq_nr_sectors (instead of assuming a single 1K block and using rq_blocknr), the raw block driver could skip `ll_blk_rw` completely, not deal with buffers at all, using a modified `make_request` to add the request directly for the lower level in sectors. Perhaps you've already done that, I was just looking at this PR's code.

Mellvik:
> > The TLVC direct drivers are already doing this. The blocknr is actually sector number
> 
> Oh, I see. I was looking at old code from this PR. In that code, it shows the raw driver stuffing the sector number in the `b_count` field and ORing it with 0x80 to indicate so. 

You're actually looking at the right code. There are a number of fixes in a forthcoming commit but no big changes.
`b_count` is the number of blocks to read/write, `b_blocknr` is the starting sector. The trick with `b_count` is to avoid introducing yet another variable (`rq_nr_sectors`) in the structure, which is large as it is...

> Sounds like your current use of rq_blocknr as the sector number is a much better idea. When I rename rq_blocknr, it will immediately flag via compiler errors all the other block drivers that use it, allowing their associated code to be easily updated to correct for the new sector value being twice as large as the old block number. (At least I hope so, I haven't looked at the cache code lately which might be a more complicated!?) One more reason to get the cache code moved out of bioshd.c!

Agreed, and yes, I've been through that one - there were a number of rather elusive assumptions about the block size hidden here and there. Now that it's done and working, the name is less important - but `rq_sector` is arguably more intuitive. 

> 
> I also see the raw driver calling `ll_blk_rw`. What I was saying is that by adding rq_nr_sectors (instead of assuming a single 1K block and using rq_blocknr), the raw block driver could skip `ll_blk_rw` completely, not deal with buffers at all, using a modified `make_request` to add the request directly for the lower level in sectors. Perhaps you've already done that, I was just looking at this PR's code.

That was my initial approach, and after a couple of attempts (some of them still visible as comments and #if 0 in the code) I decided to go for the current approach. The challenge is to have a reliable (and simple) way to return in an async setting. Buffer headers are rigged for that with the sleep/wake + add_request/end_request setup. Exporting the request queue to a higher level can work, but it becomes complicated.
So I made the choice to use buffer headers for the raw requests, wasting the buffer space. A small price to pay for the simplicity delivered. After all, raw IO is the exception, not the rule.

-M

Mellvik:
BTW - the direct floppy driver works fine with bios HD. That's how it was developed. There is a caveat about XMS, I don't remember off hand. The opposite has not been tested.

Different note: the L1 segment address is not word aligned. Which has no functional consequences but might affect performance on machines later than 8086/88.

ghaerr:
> BTW - the direct floppy driver works fine with bios HD. 

Good to know! I had figured it probably would, since all BIOSes can handle interrupts on top of a synchronous call, just as long as the interrupt code doesn't call another BIOS function, or sleep the task, neither of which the kernel does.

> the L1 segment address is not word aligned.

Arggh, thank you! And here I was thinking it was 64k overlap safe, and its not even word aligned! I think that can be fixed with a gcc alignment directive. 

On the subject of some very very old machine BIOSes not supporting I/O that may end up crossing a physical 64K boundary in RAM (described above): perhaps you can give some hardware input on this, as the kernel does take steps (at least in the BIOS driver, obviously not in the L1 data layout) to ensure this doesn't happen. IIRC this has to do with the original 8237 DMA controller and the IBM design... which I've learned from you is only really used for FDC controller access, and is seldom used otherwise. Did the 8237 have any issues with its address lines wrapping in a bad way around a 64K boundary, to your knowledge?

Perhaps its not worth keeping up all the driver code that uses bounce buffers, along with (attempted) alignments of L1 and L2 buffers (which are 1K in size and aligned to 1K boundaries for 64K wrap protection) for this case? I would like your opinion on any hardware issues you might see, especially as you've said some changes had to be made to the DMA code to get the direct FD driver running.

ghaerr:
I've been thinking through more potential issues with a raw char driver implementation, and am interested to know what BSD or other UNIX systems might do in the case of simultaneous I/O on a raw disk: what exactly is supposed to happen if a process is say, reading the raw device, but a system buffer is present with the same block, dirty and unsynced? Should all system buffers for the raw device be written before the raw read starts, or is the result undefined on simultaneous access? 

A similar issue arises if a process is writing the raw disk, but there happens to be a system buffer with a block that's within  the raw disk write block range - should the system buffer be updated, or remain unchanged, resulting in data mismatch to an application later?

What do you think? I haven't looked at the BSD code to determine whether raw devices lock the device entirely and invalidate buffers to prevent this, try to somehow manage the buffers and raw read/writes seperately, or allow undefined simultaneous operations? It would seem the latter would be a dangerous implementation.

Mellvik:
> > BTW - the direct floppy driver works fine with bios HD.
> 
> Good to know! I had figured it probably would, since all BIOSes can handle interrupts on top of a synchronous call, just as long as the interrupt code doesn't call another BIOS function, or sleep the task, neither of which the kernel does.

BTW - your script to push the task stack through the roof would be of interest. I've never seen a number above 540 even though I really tried to push it (deep and large directories on FAT, removing the statics). Possibly the BIOS drivers are more stack intensive than the direct driver, but that would be good to verify.

> 
> > the L1 segment address is not word aligned.
> 
> Arggh, thank you! And here I was thinking it was 64k overlap safe, and its not even word aligned! I think that can be fixed with a gcc alignment directive.
> 

Selective blindness - I should have noted this a long time ago, and here's what it took: In one compile the buffer segment is the usual 0x2400, the L1 seg is 0x1400 - they're so close visually that mistakes were made while tracing - and attention was aroused. Another compile and the L1 seg is 0x13ff. Suddenly the BIG question mark popped up ...

> On the subject of some very very old machine BIOSes not supporting I/O that may end up crossing a physical 64K boundary in RAM (described above): perhaps you can give some hardware input on this, as the kernel does take steps (at least in the BIOS driver, obviously not in the L1 data layout) to ensure this doesn't happen. IIRC this has to do with the original 8237 DMA controller and the IBM design... which I've learned from you is only really used for FDC controller access, and is seldom used otherwise. Did the 8237 have any issues with its address lines wrapping in a bad way around a 64K boundary, to your knowledge?

I'm not an expert on this but here's my take: The 8237 is indeed the only component that must be handled with care as far as memory boundaries are concerned. It was made for 8 bit systems (8085) and has no clue about addresses beyond 16 bits. In order to make this work with 20 bit addressing, IBM added a set of 'page registers' outside of the DMA controller, one per DMA channel. The consequence is kind of obvious: Do not cross 64k physical boundaries.

Now, since the PC and XT used DMA for both HD and floppy (and some add on hardware such as the sound blaster IIRC), and to avoid the extra complexity it would require in the BIOS, restrictions were placed on alignment when using BIOS IO.

In the TLVC case, we have complete control and are avoiding this complexity. The floppy driver (actually the DMA driver if you like) can handle the splitting of transfers that span 64k boundaries, which isn't very complicated. Like I probably said before, the DMA - even on the 386 - is too slow to use it for HD IO. Thus I plan to remove the general DMA support in TLVC and just use a simple DMA 'function' in the floppy driver. Smaller footprint, slightly faster.
 
> 
> Perhaps its not worth keeping up all the driver code that uses bounce buffers, along with (attempted) alignments of L1 and L2 buffers (which are 1K in size and aligned to 1K boundaries for 64K wrap protection) for this case? I would like your opinion on any hardware issues you might see, especially as you've said some changes had to be made to the DMA code to get the direct FD driver running.

My take (and intent) is to (carefully) remove all of this. If there is something I've missed, I'm sure to discover it that way - but I don't think so. Not having to consider BIOS requirements is a blessing in that regard.

A curiosity The DMA page registers make DMA the only place in the system where really linear addresses are being used.
Dma address pc

Mellvik:
> I've been thinking through more potential issues with a raw char driver implementation, and am interested to know what BSD or other UNIX systems might do in the case of simultaneous I/O on a raw disk: what exactly is supposed to happen if a process is say, reading the raw device, but a system buffer is present with the same block, dirty and unsynced? Should all system buffers for the raw device be written before the raw read starts, or is the result undefined on simultaneous access?
> 
> A similar issue arises if a process is writing the raw disk, but there happens to be a system buffer with a block that's within the raw disk write block range - should the system buffer be updated, or remain unchanged, resulting in data mismatch to an application later?
> 
> What do you think? I haven't looked at the BSD code to determine whether raw devices lock the device entirely and invalidate buffers to prevent this, try to somehow manage the buffers and raw read/writes seperately, or allow undefined simultaneous operations? It would seem the latter would be a dangerous implementation.

The approach to raw devices is actually very simple - the thinking being 'to get to the device'. Which in practice means not only bypassing, but ignoring the existence of buffers, file systems etc.etc. And the fundamental idea that 'don't use unless you know what you're doing'.

The raw devices were/are sys-admin access only - use with care. IOW never access  the raw device fs a mounted file system. No locks or anything, just don't to it. 

Raw access is for the likes of fsck, fdisk, makeboot, mkfs etc (fdtest eventually in our case) - plus copying drives or partitions.

ghaerr:
> I'm not an expert on this but here's my take: The 8237 is indeed the only component

Thank you @Mellvik, what a super an explanation of exactly how the 8237 operates and the 64k limitation!! I never knew about the hardware page registers. I now feel I fully understand the overall issue (and now I'm pretty sure the BIOS driver may have a bug requiring a bounce buffer on L1 requests, depending on the exact position of the L1 buffer within the kernel data segment. I had previously thought I/O never occurred directly into/out of L1, but you've also proved that incorrect). Your explanation should definitely be added to your already very informative Developer Notes.

I also wasn't aware that the XT used DMA, which would mean early HD BIOSes likely suffer from the same limitation?

> the DMA - even on the 386 - is too slow to use it for HD IO. Thus I plan to remove the general DMA support in TLVC and just use a simple DMA 'function' in the floppy driver. Smaller footprint, slightly faster.

Great idea.

> My take (and intent) is to (carefully) remove all of this.

It would still be required for the BIOS driver on older machines, something to consider. I'm not sure if you're thinking of jettisoning the BIOS drivers altogether or not?

> In one compile the buffer segment is the usual 0x2400, the L1 seg is 0x1400 - they're so close visually that mistakes were made while tracing - and attention was aroused. Another compile and the L1 seg is 0x13ff. 

Ok - in this case, we're not talking about a non-word-aligned L1 buffer, but rather, the L1 segment being an odd number, which is perfectly OK as its just a segment address which is always 16-byte aligned. The kernel is loaded code segment first, then data segment, then free memory. What you're seeing is that as code is subtracted (in this case), the data segment is able to be loaded 16 bytes lower at 0x13FF. Since the kernel data segment is always 64K, the buffer segments are always segment address 0x1000 larger (=64k) for their start address (0x1400 vs 0x2400 etc).

So we don't really know what the actual L1 buffer address offset is (lower 4 bits). It could be that gcc already word aligns the buffer, but maybe not. I'll check the .S output of buffer.c and see what's really happening with and without my fix.

> The approach to raw devices is actually very simple - the thinking being 'to get to the device'. Which in practice means not only bypassing, but ignoring the existence of buffers, file systems etc.etc. 

Your description helps - the entire purpose of the raw driver is to entirely skip over complex internal buffering and directly read/write the device, period. I can see how this could help with `fsck` etc, provided the filesystem wasn't mounted.

Mellvik:
> > I'm not an expert on this but here's my take: The 8237 is indeed the only component
> 
> Thank you @Mellvik, what a super an explanation of exactly how the 8237 operates and the 64k limitation!! I never knew about the hardware page registers. I now feel I fully understand the overall issue (and now I'm pretty sure the BIOS driver may have a bug requiring a bounce buffer on L1 requests, depending on the exact position of the L1 buffer within the kernel data segment. I had previously thought I/O never occurred directly into/out of L1, but you've also proved that incorrect). Your explanation should definitely be added to your already very informative Developer Notes.

That's a good idea, I'll put that on my list. As it turns out, I'm becoming an avid user of my own developer notes. :-)
As to DMA: I didn't mention the AT which IIRC has 2 8237s and a wider page register, but that's for another day and time (Wikipedia probably has the details). 
Also, this doesn't have to be a BIOS bug, but may be a specific requirement documented in the BIOS docs.

> 
> I also wasn't aware that the XT used DMA, which would mean early HD BIOSes likely suffer from the same limitation?
> 

Again it depends on whether we consider it a requirement or a bug. I'm leaning towards the former. And yes, the original PC and XT were born with the 8237. There was no other way to handle floppy IO at the time, as the processors were too slow to deal with a real time data stream from the (rather stupid) floppy controller. My first (home-brew) computer was 8085 based and had an 8237 to handle floppy IO. :-)

> > the DMA - even on the 386 - is too slow to use it for HD IO. Thus I plan to remove the general DMA support in TLVC and just use a simple DMA 'function' in the floppy driver. Smaller footprint, slightly faster.
> 
> Great idea.
> 
> > My take (and intent) is to (carefully) remove all of this.
> 
> It would still be required for the BIOS driver on older machines, something to consider. I'm not sure if you're thinking of jettisoning the BIOS drivers altogether or not?

No, it is convenient to have the BIOS option available, and you're right - such removal would have to be conditional.

> 
> > In one compile the buffer segment is the usual 0x2400, the L1 seg is 0x1400 - they're so close visually that mistakes were made while tracing - and attention was aroused. Another compile and the L1 seg is 0x13ff.
> 
> Ok - in this case, we're not talking about a non-word-aligned L1 buffer, but rather, the L1 segment being an odd number, which is perfectly OK as its just a segment address which is always 16-byte aligned. The kernel is loaded code segment first, then data segment, then free memory. What you're seeing is that as code is subtracted (in this case), the data segment is able to be loaded 16 bytes lower at 0x13FF. Since the kernel data segment is always 64K, the buffer segments are always segment address 0x1000 larger (=64k) for their start address (0x1400 vs 0x2400 etc).

Of course - I didn't think straight about this one. Thanks for pointing that out. Come to think of it it is very unlikely that gcc would actually allocate a buffer like that on an odd address.

> 
> So we don't really know what the actual L1 buffer address offset is (lower 4 bits). It could be that gcc already word aligns the buffer, but maybe not. I'll check the .S output of buffer.c and see what's really happening with and without my fix.
> 
> > The approach to raw devices is actually very simple - the thinking being 'to get to the device'. Which in practice means not only bypassing, but ignoring the existence of buffers, file systems etc.etc.
> 
> Your description helps - the entire purpose of the raw driver is to entirely skip over complex internal buffering and directly read/write the device, period. I can see how this could help with `fsck` etc, provided the filesystem wasn't mounted.

As tradition goes, the system utilities that need (or should) use raw devices will bark at you if the 'r' is missing from the device name in the arglist. Some of them might even check for a mount, but that's application level.

Back to the alignment thing for a second. It occurred to me after you pointed out that there are alignment checks in the buffer code, that this may be a factor in the L1/L2 mystery. I'm looking into that just to make sure.

ghaerr:
> it depends on whether we consider it a requirement or a bug. I'm leaning towards the former.

I agree - I'll consider the 64K wrap check as a requirement. Would this be a requirement for AT+ BIOSes as well? That is, was the requirement relaxed in later BIOSes? Or does it just depend on hardware, with no BIOS specification?

>  It occurred to me after you pointed out that there are alignment checks in the buffer code, that this may be a factor in the L1/L2 mystery

I just checked the BIOS driver code - it checks for XMS buffer and does the I/O into DMASEG. The L2 buffers are 64K aligned, so no check is needed. To fix the L1 alignment problem, I need to know whether this is only for PC/XT BIOSes, or also for later models: if its early systems only, then adding 1K alignment to the L1 buffers is overkill, and the BIOS can just check for DS:L1 offset being misaligned and do the exact same thing as XMS is doing, use DMASEG. Otherwise, the L1 buffers could be dynamically allocated (something I'm thinking might be a good idea anyways) and an option added to the local heap to align on a 1K boundary. (The main memory allocator already has this option, which is why the L2 buffers aren't a problem).

Mellvik:
Interesting- and very useful again - thank you @ghaerr.Another mistake that ended up not being wasted.Let's continue to make them :-)M22. aug. 2023 kl. 19:59 skrev Gregory Haerr ***@***.***>:﻿

Come to think of it it is very unlikely that gcc would actually allocate a buffer like that on an odd address.

Testing results are in, declaring an array like the following does produce word alignment:
#define WORD_ALIGNED    __attribute__((aligned(2)))
static char L1buf[NR_MAPBUFS][BLOCK_SIZE] WORD_ALIGNED;

Without WORD_ALIGNED, it is byte aligned, the .S output is:
    .local  L1buf
    .comm   L1buf,12288,1 // <-- changes to 2 with WORD_ALIGNED in

Showing byte alignment. This is likely because it's a char array, not a short/int array. So the WORD_ALIGNED is actually needed, and will stay in.
I've added a note to look at all other 'char' arrays that might be used with fmemcpyw and will mark them WORD_ALIGNED as well (for the record, libc malloc always returns word-aligned, not sure whether char arrays are word-aligned on the stack, I would guess not given the above.
Thanks!

ghaerr:
> To fix the L1 alignment problem, I need to know whether this is only for PC/XT BIOSes, or also for later models: if its early systems only, then adding 1K alignment to the L1 buffers is overkill, and the BIOS can just check for DS:L1 offset being misaligned and do the exact same thing as XMS is doing, use DMASEG. 

Guess what? After writing debug code to check for L1 addresses overlapping 64K, I find that code never got executed for reading. Instead, the system (reading using BIOS driver) runs entirely off the track cache buffer, with read I/O always into the 9K bytes starting at DMASEG!! (Of course I wrote all that code, I should remember that). So only if the track read fails, or the system is compiled without track caching, does alignment handling become an issue. The system still has potential L1 alignment issues when writing from an L1 buffer, which the below shows:

The approximate physical address (varying with kernel code and data size) of L12 (`&L1buf[NR_MAPBUFS][0]`) is:
```
L12 1541:a170 = 0001f570    (seg:addr = physaddr)
```
Given that 1K is 0x0400, even I/O at L12 won't wrap 64K, although since the actual kernel DS offset of the L1 buffers are at 0xA170 they possibly could, and the system.map file shows L1buf is included mid-way in the bss segment,. This is then why older 8088 PC/XTs can still write the filesystem and haven't crashed!

Perhaps read block access to a individual sector-oriented lower level should never be attempted for floppy devices (i.e. track cache always on), and HD should never track cache. The current system doesn't dedicate the track cache to floppy, and will reuse it for HD, and the FD driver has code in it for individual sector handling, which is not used for reading if track caching is turned on.

I'm beginning to think all this extra complexity could be moved out of the block driver, and essentially have the idea of "slow" and "fast" block drivers. For slow drivers, the upper level would always request reading full tracks (or cylinders), as the buffer cost is frankly quite low given the speedup. (Even for 2880k disks, max 18 sectors would be read, just like it does now). There would never be a case where an individual sector is read, although it does appear we'd likely have to continue to handle 2-sector writes. As I write this, I suppose rather than complicating drivers to be "slow" or "fast", just moving all the track caching code out and upstairs, and specially handling track reads above the driver, using a cache layer for "slow" devices, could work well. I have completed the conversion in ELKS for handling all block device requests using start sector and count, so this might be easier than it currently looks by looking at all the bioshd.c code.

Mellvik:
Wow @ghaerr, there is indeed a lot of learning in this exercise!! This is important. Thanks.


> > To fix the L1 alignment problem, I need to know whether this is only for PC/XT BIOSes, or also for later models: if its early systems only, then adding 1K alignment to the L1 buffers is overkill, and the BIOS can just check for DS:L1 offset being misaligned and do the exact same thing as XMS is doing, use DMASEG.
> 
> Guess what? After writing debug code to check for L1 addresses overlapping 64K, I find that code never got executed for reading. Instead, the system (reading using BIOS driver) runs entirely off the track cache buffer, with read I/O always into the 9K bytes starting at DMASEG!!
> (Of course I wrote all that code, I should remember that). 

Don't mention it - happens to me all the time!!

> So only if the track read fails, or the system is compiled without track caching, does alignment handling become an issue. The system still has potential L1 alignment issues when writing from an L1 buffer, which the below shows:
> 
> The approximate physical address (varying with kernel code and data size) of L12 (`&L1buf[NR_MAPBUFS][0]`) is:
> 
> ```
> L12 1541:a170 = 0001f570    (seg:addr = physaddr)
> ```
> 
> Given that 1K is 0x0400, even I/O at L12 won't wrap 64K, although since the actual kernel DS offset of the L1 buffers are at 0xA170 they possibly could, and the system.map file shows L1buf is included mid-way in the bss segment,. This is then why older 8088 PC/XTs can still write the filesystem and haven't crashed!

Right. And this would apply to all systems when using floppies, not only old ones. But admittedly more so on old ones if they use HD DMA.

> 
> Perhaps read block access to a individual sector-oriented lower level should never be attempted for floppy devices (i.e. track cache always on), and HD should never track cache. 

Agreed, this is what TLVC does (no HD cache for reasons mentioned previously). I've been tempted to make track caching mandatory (remove the option), the only argument to keep it is to keep the system runnable on 256k RAM systems.

> The current system doesn't dedicate the track cache to floppy, and will reuse it for HD, and the FD driver has code in it for individual sector handling, which is not used for reading if track caching is turned on.

There are some important considerations to be made in this context when introducing the raw floppy driver in TLVC, I'm coming back to that below.

> 
> I'm beginning to think all this extra complexity could be moved out of the block driver, and essentially have the idea of "slow" and "fast" block drivers. For slow drivers, the upper level would always request reading full tracks (or cylinders), as the buffer cost is frankly quite low given the speedup. (Even for 2880k disks, max 18 sectors would be read, just like it does now). There would never be a case where an individual sector is read, although it does appear we'd likely have to continue to handle 2-sector writes. As I write this, I suppose rather than complicating drivers to be "slow" or "fast", just moving all the track caching code out and upstairs, and specially handling track reads above the driver, using a cache layer for "slow" devices, could work well. I have completed the conversion in ELKS for handling all block device requests using start sector and count, so this might be easier than it currently looks by looking at all the bioshd.c code.

I have a slightly different take on this, colored by the coming raw floppy driver - as alluded to above.

First, here's how DMA transfers work at the driver level - floppy, that is, it's slightly different for HD: The driver kicks off the transfer and exits, after having told the DMA controller how many bytes to transfer. The floppy controller has no idea how much data to transfer, and the transfer continues till the DMA controller raises its transfer complete line (or some error occur, such as end-of-cylinder). Now it interrupts the device driver, which in turn does data shuffling if required and calls `end_request`.

This means that the smallest data unit that can be requested is one sector. This is also the level at which the driver can help with data alignment problems. It can split a request that spans a 64k physical boundary transparently to the upper level. Actually the driver does something similar when the system is running on L1 buffers only and the # of sectors per track is odd (as in 1.2M drives): A perfectly normal request for a 1k block spanning sectors 15 and 1 (next head) is split into two reads. When the track buffer is active (and the track buffer too small to take a full cylinder), this is conveniently taken care of by what I thought (at the time) as being smart: Read the full track + 1 sector if head == 0, read the full track if head == 1 and just ignore the first sector to simplify the logic.

The same thinking can be applied when the low level driver gets a raw request for (say) 20k bytes. Somewhat more complicated of course because there are two conditions to watch, the 64k boundary and the cylinder boundaries, but the additional logic in the driver is quite bearable. Most of the data would go directly into user space anyway, since it's below 1M, and the presence of a track buffer would not make much difference (except possibly reducing the number of physical read operations).

So, for the continuation of my TLVC project this discussion has already been very useful.

Mellvik:
One more thing about this, @ghaerr:

Since the floppy driver is transferring data to/from L2 buffers whenever possible, there has to be a 64k boundary check in the driver unless the buffers are guaranteed to be aligned. This may be easily added right away (not considering the upcoming raw device support) because there is already logic to handle addresses > 1M (UMB) via a 1K jump buffer (separate from the DMASEG bufferspace). (Since TLVC doesn't yet support UMB, this code has never been run and thus never tested, it would be about time).

Whether it's worth it to protect the track buffer from such pollution is debatable. Useful if two or more floppy drives are active, thus again only useful in a very constrained setting.

Mellvik:
> > Since the floppy driver is transferring data to/from L2 buffers whenever possible, there has to be a 64k boundary check in the driver unless the buffers are guaranteed to be aligned. This may be easily added right away
> 
> The L2 buffers are guaranteed aligned, that was added some time ago when the main memory allocator was enhanced to allow alignment on 1K boundaries, which guarantees non-64k overlap on 1K buffers. The last and only remaining alignment issue is that of L1 buffers, discussed above but not actually a problem until the linker map changes. I'm thinking of permanently solving the problem when dynamic L1 buffers are introduced perhaps.

Great; good to know! That said - if the code size or ram wasted to achieve this is significant, it may be worth it to do the work in the floppy driver, since that's the only place it will be necessary.

> 
> > because there is already logic to handle addresses > 1M (UMB) via a 1K jump buffer
> 
> You mean the HMA? I think HMA refers to the 65535 bytes above 1Mb, while UMB refers to memory between 640k and 1Mb. UMB memory support was added to ELKS in April and doesn't require any change to block drivers. HMA support could be added but got stalled because of the potential A20 line problems that were seen. Do you think it would be useful? Agreed I/O could need bounce buffers just like XMS, unless you're saying many drives have jumpers to handle EXT/HMA memory? It seems to me all of these details open up a potential can of worms given the system setup requirements that might be required to successfully boot and run the OS.

Sorry, wrong term. I was referring to the HMA. When (re)doing the direct fd driver I was pondering the reason for the 1M test in the code and concluded they (or Linus) were taking into account the possibility that HMA was supported. (Later I found that the early ELKSers, coming from 32 bit/linear memory architectures, had limited understanding of the intricacies of (then) 15 years old system :-)

> 
> > The same thinking can be applied when the low level driver gets a raw request for (say) 20k bytes. Somewhat more complicated of course because there are two conditions to watch, the 64k boundary and the cylinder boundaries, but the additional logic in the driver is quite bearable.
> 
> This was idea I was talking about at length many posts above, with ideas for very fast throughput, but at the time you responded saying that the direct driver should remain as simple as possible. 

I did, but the intent was to make it clear that this is how raw io works, so it's a given. I obviously missed on communicating that. Raw means no buffers at all, with the possible exception of bounce buffers where required by hardware.

> 
> > So, for the continuation of my TLVC project this discussion has already been very useful.
> 
> Agreed, its complicated! I've been (re)reading the BIOS driver and I find it too complicated... I would like to see in the new designs the various issues moved into reusable functions outside of specific block drivers, if possible. That would include track caching and possibly XMS/HMA bounce buffer handling. I think the kernel can handle L1/L2 address alignment so the driver doesn't have to, even though currently the driver has to handle XMS addresses with bouncing. A generic bounce function might be a good idea, which would rewrite a request, send it, then copy it afterwards.

The common goal being simplicity I feel it's important to consider where the various enhancements/optimizations are useful. It seems to me many of the issues we've been discussing are relevant for floppies only, which means relatively rare use outside of a development setting. Thus I think this should weigh in heavily when considering whether to generalize or not...

> 
> > I've been tempted to make track caching mandatory (remove the option), the only argument to keep it is to keep the system runnable on 256k RAM systems.
> 
> Good ideas. Perhaps keeping the complexity and options within the BIOS driver intact, and instead start (re)writing the caching and other routines for the new direct FD and HD drivers separately, which always use caching. In other words, leave all the old code in for systems backwards compatibility and lots of options, which new code is written from scratch for speed and future project direction.

This is a good point. When I implement raw in the floppy driver, I'll make the track cache mandatory (and finally add the code to cache a full cylinder if it fits in 9k). For the HD driver, there is no caching - and AFAIK none required.

> 
> > A perfectly normal request for a 1k block spanning sectors 15 and 1 (next head) is split into two reads. When the track buffer is active (and the track buffer too small to take a full cylinder), this is conveniently taken care of by what I thought (at the time) as being smart: Read the full track + 1 sector if head == 0, read the full track if head == 1 and just ignore the first sector to simplify the logic.
> 
> Can you clarify so I can follow what you're saying here? Is this happening now in BIOS driver, or is this what should be happening? When the multi-sector I/O was added to BIOS, I originally thought there was some debate as to whether all BIOSes would support auto-incrementing a head on a multi-sector read request. I suppose another way of saying this is: can one depend on the BIOS to read an entire cylinder at a time, or must it read only a single head (track) for compatibility? I'm aware that only some BIOS support auto-cylinder incrementing.

No, this is the direct driver and I'm just rambling (and the first point about single sector reads has already be optimized out btw). As to the debate you refer to, I remember that - including doing some testing. We landed on the optimal selection, full track IO. This is what the NEC765 (aka 8272) controller chip does, and it would just complicate things for a BIOS not to support it. So it's safe.

> 
> On that same subject, the driver went to extremes to add a DDPT (Disk Drive Parameter Table) and ensure the max sector field was accurate so that the BIOS didn't auto-increment the cylinder on a request. IIRC it is now written such that only a read request is only extended to end of track and head never incremented. That could be improved possibly by either always reading a full track (starting from sector 1 regardless of requested start sector), or what it sounds like you're saying reading a full cylinder at a time?

What I'm saying is that IMHO it's safe to assume that full track IO is safe - always. The controller will return sensible error codes if something ends up out of bounds, so it makes sense to depend on error codes (even from a BIOS call) instead of  building cushion against _possible_ errors. Of course there is the consideration you mentioned before about BIOS bugs, but I think it's reasonable to face them if they pop up rather than the cushioning.

As to the DDPT - as you can imagine it's a relief not to have to think about it at all. :-)

ghaerr:
Great discussion on this complex topic - I'm in agreement with everything you bring up.

> if the code size or ram wasted to achieve this is significant, it may be worth it to do the work in the floppy driver, since that's the only place it will be necessary.

Really great point: while at ELKS I remain concerned with always having to keep the BIOS FD/HD driver running well, it makes a LOT of sense to think in terms of keeping kluges necessary for the floppy system contained solely within that driver, rather than having to ensure that every address is aligned etc otherwise throughout the system.

BIG case in point is XMS: XMS *always* requires a bounce buffer which must be built into the driver, involving an extra copy. Why are we so worried about a possible 64K buffer overlap and also requiring a bounce buffer? The floppy is the slowest hardware in the system. Even with no alignment checking at all, the chances of overlap are only 1/64 = 1.5%. As a result of this discussion, I'm now thinking that since all drivers will have to implement at least a 1K bounce buffer for XMS, why not also just add to that a check for 64K overlap, and process that exactly the same as well? 

With this idea, it seems all the complex cases we've been discussing can be handled with a simple two step process for each request: 1) see if request needs to be split into two/multiple requests (for reasons of head or track increment and larger raw requests spanning 64k overlap), and 2) use bounce buffer (for XMS for 64k overlap within a single sector read).

Having the code hidden just in the floppy driver, instead of in the memory allocator, the raw disk handler, etc makes a lot of sense. The only real outside overhead we have right now is in the main memory allocator which aligns to a 1K boundary, which could waste up to 1K bytes. I'll look further into that.

ghaerr:
I refactored the buffer code in PR 1665 and was able to remove 6 bytes per buffer header when XMS enabled, and 4 bytes otherwise. Our deep dive into the buffer code got me to see how this could be done fairly easily.

> ebh->b_count = (chars >> SECT_SIZE_BITS) | 0x80; /* This is a kludge. */

I also noticed that the buffer structure members use an odd number of bytes storage; given that the structure isn't packed in the array of buffer heads, this means that you could declare an `unsigned char b_nr_sectors` and use that without the kluge and also not increasing the structure size - FYI.

Mellvik:
> Great discussion on this complex topic - I'm in agreement with everything you bring up.
> ...

> BIG case in point is XMS: XMS _always_ requires a bounce buffer which must be built into the driver, involving an extra copy. Why are we so worried about a possible 64K buffer overlap and also requiring a bounce buffer? The floppy is the slowest hardware in the system. Even with no alignment checking at all, the chances of overlap are only 1/64 = 1.5%. As a result of this discussion, I'm now thinking that since all drivers will have to implement at least a 1K bounce buffer for XMS, why not also just add to that a check for 64K overlap, and process that exactly the same as well?

What you're saying is to just remove the conditional and allocate (maybe `heap_alloc`) the bounce buffer regardless? I agree.

> 
> With this idea, it seems all the complex cases we've been discussing can be handled with a simple two step process for each request: 1) see if request needs to be split into two/multiple requests (for reasons of head or track increment and larger raw requests spanning 64k overlap), and 2) use bounce buffer (for XMS for 64k overlap within a single sector read).

Exactly, that's the raw driver right there.

> 
> Having the code hidden just in the floppy driver, instead of in the memory allocator, the raw disk handler, etc makes a lot of sense. The only real outside overhead we have right now is in the main memory allocator which aligns to a 1K boundary, which could waste up to 1K bytes. I'll look further into that.

Great! Please keep me posted!! Thanks.

Mellvik:
> I refactored the buffer code in PR 1665 and was able to remove 6 bytes per buffer header when XMS enabled, and 4 bytes otherwise. Our deep dive into the buffer code got me to see how this could be done fairly easily.

Great!! I'll have a look at that, sounds like something to import asap.

> 
> > ebh->b_count = (chars >> SECT_SIZE_BITS) | 0x80; /* This is a kludge. */
> 
> I also noticed that the buffer structure members use an odd number of bytes storage; given that the structure isn't packed in the array of buffer heads, this means that you could declare an `unsigned char b_nr_sectors` and use that without the kluge and also not increasing the structure size - FYI.

A good point indeed. I'll do just that… 

Thanks.

Mellvik:
> > > ebh->b_count = (chars >> SECT_SIZE_BITS) | 0x80; /* This is a kludge. */
> > 
> > 
> > I also noticed that the buffer structure members use an odd number of bytes storage; given that the structure isn't packed in the array of buffer heads, this means that you could declare an `unsigned char b_nr_sectors` and use that without the kluge and also not increasing the structure size - FYI.
> 
> A good point indeed. I'll do just that…
> 
> Thanks.

Actually, the buffer head structure is already even, maybe it's different in ELKS?
```
struct buffer_head {
    unsigned char               *b_data;        /* Address if in L1 buffer area, else 0 */
#ifdef CONFIG_FAR_BUFHEADS
};
/* a little tricky here - buffer_head is split into near and far components */
struct ext_buffer_head_s {
#endif
    ramdesc_t                   b_seg;          /* Current L1 or L2 (main/xms) buffer segment */
    block32_t                   b_blocknr;      /* 32-bit block numbers required for FAT */
    kdev_t                      b_dev;
    struct buffer_head          *b_next_lru;
    struct buffer_head          *b_prev_lru;
    unsigned char               b_count;
    unsigned char               b_locked;
    unsigned char               b_dirty;
    unsigned char               b_uptodate;
#ifdef CONFIG_FS_EXTERNAL_BUFFER
    ramdesc_t                   b_ds;           /* L2 buffer data segment */
    unsigned char               *b_L2data;      /* Offset into L2 allocation block */
    char                        b_mapcount;     /* count of L2 buffer mapped into L1 */
#endif
};
```

ghaerr:
> Actually, the buffer head structure is already even

```
struct buffer_head {
    unsigned char               *b_data;        // 2
#ifdef CONFIG_FAR_BUFHEADS
};
/* a little tricky here - buffer_head is split into near and far components */
struct ext_buffer_head_s {
#endif
    ramdesc_t                   b_seg;          // 2 or 4
    block32_t                   b_blocknr;      // 4`
    kdev_t                      b_dev;          // 2
    struct buffer_head          *b_next_lru;    // 2
    struct buffer_head          *b_prev_lru;    // 2
    unsigned char               b_count;        // 1
    unsigned char               b_locked;       // 1
    unsigned char               b_dirty;        // 1
    unsigned char               b_uptodate;     // 1
#ifdef CONFIG_FS_EXTERNAL_BUFFER
    ramdesc_t                   b_ds;           // 2 or 4
    unsigned char               *b_L2data;      // 2
    char                        b_mapcount;     // 1
#endif
};
```
Unpacked size is 23 for EXT, 27 for XMS. `b_nr_sectors` would be placed directly after `b_mapcount`.

Mellvik:
Of course - and LOL - I really managed to confuse myself this time.

Having looked at like 100.000 bh-addresses over these past weeks, all even (thanks, gcc), the actual struct barely got a glance. Ooops.

PR #24, opened 2023-09-13
[directfd] Fixes and enhancements

Mellvik:
A number off cleanups, fixes and enhancements to the direct floppy driver, including

- detection and handling of buffers crossing 64k boundaries
- elimination of the BIOS clock counter reset (now disabled at boot time)
- the floppy format code is not used and has been ifdef'ed for now
- device close now does a proper buffer cleanup

This PR is the first of a series of significant kernel/driver  related enhancements and fixes.

### DIFF
--- a/tlvc/arch/i86/drivers/block/directfd.c
+++ b/tlvc/arch/i86/drivers/block/directfd.c
@@ -135,12 +135,6 @@ static int seek = 0;
 static int nr_sectors;		/* only when raw access */
 #endif
 
-/* Should use poke() for this!! No, turns out the far pointer is better. HS */
-static unsigned char __far *fl_timeout = (void __far *)0x440L; /* BIOS floppy motor timeout counter*/
-					/* On some machines always active, must be
-					 * reset to avoid random motor shutoff.
-					 * Ihis variant takes 10 bytes */
-
 static unsigned char current_DOR = 0x0C;
 static unsigned char running = 0; /* keep track of motors already running */
 /* NOTE: current_DOR tells which motor(s) have been commanded to run,
@@ -167,14 +161,12 @@ static unsigned char running = 0; /* keep track of motors already running */
 #define MAX_BUFFER_SECTORS 18
 
 /*
- * The DMA channel used by the floppy controller cannot access data at
- * addresses >= 16MB
- *
- * Went back to the 1MB limit, as some people had problems with the floppy
- * driver otherwise. It doesn't matter much for performance anyway, as most
- * floppy accesses go through the track buffer.
+ * The 8237 DMA controller cannot access data above 1MB on the origianl PC 
+ * (4 bit page register). The AT is different, the page reg is 8 bits while the
+ * 2nd DMA controller transfers words only, not bytes and thus up to 128k bytes
+ * in one 'swoop'.
  */
-#define LAST_DMA_ADDR	(0x100000L - BLOCK_SIZE)
+#define LAST_DMA_ADDR	(0x100000L - BLOCK_SIZE) /* stick to the 1M limit */
 
 /*
  * globals used by 'result()'
@@ -319,7 +311,8 @@ static unsigned short min_report_error_cnt[4] = { 2, 2, 2, 2 };
  * The block buffer is used for all writes, for formatting and for reads
  * in case track buffering doesn't work or has been turned off.
  */
-static char tmp_floppy_area[BLOCK_SIZE]; /* for now FIXME to be removed */
+#define WORD_ALIGNED    __attribute__((aligned(2)))
+static char tmp_floppy_area[BLOCK_SIZE] WORD_ALIGNED; /* for now FIXME to be removed */
 
 #ifdef CHECK_DISK_CHANGE
 #define buffer_dirty(b)	((b)->b_dirty)
@@ -442,7 +435,7 @@ static void motor_off_callback(int nr)
 
 /*
  * floppy_on: turn on motor. If already running, call floppy_select
- * otherwise start motor and set timer to continue (motor_on_callback).
+ * otherwise start motor and set timer and return.
  * In the latter case, motor_on_callback will call floppy_select();
  *
  * DOR (Data Output Register) is |MOTD|MOTC|MOTB|MOTA|DMA|/RST|DR1|DR0|
@@ -452,10 +445,6 @@ static void floppy_on(int nr)
     unsigned char mask = 0x10 << nr;	/* motor on select */
 
     DEBUG("flpON");
-    *fl_timeout = 0;	/* Reset BIOS motor timeout counter, neccessary on some machines */
-    			/* Don't ask how I found out. HS */
-    //pokeb(0x40, 0x40, 0);	/* this variant is 10 bytes of code too, but slower than */
-				/* using the far pointer above */
     del_timer(&motor_off_timer[nr]);
 
     if (mask & running) {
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @Mellvik,

I've been reading the driver code, and have a question about how the driver (should?) work with the CMOS drive settings vs perhaps a different format floppy in the drive:

I notice that QEMU sets CMOS for a 1200k drive at startup, even if a differently-formated floppy image is provided. Does the direct driver need to use the CMOS settings, or could those be overridden with a probe or just a config setting? For instance, I tried manually setting drive 0 to 360k using `base_type[0] = find_base(0, 1)` but that resulted in I/O errors (in QEMU).

I guess a disc probing function might be more complicated than the already complicated BIOS driver's, as the low-level disc formats would have to be changed (presumably using the statement above which failed for me at init time).

Another question: I notice that some CMOS will report '5' for 2880k floppies - what's your thoughts about supporting those with another floppy_type[]/floppy_types[] entry? (Those two similar tables are confusing!).

Thank you!

Mellvik:
Hi @ghaerr, thanks for reviewing and commenting.

I seriously took the fast route on floppy type detection, intending to come back and fix it later. As is it works with 1.4, 1.2 and (IIRC) 360k drives on physical HW. I'm using it with QEMU daily - mostly 1.4M format and QEMU reports 1.4M format on boot (see below). IIRC I've also used 1.2M images w/o problems on QEMU, so I'm unsure what may be causing the symptoms you're seeing.

The intention is to come back and create a robust mechanism for floppy detection, including getting rid of the mess you mention, which is intensely confusing. Whether to keep the CMOS query or not - I haven't decided, but I'm leaning towards keeping it. For simplicity and since every conceivable system TLVC will be running on has it. I do have all variants except the 2880 physically, so it will be a ball. By the way, I'm not entirely happy with the floppy init code, so there is more to be done.

BTW (II) I agree with you on the fsync() comment. That's a bug.

```
12 L1 buffers, 12288B ram
eth: ne0 at 0x300, irq 12, (ne2k) MAC 52:54:00:12:34:56, flags 0x80
0040 003F 0000 0010 7E00 0200 003F 0000 0000 0000 4551 554D 495F 4544 2020 2020 
2020 2020 2020 2020 0003 0200 0004 322E 352B 2020 2020 5145 4D55 2048 4152 4444 
4953 4B20 2020 2020 2020 2020 2020 2020 2020 2020 2020 2020 2020 2020 2020 8010 
0001 0B00 0000 0200 0200 0007 003F 0010 003F F810 0000 0110 F810 0000 0007 0007 
IDE CHS: 63/16/63 serial# QEMU_IDE            
athd0: Multisector I/O, max 16 sects
0040 020B 0000 0004 5000 0200 0028 0000 0000 0000 4551 554D 495F 4544 2020 2020 
2020 2020 2020 2020 0003 0200 0004 322E 352B 2020 2020 5145 4D55 2048 4152 4444 
4953 4B20 2020 2020 2020 2020 2020 2020 2020 2020 2020 2020 2020 2020 2020 8010 
0001 0B00 0000 0200 0200 0007 020B 0004 0028 46E0 0001 0110 46E0 0001 0007 0007 
IDE CHS: 523/4/40 serial# QEMU_IDE            
athd1: Multisector I/O, max 16 sects
athd: found 2 hard drives
athd: /dev/dhda: 16 heads, 63 cylinders, 63 sectors (~31MB)
athd: /dev/dhdb: 4 heads, 523 cylinders, 40 sectors (~41MB)
Floppy drive(s) [CMOS]: df0 is 1.44M (4)
df: Direct floppy driver, FDC (82077) @ irq 6, DMA 2
```

Just for the heck of it I added a 1200k image to `qemu.sh` using `-fdb` and it does recognize it fine:
```
Floppy drive(s) [CMOS]: df0 is 1.44M (4), df1 is 1.2M (2)
```

ghaerr:
> Just for the heck of it I added a 1200k image to qemu.sh using -fdb and it does recognize it fine:

I just re-tested with 1440k the latest between both our repos and that works, but attaching 360k to QEMU still reports the drive as 1200k and I/O fails. Not sure why. When I attach a 720k image, the emulated CMOS is reporting 1440k. Perhaps this has something to do with 1 vs 2 heads not being recognized by QEMU?  It would be interesting to see if that happens on yours also? 

> The intention is to come back and create a robust mechanism for floppy detection, including getting rid of the mess you mention

Great, good to know. I won't try further cleanups of the original old driver source crud then. BTW, I removed BDEV_SIZE_CHK throughout ELKS as it isn't feasible to implement at the `ll_rw_blk` layer, and was a bunch of otherwise unusable ugly code. The tradeoff is that each block driver should check for whether `start` is larger than the max LBA sector and report an error and `end_request(0)`. All current block drivers already do that (BIOS, SSD, RD, etc) so you might consider doing that instead and delete BDEV_SIZE_CHK also (the code is also incompatible with the ELKS/TVLC partition sizing code).

Back to CMOS: it would be nice to have a way that a user can set the drive types, similar to CONFIG_BLK_DEV_BFD_HARD, so that the direct driver can be used without worries according to his hardware setup with no probing requirement. OTOH, perhaps that's an issue regardless if someone sticks a 360k floppy in a 1440k drive. You've spent a lot more time thinking about all this I'm sure!

I started on a breakup of the BIOS driver into separate files hoping that some of the routines could be re-used, but then realized that they likely can't due to the completely different nature of sync vs async I/O programming. Thus, I suppose hoping to reuse the BIOS driver floppy probing functions probably won't work well either.

Mellvik:
> > Just for the heck of it I added a 1200k image to qemu.sh using -fdb and it does recognize it fine:
> 
> I just re-tested with 1440k the latest between both our repos and that works, but attaching 360k to QEMU still reports the drive as 1200k and I/O fails. Not sure why. When I attach a 720k image, the emulated CMOS is reporting 1440k. Perhaps this has something to do with 1 vs 2 heads not being recognized by QEMU? It would be interesting to see if that happens on yours also?

Interestingly, QEMUS does not (just) look at the size of the image to determine what to emulate. I attached a 360 image and like you say, it does report (CMOS) 1.2M, but it works just fine. I suspect the same is the case with 720k (reporting 1.44M), I haven't tested that. All these floppies have 2 heads.

> 
> > The intention is to come back and create a robust mechanism for floppy detection, including getting rid of the mess you mention
> 
> Great, good to know. I won't try further cleanups of the original old driver source crud then. BTW, I removed BDEV_SIZE_CHK throughout ELKS as it isn't feasible to implement at the `ll_rw_blk` layer, and was a bunch of otherwise unusable ugly code. The tradeoff is that each block driver should check for whether `start` is larger than the max LBA sector and report an error and `end_request(0)`. All current block drivers already do that (BIOS, SSD, RD, etc) so you might consider doing that instead and delete BDEV_SIZE_CHK also (the code is also incompatible with the ELKS/TVLC partition sizing code).

Thanks, yes, that's definitely on the list!

> 
> Back to CMOS: it would be nice to have a way that a user can set the drive types, similar to CONFIG_BLK_DEV_BFD_HARD, so that the direct driver can be used without worries according to his hardware setup with no probing requirement. OTOH, perhaps that's an issue regardless if someone sticks a 360k floppy in a 1440k drive. You've spent a lot more time thinking about all this I'm sure!

The thing is (and this is why I'm hesitating to put any effort into this) is that most PCs (maybe all) have a CMOS to hold config data (I'm curious as to how to configure the original PC and XT CMOS - I owned one at the time, but don't remember). My thinking is that it's more flexible to set the drive in the CMOS than to hard configure the kernel with some value. Certainly, with the oldest of systems, such as my compaqs, one has to boot dos, start the setup program etc. in order to change the values, which - depending on the setting - may be more cumbersome than configuring the kernel... So my thinking is to keep the CMOS variant for now and let the driver probe (down) from there.

> 
> I started on a breakup of the BIOS driver into separate files hoping that some of the routines could be re-used, but then realized that they likely can't due to the completely different nature of sync vs async I/O programming. Thus, I suppose hoping to reuse the BIOS driver floppy probing functions probably won't work well either.

I think you're right about that one. Actually I had sort of the same discovery when I started making the IDE driver interruptdriven/asynchronous (in the works). It's a very different way of thinking, partitioning functions in new (and completely event driven) ways. That's the fun part though - and having been through the hoops with the floppy driver helps. Je..s what a mess.

Mellvik:
> I'll keep playing around with it. Although I just tested again and it works, I have seen kernel I/O errors using QEMU when using a non-matching base_type, possibly for the last reason stated above not working.

That's good. I'll look into the details of that mess when I redo the detection regime...
> 
> > it's more flexible to set the drive in the CMOS than to hard configure the kernel with some value. Certainly, with the oldest of systems
> 
> Another issue here to consider is configuring the kernel and driver with the drive format wanted even for emulation with QEMU, for testing. Agreed editing a kernel source file isn't probably the way, and possibly not adding more configuration options.
> 
> Something I'm considering is using the BIOS INT 13 AH=8 (Get Disk Parms) call to configure the disk system (used now for the BIOSHD driver but could easily work for DF driver as well). I started thinking that getting rid of all BIOS calls, even for initialization, isn't really buying much, other than an ideology, since the PC boot has to use the BIOS anyways. So I'm considering the idea of practicality - using the same mechanism to setup DF, in which case the disk "configuration setup" remains entirely consistent, regardless of disk "driver" used. The only real use of the direct FD driver is for async I/O, other than that, why not remain compatible?

My thinking is partly practical, partly ideological on the issue. The practical being 'CMOS is there, it's data and it's required for the system to operate, let's use it'. And BIOS - I'd like to pretend it doesn't exist. Instead , consider the bootstrap code  just that, some builtin software to read single sectors of some device. Just enough to get the system going.
If we got really pedantic about it, we'd take over after reading the boot sector, but that seems like a waste.
Ignoring the BIOS completely has the advantage of not worrying about data/device tables in memory etc.

> 
> Along those lines, I noticed that when running the command `mount -o remount,ro /dev/xxx /` the `/dev/xxx` is entirely ignored - instead the root filesystem is either remounted read-only (or nothing, in the case of `-o remount,rw`). So the idea is to enhance the kernel to allow remounting root onto a completely different block device, closing the former.
> 
> Thus, the system could, say, boot from a known working BIOSHD driver, then execute `mount -o remount,rw /dev/df0 /`, which would mount root read-only on the existing block device, then open the DF device, mount root on it and close the older block device. In this way, users could guaranteed boot up from the reliable BIOS driver then switch to the async DF driver for testing. And switch back if they wanted. This could work for hard disk drivers as well. Kinda cool I think.
> 
> In order to get that to work, I will have to move the `floppy_init` code into the `floppy_open` code, as right now the init code grabs the floppy IRQ and starts talking to the FDC before being opened. This means that one can't even link in the DF driver into a kernel and use the BIOS driver. So I think I'm going to fix that on ELKS.

I'm missing something here because I don't see how this is useful. Or rather, the useful part of this is already implemented AFAIK.  I've been using the `root=` directive in `/bootopts` a lot when working on the direct drivers and it has been very useful. So the point I'm missing is what this will get us.

What would have been really useful (and maybe that's the point) is the ability to mount a file as a file system (loopback file system).

Admittedly - some times usefulness is not the parameter, but rather the challenge and to see it work. Like the idea of adding swapping to TLVC...

PR #16, opened 2023-06-06
Another megaupdate - focus on direct block IO drivers

Mellvik:
This PR includes a number of smaller fixes, including an (almost) full revert of the previous 'fix' (simplification) to the timer routines in the scheduler. It turns out I had misunderstood the code and introduced some hard to track down problems with my simplification. The code is now almost back to where it was, while keeping the original bug fix (`NULL` initialization of a pointer). Also, and just as important, a bug in the execve implementation has been fixed.

The floppy driver has had a number of small fixes and now seems stable and ready for optimizations. The direct HD driver has had minimal adjustments since the  previous PR.

- `/bin/ash` has its heap size increased by approximately 200 bytes. The internal `test`command would consistently fail with `sbrk` errors. This is unsurprising given the extra memory used by the recently introduced history mechanism.
- `/bin/ash` has an updated set of (debug) trace points and trace content added while searching for the execve problem (below)
- The minix bootblock would recognize a deleted copy of linux and try to load it. This has been fixed.
- The kernel setup code would report `/bootopts` as malformed even if no `bootopts` file existed. This has been fixed.
- Since the previous PR, selecting BIOS based block IO didn't compile. This has been fixed.
- An elusive bug in the `execve()` system call has been fixed. Long story short (and a long search completed): The interrupt driven nature of the direct floppy driver and the slow speed of the floppy medium means that read requests queue up - in no particular order (at this time) and may be coming from different processes. When commands are read from the floppy and connected by a `pipe`, and if the first (source) command is much bigger than the second (receiver), the two will be loading concurrently and the latter (smaller) will complete first. A seemingly endless search commenced looking for the problem -  passing through the pipe code, a lot of `bin/ash` code and eventually ending in the `execve()` system call, which in this case is being reentered. The implementation was not reentrant and thus caused all kinds of weird behaviour. The fixed `sys_execve()` is completely reentrant (and there was little reason for it not being so in the first place - except keeping the allocation of 56 bytes per entry off the stack).
- A direct floppy driver bug caused floppy mounts just after QEMU startup to fail in the first attempt, then succeed. Adding a `SENSEI` command to the `shaker_done` routine fixed the problem (resetting  and thus re-enabling interrupts). This problem never occurred on physical hardware. Since it only appeared the first time after start (not after reboot) on QEMU, it seems like a QEMU bug.

There are still a few debug statements compiled into the direct floppy code, related to floppy writes - while the hunt for optimisations continues.

PR #25, opened 2023-09-21
Another major update - direct drivers and buffer system

Mellvik:
This PR has a number of important bugfixes, enhancements and cleanups on all levels - from applications to drivers - plus the buffer system. The following summary covers most, but not all.

- The direct drivers - the floppy driver with its very asynchronous behavior in particular - exposed serious problems in the buffer system - as mentioned in the previous 'major' PR: #19 (https://github.com/Mellvik/TLVC/pull/19#issue-1820158608). The subsequent bug hunt was complicated by the fact that the symptoms were timing sensitive and could not be replicated on QEMU. The hunt is detailed in https://github.com/Mellvik/TLVC/pull/16#issuecomment-1666959322
- When the exact problems had been identified, @ghaerr from the ELKS project stepped in to help with fixes and enhancements. The buffer code is identical between TLVC and ELKS, so the results benefit both projects equally. A couple of synchronization statements in the right places solved the main problem.
- The work on buffers led to the discovery of more bugs and problems/inefficiencies with  interesting consequences:
    - Excessive datacopying between the L1 cache and L2 buffers was discovered and located to the file system code. And subsequently eliminated in both Minix and FAT FS code. On the same note, the initialization of newly allocated buffers to zero has been changed and no longer requires the use of the L1 cache and the extra data copying back and forth. This change does not apply to XMS buffers on machines using INT15 for XMS access, which still need to pass thru L1.
    - File system metadata - superblock, bitmaps and inodes were previously locked in L1 cache at mount time and never released, which severely limited the number of full size Minix filesystems that could be mounted - and reduced the usefulness of the buffer cache. A floppy-booted system would in fact hang if such mount was attempted. This 'reservation' of cache space has been eliminated and @ghaerr added code to release cache space safely as part of the regular buffer admin regime. The jury is still out as to whether some stickiness is advantageous for the metadata buffers.
    - The much more efficient use of cache space changed the overall behavior of the buffer system significantly and mandated a serious look at what sizes are optimal for various configurations. A new `bootops` directive `cache=` now sets the size of the L1 cache, making it easier to experiment and run benchmarks with different cache sizes. The memory allocation for the cache is now done at runtime instead of compile time, making the `cache` setting real in terms of resources. The default # of L1 cache blocks is  set in `buffers.c`, `NR_MAPBUFS `, usually 12 or 8.
    - The `bufs=` directive works as before, setting the  number of L2 buffers to use - whether xms or 'external'. Running the system on L1 buffers only is still possible. Practical advice on optimal cache sizes will be posten in the Wiki.
    - During the bug hunt @ghaerr developed a collection of tracing tools, some of which have been ported to TLVC, in particular the buffer status display, triggered by the ^O key and the inode status display, triggered by ^N - if compiled in. `include/linuxmt/trace.h` is the place to look for 'switches' to turn on and off various traces. A trace tools guide will be posted in the developer notes Wiki. The new syscall trace and stack trace facilities have not been tested on TLVC.

# Other fixes & enhancements

- The direct HD driver has been substantially rewritten using experience from hundreds of hours of testing with many different physical (and QEMU) drives, from the 80s, 90s and newer, including CF cards. A number of bugs and inefficiencies have been removed and the code significantly simplified, preparing for an upcoming interrupt driven version. The driver now works with XMS_buffers.
    - The most important change to the HD drivers is the rewrite of the physical IO part, which attempted to take advantage of the IDE/ATA drives' multiIO feature. While this worked,  it did not take advantage of many of the features ATA provides, such as implied seeks and the ability to completely ignore CHS boundaries (which the drives handle automatically).
    - Several adjustments were required to get the driver to work reliably with a 1986 model 40MB vintage IDE drive on a slow 80286 machine. Many of these adjustments (mainly delay loops and status register checks) will be eliminated when the driver goes asynch/interrutpt driven later.
- A lot of inactive code has been removed.
- Various problems related to conditional compile/config variations have been fixed.
- The floppy formatting code has been 'ifdefed' from the driver, reducing the ram footprint.
- A number of bugs have been eliminated from the direct fd driver, including a 64k boundary alignment problem which revealed itself by the changes to memory allocation for the L1 buffers mentioned above.
- The CONFIG option to turn off the floppy read cache has been removed.

### DIFF
--- a/tlvc/arch/i86/drivers/block/directhd.c
+++ b/tlvc/arch/i86/drivers/block/directhd.c
@@ -64,6 +64,7 @@
 
 #define STATUS(port) inb_p((port) + ATA_STATUS)
 #define ERROR(port) inb_p((port) + ATA_ERROR)
+#define SECTOR(port) inb_p((port) + ATA_SECTOR)
 #define WAITING(port) ((STATUS(port) & BUSY_STAT) == BUSY_STAT)
 #define DRQ_WAIT(port) (STATUS(port) & DRQ_STAT) /* set when ready to transfer */
 
@@ -77,7 +78,8 @@
 __asm__("cld;rep;outsw"::"d" (port),"S" (buf),"c" (nr))
 
 
-#define USE_MULTISECT_IO	/* Enable multisector R/W */
+#define USE_MULTISECT_IO	/* Enable/disable multisector R/W - for debugging */
+//#define USE_INTERRUPTS		/* EXPERIMENTAL - test interupts */
 #define OLD_IDE_DELAY 1000	/* delay required for old drives, system dependent */
 #define DEBUG
 
@@ -132,6 +134,7 @@ static int  directhd_sizes[MAX_ATA_DRIVES << MINOR_SHIFT] = { 0, };
 static void directhd_geninit();
 static void reset_controller(int);
 static int drive_busy(int);
+static void directhd_int(int, struct pt_regs *);
 
 static struct gendisk directhd_gendisk = {
     MAJOR_NR,			/* major: major number */
@@ -170,11 +173,6 @@ void directhd_geninit(void)
     return;
 }
 
-/*
- * FIXME Maybe we can merge all this stuff into one piece of code thet handles 
- * all conditions? With the possible exception of moving data into XMS buffers...
- */
-
 /* assumes current data segment - which is kernel_ds */
 void insw(unsigned int port, word_t *buffer, int count)
 {
@@ -190,70 +188,23 @@ void read_data(unsigned int port, ramdesc_t seg, word_t *buffer, int count, int 
 {
 
 #if defined(CONFIG_FS_XMS_BUFFER) || defined(USE_LOCALBUF) /* use bounce buffer */
-    if (!raw) {				// FIXME: Fix indentation
-    while (count) {
-	int chars = (count > BLOCK_SIZE)? BLOCK_SIZE : count;
-
-	insw(port, (word_t *)localbuf, chars);
-	printk("insw %d %04x %04x %04x;", chars, localbuf, buffer, *(word_t *)localbuf);
-#ifdef CONFIG_FS_XMS_BUFFER
-	xms_fmemcpyw(buffer, seg, localbuf, kernel_ds, chars/2);
-#else
-	fmemcpyw(buffer, seg, localbuf, kernel_ds, chars/2);
-#endif
-	count -= chars;
-	buffer += chars/2; 
-    }
+    if (!raw) {	
+	insw(port, (word_t *)localbuf, count);
+	//printk("insw %d %04x %04x %04x;", count, localbuf, buffer, *(word_t *)localbuf);
+	xms_fmemcpyw(buffer, seg, localbuf, kernel_ds, count/2);
     } else
 #endif
     {
 
-    unsigned int __far *locbuf = _MK_FP(seg, (unsigned)buffer);
+	unsigned int __far *locbuf = _MK_FP(seg, (unsigned)buffer);
 
-    count >>= 1;	/* bytes -> words */
-    //printk("%x,%x,%x,%lx,%d;", port, buffer, seg, locbuf, count);
-    do {
-	*locbuf++ = inw(port);
-    } while (--count);
+	count >>= 1;	/* bytes -> words */
+	//printk("%x,%x,%x,%lx,%d;", port, buffer, seg, locbuf, count);
+	do {
+	    *locbuf++ = inw(port);
+	} while (--count);
     }
-    
 }
-
-#if 0
-
-/* this _should_ work on an 8088/86 now, but I haven't tested it yet */
-
-void insw(unsigned int port,unsigned int *buffer,int count)
-{
-#asm
-	push	bp
-	mov	bp, sp
-	push	cx
-	push	di
-	push	dx
-	mov	dx, [bp + 04]
-	mov	di, [bp + 06]
-	mov	cx, [bp + 08]
-	shr	cx, 1
-	cld
-#ifndef CONFIG_PC_XT
-	repz
-	insw
-#else
-			/* this should work on an 8088 */
-dhd_insw_loop:
-	in	ax, dx
-	stosw
-	loop	dhd_insw_loop
-#endif
-	pop	dx
-	pop	di
-	pop	cx
-	pop	bp
-#endasm
-}
-
-#endif
 
 #if defined(USE_LOCALBUF) || defined(CONFIG_FS_XMS_BUFFER)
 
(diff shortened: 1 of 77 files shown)


tlvc/arch/i86/drivers/block/ll_rw_blk.c
```diff
@@ -154,21 +139,28 @@ static void add_request(struct blk_dev_struct *dev, struct request *req)
 {
     register struct request *tmp;
 
-    clr_irq();
+    //clr_irq();
```
ghaerr, line 142: Not sure why disabling interrupts was removed here: this introduces a big-time race condition, where the following line `tmp = dev->current_request` is checked for NULL, and `end_request` called at interrupt time from the direct floppy driver executes `CURRENT = req->rq_next` in blk.h which writes `dev->current_request`.

When the race condition causes an error, what happens is either the next pending I/O request in the queue or the next request to be added in `make_request` is dropped or the request queue becomes corrupted.

tlvc/arch/i86/drivers/block/ll_rw_blk.c
```diff
@@ -181,27 +173,14 @@ static void make_request(unsigned short major, int rw, struct buffer_head *bh)
...
-	blks = (sector_t)(ebh->b_count & 0x7f); /* # of sectors */
-	ebh->b_count = 1;
-    }
-#endif
 #ifdef BDEV_SIZE_CHK	/* NOTE: not updated for raw IO */
     sector_t count = BLOCK_SIZE / FIXED_SECTOR_SIZE;	/* FIXME must move to lower level*/
     sector_t sector = buffer_blocknr(bh) * count;
     if (blk_size[major])
-	if (blk_size[major][MINOR(buffer_dev(bh)&0x7fff)] < (sector + count) >> 1) {
+	if (blk_size[major][MINOR(buffer_dev(bh))] < (sector + count) >> 1) {
```
ghaerr, line 183: Just an FYI that the `blk_size` code is a bit messed up along with BDEV_SIZE_CHK and should be considered for removal. Note that `blk_size` is declared `int` (a carryover from Linux where sizeof(int) == 32) and count and sector are sector_t (long). Any drive over 65Mb will cause integer overflow. The mechanism also requires drivers to allocate a blk_size for each partition on each minor drive, using excess storage as you previously brought up as a problem with the BIOSHD driver and its MINOR_SHIFT == 5.

Finally, there's a major bug in the genhd.c extended partition recognition code if the `size` member of struct gendisk isn't initialized (line 262). That struct member is also an `int` and thus useless for large partitions. At a bare minimum genhd.c's line 262 should also be ifdef BDEV_SIZE_CHK.

I ended up deleting all of it and saving kernel data space in ELKS (along with also changing BIOSHD MINOR_SHIFT to 3 which freed up 1K bytes kernel data, thanks to your commenting about it in the source some time ago).

tlvc/arch/i86/drivers/block/ll_rw_blk.c
```diff
@@ -261,12 +247,13 @@ static void make_request(unsigned short major, int rw, struct buffer_head *bh)
     }
 
     /* fill up the request-info, and add it to the queue */
+    clr_irq();
```
ghaerr, line 250: Conversely to the comment above, disabling interrupts isn't required here because `get_request`, which is always called with interrupts disabled, sets `req->rq_status = RQ_ACTIVE` after selecting an RQ_INACTIVE request. This guarantees the request won't be found in any driver queue and thus not referenced by an interrupt routine.

ghaerr:
> The CONFIG option to turn off the floppy read cache has been removed.

I didn't see this change - is this in `config.in`, or in the direct floppy driver? I assume the idea that the direct floppy driver is far better off always keeping a track buffer. If the option was removed in config.in, then it will be important to also adjust the ifdefs in linuxmt/config.h so that the track buffer is always allocated following DMASEG.

tlvc/arch/i86/drivers/block/ll_rw_blk.c
Mellvik: Thanks @ghaerr - great catch!!!
It was removed because `add_request()` was called with interrupts disabled for an experimental period, then never fixed.

tlvc/arch/i86/drivers/block/ll_rw_blk.c
Mellvik: Thanks - and yes, the BDEV_SIZE_CHK stuff has never been used in TLVC and is on its way out (the `&0x7fff` was a bulk change as part of an experiment, now reverted). Which means there is no size check at this time. To be fixed.

The extended partition code has never been used (in TLVC) and I've wanted to get rid of it. Thanks for the hint - it's now ifdef'ed.

The MINOR_SHIFT reduction is on my list - thanks for the reminder.

tlvc/arch/i86/drivers/block/ll_rw_blk.c
Mellvik: Thanks, good point!

Mellvik:
Again, great catch - I ended up removing it from the config files without completing the job. In addition, the config files were left out from the PR.

Thanks.

tlvc/arch/i86/drivers/block/directhd.c
```diff
@@ -678,7 +597,8 @@ void do_directhd_request(void)
 
 	if (directhd_initialized != 1) {
 	    end_request(0);
-	    continue;
+	    //continue;		/* no reason to continue */
+	    return;
```
ghaerr, line 601: As per my comment in https://github.com/Mellvik/TLVC/pull/95#issuecomment-2430080543, this is a good example of an illegal action in an async I/O interrupt handler. The problem here is that if execution returns (instead of continues) here, and the request queue is non-empty, nothing will ever restart any queued I/O request, and add_request will only continue to queue I/O, never calling the request function directly as it does when the queue is empty. The driver will effective hang and never continue any existing nor future I/O request, even if a later open succeeds.

In the above case, the system will just hang by sleeping all processes waiting for I/O on this device, permanently when the I/O queue max size is reached, with no indication of why the failure is occurring. Continuing would at least empty the queue, so that the I/O request subsystem continues to function. Of course, all this begs the question of how/why an I/O request can be submitted when the driver open routine failed. IMO it can't. So that begs the question of why check for driver initialization in the request routine. IMO there's no point in it, that's what the _open routine return value is for, and the buffer system will never submit a block I/O request on an unopened device. So it seems the most correct action is to discard the initialization check and don't return or continue!  If the driver can't complete the request for any reason, a later call to end_request(0) I'm sure is already coded somewhere.


FYI.

Note that the current DF driver has no such initialized check.

Mellvik: Thanks for the code review and the comment @ghaerr. This is not an async driver - and turning it into one is quite a bit down on the list because there is minimal gain, possibly the opposite because everything it PIO and consuming cycles anyway.

PR #26, opened 2023-09-22
[kernel] Fixes and cleanups

Mellvik:
Cleanups and fixes related to PR #25 - @ghaerr, thanks for the code review.

The RAW directhd driver now works - it needed some rather ugly patches now that the `b_seg` element of the buffer head structure is gone. 

The `CONFIG_TRACK_CACHE` `#ifdef` is now completely gone.


ghaerr:
Out of curiosity, have you ever seen or used an HD with extended (MSDOS) partitions? I notice support has been removed for that. 

Not only does the extended partition code seem a bit large for its non-use, but keeping 8 minor devices per drive instead of 4 seems excessive too, for something that may never be used. Removing extended partition support permanently would also allow for changing MINOR_SHIFT to 2 (from 6 on TLVC). On ELKS, over 1k bytes kernel data was saved just moving MINOR_SHIFT from 6 to 3, along with some slight restructuring - well worth it, given our problems with kernel data limits.

I had thought of removing support for extended partitions too, but was unsure as to whether there were many of them out there in the wild. What's your take? When did extended partitions come into use? I thought they were used a lot at one time, when MBR was the only boot sector format supported (before GUID, etc).

Mellvik:
Well, yes I have even used them, but I completely forget why and for what purpose, and consider it completely useless today. 

My vague recollection is that in the early 90s when this was written, we were probably trying to get old filesystems to work well with hiiiiigh capacity drives - while avoiding huge block sizes (smaller filesystems, more partitions). now they're simply a useless complication.

i have in my list to change the minor shift, there's indeed a lot to save. Thanks.

PR #27, opened 2023-09-23
[cmd] Added medium size check to mkfs

Mellvik:
`mkfs` used to accept any oil system size up to the maximum # of blocks for Minix1 file systems.

With this update, `mkfs` checks the size of the media by different means to ensure the request is sane. 

Also added is a means to dry-run the `mkfs` command: Add a '+' in front of  the block size and `mkfs` will do its job but not write anything.

```
tlvc15# mkfs /dev/dhda +31752
10584 inodes
31752 blocks
Firstdatazone=339 (339)
Zonesize=1024
Maxsize=32514048

mkfs: No data written to device
tlvc15#

tlvc15# mkfs /dev/dhda 31753
mkfs: Requested size larger than device
tlvc15#

```

### DIFF
--- a/tlvccmd/disk_utils/mkfs.c
+++ b/tlvccmd/disk_utils/mkfs.c
@@ -157,13 +157,13 @@ unsigned char test_bit(unsigned int nr, void * add)
  * to compile this under minix, volatile gives a warning, as
  * exit() isn't defined as volatile under minix.
  */
-volatile void fatal_error(const char * fmt_string,int status)
+volatile void fatal_error(const char * fmt_string, int status)
 {
 	printf(fmt_string);
 	exit(status);
 }
 
-#define usage() fatal_error("Usage: mkfs /dev/name blocks (Max blocks=65535)\n",16)
+#define usage() fatal_error("Usage: mkfs /dev/name [+]blocks (Max blocks=65535)\n",16)
 #define die(str) fatal_error("mkfs: " str "\n",8)
 
 void write_tables(void)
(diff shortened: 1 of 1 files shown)


tlvccmd/disk_utils/mkfs.c
```diff
@@ -309,10 +314,22 @@ int main(int argc, char ** argv)
 		die("unable to open device");
 	if (fstat(DEV,&statbuf)<0)
 		die("unable to stat %s");
-	//else if (statbuf.st_rdev == 0x0300 || statbuf.st_rdev == 0x0340)
-		//die("Will not try to make filesystem on '%s'");
-
+	if (!(buf = malloc(512))) 
+		die("Cannot malloc buffer memory"); 
+	end = lseek(DEV, 0, SEEK_END);
```
ghaerr, line 319: Instead of `lseek`, why not just use `statbuf.st_size` which has the same value from above?

Also, why is an additional `lseek` and actual `read` required below, rather than just believing the block device inode size reported? It would seem to me either that the inode size (set by the driver open routine) should be enforced by the driver `do_request` function anyways? 

I know that presently there may be the case that not all block drivers have the max size check (comparing req->rq_sector with max sectors) present in the block driver.... but if that's the case then the `read` below would succeed anyways.

Mellvik: Thanks @ghaerr,
Statbuf.st_size - of course, I forgot. 

The rest is trial and error. The extra lseek and read is insurance, the max (SEEK_END) turns out to be unreliable (for now) (floppies).

Of course this should (and will) be fixed driver level - for now this helps avoiding mistakes. And mkfs becomes the litmus test when the low level supposedly works.

ghaerr:
> The rest is trial and error. The extra lseek and read is insurance

I'm not sure about "insurance", but I can now see why a `read` might be required given the current driver design problem (see below). The block driver "max size" is stored in only one place (`inode->i_size`), and the `lseek` uses this information for SEEK_END to return an error or not.

Looking at the driver, I see the problem: `inode->i_size = floppy->size` is set at open time, but this is before any auto probing. So the block device size is set to the initial base_type[] entry from the CMOS drive type. When the driver is finally given its first read/write request through `do_fd_request`, the driver will try probing (since `current_type[] == NULL`) and reset the global `floppy` to the correct floppy type, but doesn't set `inode->i_size` again. The fix is a bit complicated since `inode` isn't available at I/O time, so either a copy has to be made (could work but its a hack), or probing needs to occur at open time (this changes the current probing design somewhat heavily).

Thus, given the current design, a `read` is required to force a probe after which the driver could set `save_inode->i_size` for retrieval with a later `stat`. Executing a `read` past end of media could display a kernel I/O error message which `mkfs` wouldn't know about, then get a -1 read return. However, the read doesn't need to be from the end of media (just read from position 0), and it doesn't need to be 512 bytes. Last I checked, `mkfs` was extremely short on heap space for filesystems with large inode counts, so you might consider reading into a very small non-malloc'd buffer instead of calling `malloc`.

Mellvik:
> 
> Thus, given the current design, a `read` is required to force a probe after which the driver could set `save_inode->i_size` for retrieval with a later `stat`. Executing a `read` past end of media could display a kernel I/O error message which `mkfs` wouldn't know about, then get a -1 read return. However, the read doesn't need to be from the end of media (just read from position 0), and it doesn't need to be 512 bytes. Last I checked, `mkfs` was extremely short on heap space for filesystems with large inode counts, so you might consider reading into a very small non-malloc'd buffer instead of calling `malloc`.

Thanks for the code review, @ghaerr - this is indeed a good point. 
The malloc is completely superfluous, I'll go ahead and change that!

PR #31, opened 2023-10-10
[directfd] Added raw access to direct floppy, cleanups and fixes

Mellvik:
This PR continues where the previous left off, with more updates to the `directed` driver, notably adding raw device access. More details:

- Floppy driver cleanups
- Added config option (`#define`) for full cylinder read buffering on small floppies, which may not always be desirable: Reading two tracks instead of one track means another ⅙ of a second to wait, which is advantageous for big file/device reads but may not average out in a positive way in general usage. (The full track buffer would be used by 360k and 720k floppies.)
- Raw/char device support for floppies. The raw devices are `/dev/rdf0, ...`.
- Removed the CONFIG_BLK_DEV_CHAR configuration option. Raw access is now always included with the direct block drivers and (soon to be) required by some of the system disk utilities, such as `disk`,  `mkfs`, `fsck`, `makeboot`… in the future possibly a `format` program for floppies. The code size implementing raw access is minimal.


tlvc/arch/i86/drivers/block/directhd.c
```diff
@@ -565,7 +560,7 @@ void directhd_release(struct inode *inode, struct file *filp)
 
     access_count[target]--;
```
ghaerr, line 561: Always decrementing the driver reference counter here but only incrementing it for block device opens will cause problems. Why is that being done, just to skip the invalidate routines on close?

Mellvik: Thanks for the code review, @ghaerr.

Actually, the block device close routine is never called from the raw driver. Not obvious since open is called. Then again there is nothing to do when closing a raw access.

-M

PR #35, opened 2023-12-08
[net] Experimental - the ne2k driver now uses heap_alloc() for buffers

Mellvik:
This PR (optionally) changes network driver I/O buffer allocation from static to using the kernel heap (`heap_alloc()`). Buffer configuration is still done in `netbuf.h` and the allocation scheme is easily switched between static (old) and heap (new) (`#define USE_HEAP_BUFFER`).

Only the `ne2k` driver fully implements this ability at this time, the other drivers are pending (more testing needed).

Heap buffer allocation can be tracked via the `meminfo` command which will reveal that the buffers are allocated individually, not in one (or two) chunks. For now this is the simplest. However tempting (and efficient) it may be to allocate one buffer and split it in the driver(s), the current scheme is more flexible and will allow (future) backtracking in case we run out of heap space while allocating.

In order to add more flexibility, a `bootopts` option to set the # of receive and transmit buffers (and their sizes) is attractive, not the least to investigate the relationship between buffers and speed.

### DIFF
--- a/tlvc/arch/i86/drivers/net/ne2k.c
+++ b/tlvc/arch/i86/drivers/net/ne2k.c
@@ -436,6 +436,13 @@ static int ne2k_open(struct inode *inode, struct file *file)
 		}
 		ne2k_reset();
 		ne2k_init();
+#if (NET_OBUFCNT > 0)
+		tnext = netbuf_init(net_obuf, NET_OBUFCNT);
+#endif
+#if (NET_IBUFCNT > 0)
+		rnext = netbuf_init(net_ibuf, NET_IBUFCNT);
+#endif
+
 		ne2k_start();
 	}
 	return 0;
@@ -449,6 +456,10 @@ static void ne2k_release(struct inode *inode, struct file *file)
 {
 	if (--usecount == 0) {
 		ne2k_stop();
+#ifdef USE_HEAP_BUFFER
+		netbuf_release(tnext);
+		netbuf_release(rnext);
+#endif
 		free_irq(net_irq);
 	}
 }
@@ -506,13 +517,6 @@ void INITPROC ne2k_drv_init(void)
 	mac_addr = (byte_t *)&netif_stat.mac_addr;
 
 	net_port = NET_PORT;    // ne2k-asm.S needs this.
-
-#if (NET_OBUFCNT > 0)	/* redo this if we add more than 2 buffers */
-	tnext = netbuf_init(net_obuf, NET_OBUFCNT);
-#endif
-#if (NET_IBUFCNT > 0)	/* redo this if we add more than 2 buffers */
-	rnext = netbuf_init(net_ibuf, NET_IBUFCNT);
-#endif
 
 	while (1) {
 		int j, k;
--- a/tlvc/arch/i86/drivers/net/netbuf.c
+++ b/tlvc/arch/i86/drivers/net/netbuf.c
@@ -3,13 +3,35 @@
  */
 
 #include "netbuf.h"
+#include <linuxmt/kernel.h>
+#include <linuxmt/heap.h>
 
-struct netbuf * netbuf_init(struct netbuf *buf, int cnt) {
+struct netbuf *netbuf_init(struct netbuf *buf, int cnt) {
 	int i;
 	for (i = 0; i < cnt; i++) {
+#ifdef USE_HEAP_BUFFER
+		buf[i].data = NULL;
+		if (!(buf[i].data = heap_alloc(MAX_PACKET_ETH, HEAP_TAG_BUFHEAD))) {
+			printk("eth: Buffer alloc failed\n");
+			netbuf_release(&buf[0]);
+			return(NULL);
+		}
+		printk("netbuf got %x\n", buf[i].data);
+#endif
 		buf[i].len = 0;
 		buf[i].next = &buf[i+1];
 	}
 	buf[--i].next = &buf[0];
 	return(&buf[0]);
 }
+
+#ifdef USE_HEAP_BUFFER
+void netbuf_release(struct netbuf *buf) {
+	struct netbuf *n = buf;
+	do {
+		heap_free(n->data);
+		printk("netbuf rel %x\n", n->data);
+		n = n->next;
+	} while (n != buf);
+}
+#endif
--- a/tlvc/arch/i86/drivers/net/netbuf.c
+++ b/tlvc/arch/i86/drivers/net/netbuf.c
@@ -10,7 +10,7 @@ struct netbuf *netbuf_init(struct netbuf *buf, int cnt) {
 	int i;
 	for (i = 0; i < cnt; i++) {
 #ifdef USE_HEAP_BUFFER
-		if (!(buf[i].data = heap_alloc(MAX_PACKET_ETH, HEAP_TAG_BUFHEAD))) {
+		if (!(buf[i].data = heap_alloc(MAX_PACKET_ETH, HEAP_TAG_NETWORK))) {
 			printk("eth: Buffer alloc failed\n");
 			netbuf_release(&buf[0]);
 			return(NULL);
--- a/tlvc/arch/i86/drivers/net/netbuf.c
+++ b/tlvc/arch/i86/drivers/net/netbuf.c
@@ -10,13 +10,12 @@ struct netbuf *netbuf_init(struct netbuf *buf, int cnt) {
 	int i;
 	for (i = 0; i < cnt; i++) {
 #ifdef USE_HEAP_BUFFER
-		buf[i].data = NULL;
 		if (!(buf[i].data = heap_alloc(MAX_PACKET_ETH, HEAP_TAG_BUFHEAD))) {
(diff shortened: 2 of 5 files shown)


ghaerr:
> a bootopts option to set the # of receive and transmit buffers (and their sizes) is attractive, not the least to investigate the relationship between buffers and speed.

Pretty interesting, really. I was thinking that if the transmit buffer size was increased, that would allow for potentially faster throughput since `ktcp` might not have to wait for the ethernet `write` to complete before returning and processing the next outgoing packet. But since this was all written prior to the kernel becoming async I/O capable, I'm not sure whether the kernel TCP driver or the NE2K driver operate async I/O, we'd have to check. In general, I would bet not.

Now I am wondering whether throughput would actually get increased or not by increasing the NE2K transmit buffers, except for short data-packet total transfers less than the combined buffer size, as `ktcp` would ultimately have to wait for either the outgoing window and/or slowly process the data anyways. Obviously a good candidate for testing.

For increased read throughput, my first guess would be, similar to async I/O reads, not as great an improvement, since ultimately the receive queue is also constrained by the incoming max window size. So also a good candidate for testing!

To move forward, the first step would be to look at your NE2K driver to see whether it's async I/O capable (and its a char driver, not block, so none of the block async I/O enhancements we did will help). Then the bigger issue would be whether the /dev/tcpdev kernel driver is async, and the answer there is definitely not (yet). This is because, for all the reasons we found several years ago, the kernel has to run lock-step with `ktcp` for processing various types of TCP state machine scenarios, and there's no easy way of converting `ktcp` itself to run async I/O. Nonetheless, the idea is interesting, perhaps async I/O isn't a bottleneck.

Mellvik:
It seems to me we've had this discussion before, but I couldn't find it.

> > a bootopts option to set the # of receive and transmit buffers (and their sizes) is attractive, not the least to investigate the relationship between buffers and speed.
> 
> Pretty interesting, really. I was thinking that if the transmit buffer size was increased, that would allow for potentially faster throughput since `ktcp` might not have to wait for the ethernet `write` to complete before returning and processing the next outgoing packet. But since this was all written prior to the kernel becoming async I/O capable, I'm not sure whether the kernel TCP driver or the NE2K driver operate async I/O, we'd have to check. In general, I would bet not.

I'm not following your reasoning here. Network I/O was always async/interrupt driven, adding driver-local buffers doesn't change that, but it does change the dynamics quite a bit, depending on which driver we look at.
If we stick to the NE2K case, the write - before local buffering - was just a transfer of data to the NIC buffer, then return. If a transfer was already in progress there would be a wait - which is where the local buffer helps. It reduces the chance of a wait-for-resource and allows for back-to-back outgoing packets, both of which speed up the transfer - if the machine is fast enough, but it doesn't change the already async nature of the IO. The read is similar except we get the benefit of the NIC buffer being larger. The driver local netbuf allows optimal usage of the bandwidth between the NIC and the kernel using interrupts instead of polling. There will still be sleeps and wakeups, but less frequent.
 
> 
> Now I am wondering whether throughput would actually get increased or not by increasing the NE2K transmit buffers, except for short data-packet total transfers less than the combined buffer size, as `ktcp` would ultimately have to wait for either the outgoing window and/or slowly process the data anyways. Obviously a good candidate for testing.

Yes, this is governed by the window handling code, not the IO. 

> 
> For increased read throughput, my first guess would be, similar to async I/O reads, not as great an improvement, since ultimately the receive queue is also constrained by the incoming max window size. So also a good candidate for testing!

Adding the proposed netbufs setting in /bootopts is becoming increasingly attractive :-)

> 
> To move forward, the first step would be to look at your NE2K driver to see whether it's async I/O capable (and its a char driver, not block, so none of the block async I/O enhancements we did will help). Then the bigger issue would be whether the /dev/tcpdev kernel driver is async, and the answer there is definitely not (yet). This is because, for all the reasons we found several years ago, the kernel has to run lock-step with `ktcp` for processing various types of TCP state machine scenarios, and there's no easy way of converting `ktcp` itself to run async I/O. Nonetheless, the idea is interesting, perhaps async I/O isn't a bottleneck.

This is where I get lost. Maybe you could explain what kind/level of asynchronicity is missing?

ghaerr:
> Maybe you could explain what kind/level of asynchronicity is missing?

Think of the way that block device async I/O works: various applications (via the filesystem) request blocks to be read or written; these are converted into I/O requests and queued. The block drivers themselves, after completing I/O disable interrupts and dequeue the next request, while signaling a wakeup for any tasks waiting for the block, etc.

This isn't really how "async" networking works in ELKS/TLVC: the /dev/tcpdev driver sits in between a network application, /bin/ktcp and the network driver. The /dev/tcpdev driver accepts only a single request at a time, and only uses one read and and one write buffer for all networking. Perhaps I didn't use the correct word, but requests are handled more synchronously throughout the application <-> ktcp <-> network driver process than a true "async" equivalent, like block I/O. This is because the traffic can't be handled exclusively in the kernel, but has to be threaded through /bin/ktcp etc as I described above.

> If we stick to the NE2K case, the write - before local buffering - was just a transfer of data to the NIC buffer, then return. If a transfer was already in progress there would be a wait - which is where the local buffer helps. It reduces the chance of a wait-for-resource and allows for back-to-back outgoing packets, both of which speed up the transfer - if the machine is fast enough, but it doesn't change the already async nature of the IO. The read is similar except we get the benefit of the NIC buffer being larger. The driver local netbuf allows optimal usage of the bandwidth between the NIC and the kernel using interrupts instead of polling. There will still be sleeps and wakeups, but less frequent.

Yes, agreed, this is the basic way the system works; I was just thinking along the lines of the other synchronous bottlenecks may be that might decrease the effectiveness of say, a large number of network buffers in the NIC driver. For instance, were we to have five read and five write buffers in the NIC driver, would that actually speed up application throughput, or would the system quickly get bogged down in the /dev/tcpdev driver that only has a single kernel buffer for all processes... ?

> Network I/O was always async/interrupt driven, adding driver-local buffers doesn't change that

True - but in summary, the NIC driver, although written to take advantage of NIC transmit and receive interrupts, doesn't really operate like our block I/O request queues - they can't select the next network request and immediately work on it - instead they're actually synchronously piping data to/from the kernel's single read/write network buffers. As always, it will be quite interesting to see the results of adding buffers. My guess would be some increase from 1 to 2, and much less adding a 3rd or more. Testing will tell :)

ghaerr:
> Admittedly I've never looked at the tcpdev code, I need to do something about that.

You won't like what you see - its complicated and somewhat ugly. It is/was the source of many issues when I first jumped into trying to understand and fix early ELKS networking problems. Its requirement of running lockstep with /bin/ktcp, especially when errors occur, is definitely complicated.

> When using network buffers, it seems to me these are exactly the same, the network buffers being the request queue. The NIC gets a completion interrupt, grabs the next request (buffer) and operates on it. If there is no data (read) or no buffer (write), then wait.

We'll see. Given that all network data has to pass through the single /dev/tcpdev kernel buffer (IIRC 1536 bytes), it will be interesting to see whether adding a single extra transmit buffer actually gets used. IMO, the NIC network buffers aren't similar to a request queue because the /dev/tcpdev driver does the read/write scheduling, synchronization, and additional buffering. I could be wrong about this in the particular case we're discussing though, because its actually /bin/ktcp that reads/writes directly to /dev/eth (synchronously only), and then passes the data back up to the application via /dev/tcpdev. Its definitely complicated. /bin/ktcp may need to be modified in order to get the increased throughput you're looking for.

> It's also worth remembering that when we look at network performance, we tend to always look at a single connection, which will be window/ack/retransmit buffer constrained. The picture becomes different if multiple connections are active, less 'pack-ack, pack-ack' oriented, less predictable. 

Agreed there. Single vs multiple application network performance could be entirely different.

> It runs on top of MTCP on the DOS side

So this would only really allow DOS systems to access ELKS/TLVC for external FAT images. Converting a DOS application for running on ELKS/TLVC probably won't be straightforward.


> It would be an interesting addition to TLVC and ELKS given that it folds into the already established FAT implementation, which I suspect it does.

So the Linux version is kind of an NFS server for FAT? In essence an SMB server but non-standard protocol? The server version could possibly be compiled for ELKS/TLVC and run as a daemon, I suppose, it is uses standard socket lib calls. I don't really see much use in allowing DOS systems to access ELKS/TLVC though.

Mellvik:
NetDrive:
This is a hunch only at this time. What I'm eying is a simple client for TLVC talking to a server on Linux. NFS light! Unfortunately FAT based, but the again, gotta start somewhere... :-)

ghaerr:
> What I'm eying is a simple client for TLVC talking to a server on Linux. NFS light! 

Hey if you really want to have fun, perhaps now that you know ELKS block I/O so well, just divvy out block read/write requests via the network over to the TLVC/MTCP server on Linux :)

Mellvik:
Sounds like a party to me 2. I'll put a toe into the water first by having a closer look and test on DOS.  This Brutman guy is doing some really great stuff and he's been at it for a looooong time.

Mellvik:
Turns out the implementation uses UDP. Elegant simplicity and some really good ideas. To be revisited when we have UDP...

PR #51, opened 2024-03-23
[kernel] Fix elusive scheduler bug

Mellvik:
An interesting rabbit hole revealed itself while testing the new (PR forthcoming) MFM disk driver (aka xd.c). As the debugging `printk's` were deleted, the driver suddenly stopped working - that is, it crashed the system in the first read operation (read the partition table). More testing and it turned out that keeping a printk at the end of the io-request (`do_xd_request()`) function eliminated the problem. It definitely had the signature of a race condition/stack corruption/memory pollution issue.

The flow of operation of a TLVC IO request is that a request is passed down through the various levels, ends up in the io-request function - directly or via the request queue (this being an interrupt driven driver) if the drive is busy. The io-request function starts the actual IO, returns immediately and the the call ends up in `wait_on_buffer()` while waiting for the actual IO to complete. 

In this case (the first read since boot) there is no queue (and no processes actually) and the read is initiated right away. Now, since the machine (4.77MHz 8088 PC/XT) is very slow and the DMA IO relatively fast, the completion interrupt may occur before the `wait_on_buffer()` has completed - or even called. In this case, the call to `wait_on_buffer()` finds the buffer already unlocked and returns immediately with IO completed. This is exactly what happens if the return from the driver is slightly delayed - such as by a call to `printk` or even a short delay loop: The interrupt processing completes before  `wait_on_buffer()` is called and everything works fine.

Without the extra delay, `wait_on_buffer()` proceeds and calls `schedule()` to let the system do other stuff while waiting. All good - except that `schedule()` isn't prepared for this particular situation. At this point, the only running process is the `idle_task`, task 0, which is where the system spends its time when there is nothing to do. A task that - per definition - is never really running and thus never has the `RUNNING` flag set.

This is where the proverbial rabbit is hiding. When attempting to switch to the next task, `schedule()` finds that the current task isn't running (no `RUNNING` flag) which means it must be put to sleep - by calling `del_from_runqueue()`. 
Rabbit identified: `del_from_runqueue()` has all sanity checks commented out and does not detect that it's attempting to sleep task 0. It goes ahead and sets the `next` and `prev` task pointers to NULL. On return, the obvious happens in `schedule()`: Instead of task-switching to itself, it task-switches to a NULL pointer and the system is toast.

The simple fix (tiny rabbit indeed) is reenabling the last half of the previously disabled sanity check in `del_from_runqueue()`.

```
void del_from_runqueue(register struct task_struct *p)
{       
#if 0       /* sanity tests */
    if (!p->next_run || !p->prev_run) {
        printk("task %d not on run-queue (state=%d)\n", p->pid, p->state);
        return;
    }
                                                   <—— move the #endif here, comment out the printk
    if (p == &idle_task) {
        printk("idle task may not sleep\n");
        return;
    }   
#endif                                        <——————————^^^^^^^^
    //nr_running--;
    (p->next_run->prev_run = p->prev_run)->next_run = p->next_run;
    p->next_run = p->prev_run = NULL;
    
}
```

### DIFF
--- a/tlvc/kernel/sched.c
+++ b/tlvc/kernel/sched.c
@@ -39,11 +39,11 @@ void del_from_runqueue(register struct task_struct *p)
 	printk("task %d not on run-queue (state=%d)\n", p->pid, p->state);
 	return;
     }
-    if (p == &idle_task) {
-        printk("idle task may not sleep\n");
-        return;
-    }
 #endif
+    if (p == &idle_task) {
+	//printk("idle task may not sleep\n");
+	return;
+    }
     //nr_running--;
     (p->next_run->prev_run = p->prev_run)->next_run = p->next_run;
     p->next_run = p->prev_run = NULL;
@@ -92,8 +92,7 @@ void schedule(void)
         if (prev->signal || (prev->timeout && (prev->timeout <= jiffies))) {
             prev->timeout = 0UL;
             prev->state = TASK_RUNNING;
-        }
-        else {
+        } else {
 	    timeout = prev->timeout;
 	}
     }
@@ -118,7 +117,7 @@ void schedule(void)
 
         previous = prev;
         current = next;
-	debug_sched("sched: %d\n", current->pid);
+	debug_sched("sched: %d %d\n", prev->pid, next->pid);
         tswitch();  /* Won't return for a new task */
 
         if (timeout) {
@@ -246,3 +245,4 @@ void INITPROC sched_init(void)
     t->state = TASK_RUNNING;
     t->next_run = t->prev_run = t;
 }
+


ghaerr:
Hello @Mellvik,

You're pretty good at chasing and finding rabbits!!

I wanted to comment to fully understand this issue. The scenario is new to an async-I/O kernel as otherwise there would never be a sleep in `wait_on_buffer`. What I don't quite understand is how I/O is being performed in the kernel prior to initializing the idle task. The very first thing that occurs in kernel_init is a call to sched_init:
```
void INITPROC kernel_init(void)
{
    seg_t base, end;

    /* sched_init sets us (the current stack) to be idle task #0*/
    sched_init();
    setup_arch(&base, &end);
    mm_init(base, end);
    irq_init();
    tty_init();
...
```
and sched_init initializes the task state to TASK_RUNNING:
```
void INITPROC sched_init(void)
{
...
/*
 *  Now create task 0 to be ourself.
 */
    kfork_proc(NULL);

    t->state = TASK_RUNNING;
    t->next_run = t->prev_run = t;
}
```
What did you find that is happening such that I/O is in progress prior to this? I assume we're talking your new MFM interrupt-driven disk driver causing the I/O complete interrupt, right? There should be no I/O until well after most other kernel initialization...

Thank you!

ghaerr:
I now see the full problem: in the case of booting from a HD using an interrupt-driven driver, the kernel will start block I/O using `bread` on behalf of the idle process, rather than on behalf of an application. This would be done when reading the partition table.

In this case, for race reasons, the `wait_on_buffer` sleep code is wrapped with something like the following:
```
    wait_set((struct wait_queue *)bh);       /* use bh as wait address */
    for (;;) {
        current->state = TASK_UNINTERRUPTIBLE;
        if (!ebh->b_locked)
            break;
        schedule();
    }
```
This is where the idle process' state is set to not TASK_RUNNING, which then causes the problem you've identified.

The original code preventing attempted sleeps from the idle process was added to identify problems with device drivers calling sleep from an interrupt handler, which would have the same result, and can inadvertently happen, given the complexity of kernel functions that sometimes sleep without the programmer thinking of it. The current fix would allow an idle sleep in that scenario, although the message "idle task may not sleep" has been removed so no one would know.

It may be that idle task "sleeping" should be handled differently in the kernel startup phase, when the partition table is read, and then back to "normal" later. I'll think some more about all this.

Mellvik:
Thanks for your comments - and the engagements @ghaerr.

A  couple of observations:
- AFAIC the idle task has been initialized just fine at this point, things are working as expected.
- Whether the system is booting from this drive or not, doesn't influence this scenario: The first kernel read (driver initialization actually) of a HD device is the partition table regardless.
- The current fix **prevents** an idle sleep - which as I see it is beneficial regardless.

Thus - unless I'm missing something - the fix adds a safeguard that should have been there all the time, although admittedly, the problem at hand is quite esoteric.

Mellvik:
BTW - a side note, to be noted in the driver PR: 86Box has been a unique tool in the development, with the occasional reality check on real XT hardware. 86Box has improved dramatically since I tested it a couple of years ago.

ghaerr:
> The current fix prevents an idle sleep - which as I see it is beneficial regardless.

Yes, agreed the idle task cannot be allowed to sleep. I suppose what I didn't make clear is that, with the `printk("idle task may not sleep")` warning removed, it opens the door for device drivers to unwittingly, possibly through a chain of function calls not allowed during interrupt handling time, to call sleep/wait, etc., and there would be no warning. I originally added the `#if 0` to remove the check in an attempt to improve context switching speed. In ELKS, the check (and warning) have been added back in using the conditional `#ifdef CHECK_SCHED`, along with other scheduler checks.

In summary, there are two issues: checking device drivers for disallowed kernel function calls, and allowing the kernel to make I/O calls on behalf of the idle process, something that is done only during HD partition reading startup code.

ghaerr:
>  86Box has improved dramatically since I tested it a couple of years ago.

Good to know! Are you using that instead of QEMU? What does it do better, and is it easy to setup?

Mellvik:
> > 86Box has improved dramatically since I tested it a couple of years ago.
> 
> Good to know! Are you using that instead of QEMU? What does it do better, and is it easy to setup?

The thing is - QEMU is useless for XT (8088) work for several reasons, and timing in particular. It's a real pain to run an emulator at REAL 8088/4.77MHz speed, but that's what it takes. Bumping up the clock, and things stop working (remember the days when PCs har the turbo switch? It's the same scenario, the timing must be right to ensure it works on real HW).
So booting (and running) a 360k floppy is exactly as slow in 86Box as on my 5155 and the Compaq Portable Plus. Patience is not only beneficial, it's a requirement.

BTW - The XD driver working, adding interrupts to directhd should be a breeze.

Mellvik:
I forgot - yes, 86Box is easy to set up. GUI unless you want to do something magic.

Mellvik:
> 
> In summary, there are two issues: checking device drivers for disallowed kernel function calls, and allowing the kernel to make I/O calls on behalf of the idle process, something that is done only during HD partition reading startup code.

I read you, the printk is important for development scenarios, I can see that. I'll enclose the printk in a `CHECK_SCHED` ifdef (TLVC inherited this facility). It makes sense.

ghaerr:
> The XD driver working, adding interrupts to directhd should be a breeze.

What are the big differences between your XD driver and the original directhd driver? Did it require a new driver, that much difference?

Mellvik:
Yes, it's an entirely new driver - and as I found out along the way, Linux used to have one, but it was 'decommissioned' a few years ago. And very bloated.
In short, the IDE driver (directhd) talks to each (smart) drive, the mfm driver talks to a (not very smart) controller with a different command set - and not very standardized behaviour.

ghaerr:
Nice!!! Looking forward to seeing your MFM and interrupt-driven directHD drivers :) I suppose its amazing you have a number of actually working MFM drives, given how old they must be!!

Mellvik:
Funny you should mention it. Dealing with these drives (their mtbf is 12000 hrs yes some of them still work) is a challenge. One (10MB Rodime) spins down randomly, then back up 10 or 20 minutes later. I'm in a different room, had no clue why I was getting errors. The other (20MB Miniscribe) works only half way (!), not past cyl 400 (of 616). Etc etc. Thank god 86box was stable :-)

another educating experience. The 'moderate' standardization cause  the 'big linux' driver to go as far as to search for strings in the ROM to determine the brand and version in order to handle the variations correctly. 

Not all that far from their big cousins, the SMD drives of the same age.

Thanks.

PR #54, opened 2024-03-25
[kernel] Change bootopts chs0 option in preparation for MFM driver

Mellvik:
PC/XT generation systems do not have any standardized means to specify disk drive types or parameters. Some use controller switches, some allow BIOS calls, some have proprietary extensions in the controller firmware etc.
We don't have the luxury (i.e. RAM) to look for all these variants, and `/bootopts` does the trick anyway.

Thus this extension to the `chs0` `/bootopts` option in preparation for introducing the MFM diver.
- The option has been renamed to `hdparms`, which is easier to recognize/understand.
- Instead of a CHS triple, the per drive data is now a quadruple, CHSW, W being the WritePrecompensation track, which many MFM drives require.
- The option supports 2 drives, 4+4 values: `hdparms=406,4,17,128,614,4,17,-1`
- As before, the format is flexible, permitting empty fields and shortened format, as in `=406,4,17` which sets all missing fields to zero. Or `=,,,416,4,17` which sets all drive 0 values and drive 1 wpcomp to zero.
- CHS values will be ignored if the cylinder value is zero.
- -1 is a frequently used wpcomp value in all kinds of tables from yesteryear (and means 'turn off/not used'). Thus -1 is permitted and converted to zero in order to avoid adding support for negative numbers in `init/main.c`. The driver then converts the zero wpcomp to whatever is required by the controller.
- The IDE driver (`directhd.c`) has been updated to take advantage of the 2-drive support.

The drives using `hdparms` will use default values when parameters are missing. For IDE drives, that is the values provided by the drive itself. For MFM it's the values for a type 1 drive, 10MB - 406/4/17/128.

### DIFF
--- a/tlvc/init/main.c
+++ b/tlvc/init/main.c
@@ -45,7 +45,7 @@ int nr_mapbufs;
 
 #define CHS_DRIVES 2	/* # of drives to config in bootopts */
 #define CHS_ARR_SIZE	CHS_DRIVES * 4
-int hdparms[CHS_ARR_SIZE] = {0};	/* cover 2 drives */
+int hdparms[CHS_ARR_SIZE];	/* cover 2 drives */
 
 int netbufs[2] = {-1,-1};	/* # of network buffers to allocate by the driver */
 static int boot_console = 0;
@@ -333,8 +333,8 @@ static int INITPROC parse_dev(char * line)
 	return (base + atoi(line));
 }
 
-/* Parse comma-separated list of numbers. Empty fields (adjacent commas) mean
- * zero, negative numbers become zero */
+/* Parse comma-separated list of numbers. Missing and empty fields (adjacent
+ * commas) mean zero */
 static void INITPROC parse_parms(int cnt, char *line, int *nums, int base)
 {
 	int i;
@@ -345,7 +345,7 @@ static void INITPROC parse_parms(int cnt, char *line, int *nums, int base)
 		m = l;
 		while ((*l) && (*l != ',')) l++;
 		c = *l;		/* ensure robust eol handling */
-		if (l > m && *m != '-') {
+		if (l > m) {
 			*l = '\0';
 			nums[i] = (int)simple_strtol(m, base);
 		} else nums[i] = 0;	/* item missing or negative */
@@ -361,25 +361,6 @@ static void INITPROC comirq(char *line)
 	parse_parms(MAX_SERIAL, line, irq, 0);
 	for (i = 0; i < MAX_SERIAL; i++) 
 		if (irq[i]) set_serial_irq(i, irq[i]);
-}
-
-static void INITPROC init_hdparms(int dev, char *line)
-{
-	int i, val[CHS_ARR_SIZE] = {0};
-
-	parse_parms(CHS_ARR_SIZE, line, hdparms, 10);
-	printk("init_hdparms: ");
-	for (i = 0; i < CHS_ARR_SIZE; i++) {
-#if 0
-		if (!val[i]) {	/* sanity check, no zero values accepted */ 
-			val[0] = 0;
-			return;
-		}
-#endif
-		//hdparms[i] = val[i];
-		printk("%d ", hdparms[i]);
-	}
-	printk("\n");
 }
 
 static void INITPROC parse_nic(char *line, struct netif_parms *parms)
@@ -521,7 +502,6 @@ static int INITPROC parse_options(void)
 
 		if (!strncmp(line,"hdparms=", 8)) {
 			parse_parms(CHS_ARR_SIZE, line+8, hdparms, 10);
-			//init_hdparms(0, line+8);
 			continue;
 		}
 		if (!strncmp(line, "netbufs=", 8)) {
--- a/tlvc/init/main.c
+++ b/tlvc/init/main.c
@@ -42,7 +42,11 @@ struct netif_parms netif_parms[MAX_ETHS] = {
 __u16 kernel_cs, kernel_ds;
 int tracing;
 int nr_mapbufs;
-int ide_chs[3] = {0,0,0};
+
+#define CHS_DRIVES 2	/* # of drives to config in bootopts */
+#define CHS_ARR_SIZE	CHS_DRIVES * 4
+int hdparms[CHS_ARR_SIZE] = {0};	/* cover 2 drives */
+
 int netbufs[2] = {-1,-1};	/* # of network buffers to allocate by the driver */
 static int boot_console = 0;
 static char bininit[] = "/bin/init";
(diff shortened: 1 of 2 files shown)


tlvc/init/main.c
```diff
@@ -352,19 +363,23 @@ static void INITPROC comirq(char *line)
 		if (irq[i]) set_serial_irq(i, irq[i]);
 }
 
-static void INITPROC init_chs(int dev, char *line)
+static void INITPROC init_hdparms(int dev, char *line)
 {
-	int i, val[3];
+	int i, val[CHS_ARR_SIZE] = {0};
```
ghaerr, line 368: Hello! A couple of comments: initializing auto (stack) variables like this is very expensive. Under the hood, `gcc-ia16` will generate code to call `memset` (which is normally found in libgcc.a but is overridden in this case by the same named routine in lib/string.S):
```
   9:	b8 06 00             	mov    $0x6,%ax
   c:	50                   	push   %ax
   d:	b8 00 00             	mov    $0x0,%ax
  10:	50                   	push   %ax
  11:	52                   	push   %dx
  12:	16                   	push   %ss
  13:	1f                   	pop    %ds
  14:	e8 fe ff             	call   15 <init_hdparms+0x15>
			15: R_386_PC16	memset
  17:	83 c4 06             	add    $0x6,%sp
```

It might be better to consider handling zero initialization as part of an else clause, as in kernel code calling implicit procedures can have unintended side effects. FYI.

ghaerr: I just noticed in the code above that `gcc-ia16` actually resets DS from SS to guarantee that `memset` has the correctly addressable segment for the variable to be cleared. This is a good example of unintended consequences, as if the kernel for some reason at the moment required DS != SS, this would reset DS without one knowing it, all from a relatively innocuous variable declaration!

tlvc/init/main.c
```diff
@@ -334,10 +345,10 @@ static void INITPROC parse_parms(int cnt, char *line, int *nums, int base)
 		m = l;
 		while ((*l) && (*l != ',')) l++;
 		c = *l;		/* ensure robust eol handling */
-		if (l > m) {
+		if (l > m && *m != '-') {
```
ghaerr, line 348: > Thus -1 is permitted and converted to zero in order to avoid adding support for negative numbers in init/main.c. 

`simple_strtol` does handle (or at least it's coded to supposed to handle) negative numbers. If this isn't working, perhaps there's a bug in it? I can chase that down if required.

tlvc/init/main.c
Mellvik: Wow, thanks @ghaerr - this is doubly useful. First the knowledge (actually somewhat scary about the DS reset), secondly the fact that I forgot to delete the entire thing. The `init_hdparm()` function is no longer used.

Appreciated.

tlvc/init/main.c
```diff
@@ -42,7 +42,11 @@ struct netif_parms netif_parms[MAX_ETHS] = {
 __u16 kernel_cs, kernel_ds;
 int tracing;
 int nr_mapbufs;
-int ide_chs[3] = {0,0,0};
+
+#define CHS_DRIVES 2	/* # of drives to config in bootopts */
+#define CHS_ARR_SIZE	CHS_DRIVES * 4
+int hdparms[CHS_ARR_SIZE] = {0};	/* cover 2 drives */
```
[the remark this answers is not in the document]
Mellvik: Again, appreciated @ghaerr. I keep forgetting this it seems.

tlvc/init/main.c
Mellvik: Interesting - I'm guilty in not testing this, instead assuming based on some vague (unreliable) memory that negative #s weren't supported.  Time to simplify even more.
Thanks.

tlvc/init/main.c
ghaerr: Further analysis shows that the push SS/pop DS is eliminated with optimization on, which is the default for the kernel compilation (`memset` is still used to initialize the array variable). Initializations of single `char` or `int` variables are performed with `mov` instructions. 

Interestingly, an initialization like `int val[3] = {1};` first does a `memset(...,0,6)` for the entire array, followed by a `mov $1,-6(%bp)` to the first element, and `int val[3] = {1, 1};` does a `memset` followed by two moves to the first two elements!!

Finally, a full (non-zero) initialization like `int val[3] = {1, 1, 1};` generates three words in the data segment of value 1, and then uses internal `memcpy` to initialize the auto stack variable from the data segment. That I didn't know - and I think I'll stay away from any auto array initializations knowing that!

Mellvik: Again, very interesting indeed @ghaerr. Incidentally, while working on this PR, I became unsure about how array initialization worked, wondering about the (possible) difference between say `int i[10] = {0,}` and `int i[10] = {0}`.

What I found was both educating and surprising - including the fact that this is all gcc only and thus non-standard. OTOH - to most of us, gcc IS the standard and this is what counts.

Anyway, this is the way I now understand array initialization:
- There is no difference between the two variants above.
- Zero has special significance as an initializer. 
- While `int i[10] = {0};` does what you'd expect and initializes all elements to 0, `int[10] = {1};` does not what I'd expected. It sets the first element to 1, the rest to zero.

There is more to it, but this is the gist of it - and it seems coherent with the discoveries (initialization mechanisms) you've made.

This knowledge - and the real cost of array initialization as you've demonstrated - will surely change my focus on initializers and initializing. There may be many bugs out there caused by erroneous assumptions about how initializers actually work. 

Thanks.

ghaerr: > While int i[10] = {0}; does what you'd expect and initializes all elements to 0

Yes, that is standardized in C89.

> int[10] = {1}; does not what I'd expected. It sets the first element to 1, the rest to zero.

Yes, that also is standardized in C89 - "If there are fewer initializers in a list than there are members of an aggregate, the remainder of the aggregate shall be initialized implicitly the same as objects that have static storage duration." In other words, non-specified values in an initialization list will be set to 0.

> the real cost of array initialization as you've demonstrated - will surely change my focus on initializers and initializing.

Good, I'm glad I was able to help. What I try to do is guess at what I would think the compiler is generating for any line of C written... and then occasionally verify because sometimes it isn't what one thinks! GCC has all sorts of other "optimizations" one has to watch out for; one that's gotten me a number of times is `printf("%s\n", str)` is converted to `puts(str)` without any mention. `memset` and `memcpy` are called behind the scenes for lots of things, including initialization and structure assignment.

Mellvik: So C89 it is, thanks for pointing that out - I'm guilty at putting too much trust in the source I found on the net. 
It seems even after all these years, there's still stuff to be learned about C.

PR #56, opened 2024-04-02
MFM disk driver 

Mellvik:
This PR introduces a new set of peripherals to TLVC - support for pre-IDE disk drives, aka MFM or ST506 drives.

The MFM drives predates the IBM PC and DOS, and turned 1980 vintage PCs into real computers. However, their price level prevented them from becoming widespread until the PC/XT and clones, with their high volumes sent the prices spinning.

In Vintage Computing, these drives live in the far low end - the XT, 8088/4.77MHz class of systems. Given their age and 'brittleness' compared to later systems and components, surprisingly many of these disks (and systems) still exists and works. Thus the effort to make the available via TLVC - not only to recreate the past but to recover 'lost' data.

The driver supports the basic ('standard') ST506 interface with no attempts to include extra functionality added by 2nd and 3rd gen controllers. The DMA channel (3) and IO address (0x320) are hardwired into the older controllers - beware of ti IO address in particular, it's sometimes used for network controllers and other devices. The driver supports 1 controller, 2 drives. Development and testing on 86Box which turns out to be very good at true emulation of really old hardware, and on real hardware, IBM 5155 and Compaq Portable Plus.

The driver has been written from scratch - with inspiration from the Venix XT driver and the Linux 1.0 MFM driver. At this time there is still a lot of debugging code commented out in the driver itself, to ble cleaned out as more practical experience has been gathered. That said, the driver has been operational and continuously improved for several months. 

**Todo**
- Add raw driver
- Add IOCTLs for formatting and to turn on/off automatic retries (for performance testing)
- Add format utility
- Remove extra `printk`s during boot
- Minor change to `ps` to ensure that the `uptime` alias always works (even with a path added to it).

Refer to developer notes below for more details, choices made and practical experiences.

**Also in this PR:**
- The BIOS-only storage configuration now compiles and works.
- The IRQ vector code, changed in PR #33, has been simplified (thanks @ghaerr).
- Given the missing ability to detect the physical geometry of MFM drives via their controllers, the `/bootopts` `chs0` option has been replaced and expanded, see #54.
- Unrelated to this driver: Add a utility to pick up the date and time from the **AST 6pack plus** which seems to be the standard for this generation of machines.

Boot messages (ignore the date):
```
ttyS0 at 0x3f8, irq 4 is an 8250
16 ext buffers (16384B ram), 10 L1 buffers, 10240B ram
eth: ne0 at 0x300, irq 12 not found
eth: wd0 at 0x280, irq 2, ram 0xcc00, (wd8003) MAC 00:00:C0:4D:F6:13, flags 0x80
Floppy drive(s) [no CMOS]: df0, df1 set to 360k/PC (type 1)
df: Direct floppy driver, FDC (8272A) @ irq 6, DMA 2
xd: MFM disk driver @ 0x320, irq 5, DMA 3
xd0: CHS 306/4/17/128 (from /bootopts)
xd: Found 1 hard drive 
xd0: 4 heads, 306 cylinders, 17 sectors (~10MB)
Partitions: xda:(0,20808)  xda1:(17,20791) 
device_setup: root device 0x601
PC/XT class machine, syscaps 0x0, 640K base ram, 16 tasks.
TLVC kernel 0.6.0 (59632 text, 14320 ftext, 10896 data, 29248 bss, 25390 heap)
Kernel text 2d0:0, ftext 115f:0, init 1316:0, data 14de:0, top a000:0, 499K free
MINIX: inodes 3464 imap 1 zmap 2, first data 114
VFS: Mounted root 0x0601 (minix filesystem).
Running /etc/rc.sys script
Starting networking on wd0
ktcp -b -p wd0 10.0.2.18 10.0.2.2 255.255.255.0
ktcp: ip 10.0.2.18, gateway 10.0.2.2, netmask 255.255.255.0
ktcp: /dev/wd0 mac 00.00.c0.4d.f6.13 mtu 1500
Thu Feb  9 00:14:47 2073

TLVC 0.6.0

login: 

```

**Developer notes** (from the driver source code):
```
/* DEVELOPER NOTES:
 * - We don't check for extreme transfer sizes (like > 32k) which will crash
 *   the system because the RAM refresh is disturbed. This is safe because
 *   TLVC cannot - without special programming - transfer more than 20k in a raw
 *   read/write operation anyway. Block mode operations are always 1k.
 * - Sectors - controller level - start at 0, not 1!!
 * - The 1st MFM controller is always at 0x320, IRQ5, DMA3. This is a commonly used
 *   address on network interfaces and others. Collisions have interesting consequences,
 *   not necessarily easy to decipher.
 * - Using a MFM controller on a AT or later machine may or may not work. 1st gen
 *   AT machines emulate MFM controllers @ 0x320 for compatibility, masking the fact that
 *   the disk(s) are IDE/ATA. Adding a physical mfm controller at that address
 *   may or may not work.
 * - There is no 'standard' MFM controller. All have the same basic command set, most
 *   have extensions and the behaviour of the 'state machine' governing the command
 *   phase varies. E.g. there is no standard way to determine disk drive data (CHS etc.).
 *   This is why the 'big' Linux MFM driver pokes into the controller ROM to find the
 *   type and details. We don't have space for that, and assume the
 *   default 10MB CHS values (304/4/17/128) unless /bootopts is used to define the size.
 *   The last value in the quadruple is the Write Precompensation track, aka WPCOM.
 *   The good thing is that since all drives in this category have (or emulate) 17
 *   sectors, there is no problem booting even if the real size of the drive is unknown
 *   at that point. After booting, TLVC can make its own assumptions about size.
 *   Note however, that the controller needs to be told the CHS values we plan to use
 *   in order to provide access to the entire drive w/o errors (see the xd_setparam()
 *   function()).
 * - A drive works only with the type of controller with which it was formatted.
 *   IOW moving drives around is a no-go. Newer controllers (actually most except the
 *   very first generation, pre 1985) have a formatter included in the firmware, to be
 *   started from MSDOS DEBUG, typically at C800:5. For the older controllers, other
 *   tricks are required.
 * - Pending the availability of a formatting utility for TLVC, a crude formatter is
 *   included in the driver. If compiled in (#define ALLOW_FORMATTING),
 *   set the /bootopts sector count for the drive to 0, and the drive will be formatted.
 *   [Obviously, the sector count needs to be reset to normal before the next boot.]
 *   There is no safe way (at this point) to detect the completion of the format process,
 *   so keeping an eye on the activity LED is required. It's important to set the correct
 *   CHSW parameters before the formatting. Also notice that the interleave is preset to 5,
 *   which is good for <8MHz machines, otherwise use 4.
 * - INTERLEAVE: The Hard Drive Bible dieagrees with the above (which comes from a WD document).
 *   Their recipe is 4:1 @4.77MHz or less, 3:1 @ 5-10MHz, 2:1 @ 10-16 MHz, 1:1 for higher.
 *   Remember - this is all about old MFM/RLL drives.
 * - Beware that the sluggishness of 4.77MHz PCs may deliver surprisese, in
 *   particular related to interrupts. Like the interrupt occuring before the request call
 *   has completed (see also comments in the code).
 * - If drive 0 is missing and drive 1 present, some BIOSes will seem to hang at boot time.
 *   Be patient, 15-20 seconds, and it continues - TLVC will boot fine. Such configuration
 *   is not hard-disk bootable though (floppy OK).
 * - RESET: On some controllers, the drive parameters must be re-programmed after a reset.
 */
```


tlvc/fs/buffer.c
```diff
@@ -438,23 +442,25 @@ struct buffer_head *get_free_buffer(void)
...
 	     */ 
-		printk("RelL1:");
+		//printk("RelL1:");
 		for (i=0; i<nr_mapbufs; i++)
 			brelseL1_index(i, 1);
-		printk("\n");
+		//printk("\n");
 		break;
 #else
 		continue;
```
ghaerr, line 451: Just happened to notice this - in the case of L1 buffers only, `continue` would continue the while loop but without incrementing sync_loop. It would seem that would cause an infinite loop if ebh->b_next_lru were NULL (meaning no buffers available)? Perhaps should be `break` like the other switch cases?

Mellvik: Indeed, @ghaerr - much appreciated. A reminder that the L1 only setting is 'under-tested'. Thank for reviewing!

Issue #67, opened 2024-06-29
ktcp send is too slow


Mellvik:
This issue is described in detail in #64 under the heading **TCP speed kludge**, which essentially observes that introducing a delay in the ee16 driver's `write select` function will drastically increase send-speed.

What happens is that the delay prevents an extra round through `select_wait/wakeup`. A small delay causing a big speedup by 'disturbing' the transactional rhythm of the send process between `ktcp` and the driver. A complicated issue indeed because of the design of the system, but still one of great interest because of the improvement potential.

The 3C509 driver was just updated (by accident directly to `master`) with a delay similar to the `ee16` driver and the ftp-send speed improvement was a factor of 4, sometimes beating 100kbps. This is remarkable - not the least because the max outgoing tcp packet size is 512, meaning there are a lot of packets processed.

ghaerr:
Interesting. I'm not looking at the source, but it seems what you're saying is that the NIC driver, after finishing the transmit of a packet, does not have access to another queued packet, so it stops sending. Meanwhile, ktcp is waiting in `select` for either write OK from the driver (or a read OK indicating another packet received, which is not this case, or is it?). Ktcp then returns from select and writes another packet. How exactly is the write driver delay (before or after sending a packet?) allowing this to speed up?

If ktcp is actually waiting for an ACK from the driver before continuing to provide the next packet to write, then I suppose this is more complicated in that it opens an additional possibility that ktcp is running processing the received ACK when an additional NIC write complete interrupt occurs. In this case perhaps the select wait code isn't correct and the write-available status isn't set properly, causing ktcp's next call to select to wait for yet another interrupt (from something?) in order to be allowed to proceed. In this case it is extremely important that the driver mutex read/write rules are followed to prevent an inadvertent race condition on accessing the write-OK variable. I believe it is correct but its been a while since looking at it. Given that another NIC driver is showing the same symptoms, this likely isn't the cause, but I think all the drivers share much the same mutex access code.

With regards to select wait/wakeup, other than going into the kernel and back which isn't exactly super fast but not really slow, there could be the following reasons a select wait/wakeup cycle taking lots of time:
- wakeup is a bit slow, it for-loops through all 16 tasks every time its called, which previous kernel inspection has shown may be lots more than one thinks.
- select itself is somewhat slow, due to its nature as well as the tricky method used to add wait addresses to the poll queue.
[...]

PR #68, opened 2024-07-10
[net, ktcp] Enhancements: Speed, retransmits

Mellvik:
This PR fixes the `ktcp` performance problem described in #67 by increasing the allowed max send window. The problem popped up only on relatively fast systems and only if there was a router between the TLVC system and the peer. The fix eliminates a wait/wakeup cycle and allows the TLVC sender to send packets more or less continuously, reaching the same level of performance as if the hosts were on the same network segment.

Also fixed is an old `ktcp` problem which caused lots of unneccessary retransmissions when sending to slow host. The retransmit timeout calculation simply didn't work in such settings. The fix is to check the roundtrip time before setting the first retransmit timeout. If the peer system is found to be very slow or heavily loaded, double the initial timeout.

Included:
- The delay kludge previously added to the ee16 and the el3 drivers to get around the slow-send problem has been removed.
- Shortening the read_select wait timeout from 100ms to 50ms  did speed things up, but not as much as the selected solution, increasing the allowed retransmit window. Also, the shortened wait timeout seems to create deadlocks in certain situations. A comment about this has been added to the `af_inet.c` file.

### DIFF
--- a/tlvc/arch/i86/drivers/char/tcpdev.c
+++ b/tlvc/arch/i86/drivers/char/tcpdev.c
@@ -24,6 +24,12 @@
 
 #ifdef CONFIG_INET
 
+#ifdef LOCAL_DEBUG	/* for flow tracing */
+void kputchar(int);
+#else
+#define kputchar(x)
+#endif
+
 unsigned char tdin_buf[TCPDEV_INBUFFERSIZE];	/* for reading tcpdev*/
 unsigned char tdout_buf[TCPDEV_OUTBUFFERSIZE];	/* for writing tcpdev*/
 
@@ -55,15 +61,16 @@ static size_t tcpdev_read(struct inode *inode, struct file *filp, char *data,
 	    return -ERESTARTSYS;
 	}
     }
+    kputchar(',');
 
     /*
      *  If the userspace tcpip stack requests less data than what is in the
      *  buffer it will lose data, so the tcpip stack should read BIG.
      */
-    if (len < tdout_tail)
+    if (len < tdout_tail) {
 	debug_net("TCPDEV(%d) read len too small %u\n", current->pid, len);
 	//printk("TCPDEV read len too small %u (%u)\n", len, tdout_tail);
-    else len = tdout_tail;
+    } else len = tdout_tail;
     memcpy_tofs(data, tdout_buf, len);
     tdout_tail = 0;
     up(&bufout_sem);		/* flag buffer available */
@@ -127,10 +134,13 @@ static int tcpdev_select(struct inode *inode, struct file *filp, int sel_type)
 	break;
     case SEL_IN:
 	debug("TCPDEV(%d) select SEL_IN\n", current->pid);
-	if (tdout_tail != 0)
+	if (tdout_tail != 0) {
+	    kputchar(':');
 	    ret = 1;
-	else
+	} else {
+	    kputchar(';');
 	    select_wait(&tcpdevq);
+	}
 	break;
     default:
 	debug("TCPDEV(%d) select bad type %d\n", current->pid, sel_type);
--- a/tlvc/arch/i86/drivers/net/ee16.c
+++ b/tlvc/arch/i86/drivers/net/ee16.c
@@ -6,18 +6,12 @@
 */
 /*
  * TODO:
- * - Fix and test 8bit bus functionality
- * - More testing with 16k, currently seems to work fine with both PIO and shmem,
- *   shmem is slightly (4-5%) faster.
+ * - Fix and test 8bit bus functionality (low pri)
  * - There are still occasional hangs in probe() when changing from shmem to pio mode
  *   w/o powercycling (ie. via /bootopts and reboot)
  * - Much of the error handling code is untested as it's hard to get errors to happen in 
  *   protected environments. This also applied to the CUwedged recovery code.
  *
- */
-
-/*
- * Developer notes:
  */
 
 /* Here's the scoop on memory mapping 	(in verbatim from the Linux driver )
@@ -52,7 +46,6 @@
  */
 
 #include <arch/io.h>
-#include <arch/ports.h>
 #include <arch/segment.h>
 #include <linuxmt/memory.h>
 #include <linuxmt/errno.h>
@@ -278,7 +271,7 @@ static unsigned short SHADOW(unsigned short addr) {
  * The NetBSD driver is smart about this, setting an async flag whenever possible.
  * FIXME!!
  */
-#define CMD_CLEAR_TIMEOUT 40	/* approx .8 seconds */
+#define CMD_CLEAR_TIMEOUT ((HZ/10)*8)	/* approx .8 seconds */
 
 static void ee16_cmd_clear(unsigned int ioaddr)
 {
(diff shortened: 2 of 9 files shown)


ghaerr:
Nice working solving these problems @Mellvik! Fixes look good to me. I am a bit interested to know whether ktcp shows any signs of running out of memory with the expanded send window and max retransmit queue sizes, given the number of simultaneous connections you're testing with.

Mellvik:
Thanks, @ghaerr.
The answer is no, but here's the thing: Testing (really, pushing) this requires more processes than the system has available.  Which arguably means that everything is fine with `ktcp` since there is another resource running before it reaches its limits. Not good enough though - and I'm adding dynamic task array from ELKS as we speak.

Two heavy outgoing file transfers now run fine at great speed - same segment and through the router, even if disturbed by other activities. The litmustest that really pushes ktcp and the retransmit implementation is 3 incoming telnet sessions running `hd` on some file - concurrently. Adding floppy IO to the equation and we have some rather unpredictable (and big) delays. Unrealistic load, but this is as much about robustness as it is about capability.

IOW - the jury is still out ...

Mellvik:
@ghaerr,
Turns out I was too fast on declaring the retransmit problem between slow systems as fixed. It is not, and it runs deeper than just the RTT/RTO values. A classic 'state' problem - works in this case, but not in that. 

As to the dynamic task array: I ported it and it seems to work fine, but as we're running low on memory, processes start to hang. The ktcp problem prevented me from diving into this one right away, but just out of curiosity - and before diving in: How much has this feature been tested? My setting is 20 tasks by the way.

ghaerr:
> I was too fast on declaring the retransmit problem between slow systems as fixed.

Oh geez, that's too bad. It looks as though you're on to something though. This first fix only doubles the initial RTO in one case, perhaps RTO needs to be calculated over a longer time period. I don't have enough data on how RTO and retransmit works to suggest anything intelligent at this point.

> How much has this feature been tested? My setting is 20 tasks by the way.

I tested the feature very heavily and believe it works quite well, as I was careful not to change anything structurally other than using a pointer for `struct task_struct *task` rather than having it be static declaration of an array. However - the full change may have occurred over more than one commit, so check that. The other huge issue is that the task array is still allocated out of kernel data segment (heap) and each task struct is large 700+ bytes stack, etc. To more smoothly determine whether dynamic tasks work on TLVC, I would suggest starting at 16 and seeing no problems, then inching upwards.

Moving the task array into far memory and/or allocating different stack sizes per task turns out to be extremely complicated, so unfortunately tasks remain a heavy user of kernel heap. I added dynamic inodes later in an attempt at allowing a smaller inode store to increase kernel memory for tasks. Ultimately, all of this points to the huge need to get all driver buffers out of the kernel heap unless absolutely necessary. A rule of thumb would be one kernel buffer is about the same as an additional task in terms of data usage.

ghaerr:
@Mellvik, super writeup and explanation on RTT/RTO. I don't think I ever fully understood it myself, but this helps.

To help me understand, is this PR fix trying to solve multiple problems (e.g. increasing max send window to allow for more time for ACKs so that efficiency is increased, and also trying a new RTT/RTO algorithm because of our previous multiple-ACK problem/hack), or something else?

It seems the initial issue was that you thought the send window size increase could solve the efficiency issue and allow removal of the read delay kluge hack. Did that get partially solved, or are you finding that somehow retransmits are getting involved in all of this, thus leading to the RTT/RTO calc issues?

Mellvik:
Actually, @ghaerr - with the upcoming update to the pr, both are fixed. It is purely incidental that the two fixes virtually end up in adjacent lines in the code.

The original problem was fixed in the pr as is and it's truly a relief to have found and fixed (testing now)  the second one - finally. The fix needs testing in a real high latency environment to make sure nothing is broken. Planned for tomorrow.

Then back to the dynamic task array testing. Thanks.

ghaerr:
> Still, I added a per-CB counter to keep track of how many packets this CB has in the retransmit buffer at any point. Like tcp_timeruse except per CB instead of total. Using this, a buildup of unacked packets can trigger a temporary increase in RTO for the CB.

This sounds like a potentially good idea. If the problem is that the retransmit buffer is unloading everything all at once when a single RTO expires, perhaps the first retrans packet should be sent, and then either the remaining RTT/RTOs adjusted and/or the send window itself decreased so that the retrans packets go out more slowly. That is, when the machine-to-machine communication "clogs up" and effectively halts, awaiting a retransmit (or provoking other problems with a faked double ACK), the sender might want to start sending quite differently than it otherwise would have, by starting quite slow with a single packet, then speeding up slowly. This could possibly be done using RTT/RTO fixups like you're thinking but applying them to all the packets in the retransmit queue as well, and adjusting their first/next retrans start time.

I hope I'm making sense here, I probably need to dive in to help more specifically. I sure wish there was an easy way to duplicate this using emulators, as of course localhost connections don't even hit the NIC cards.

Mellvik:
Thanks @ghaerr - you are making sense. 
I'm experimenting with various mechanisms righty now, and sometimes discovering behaviour I still don't quite understand. More later.

-M

Mellvik:
> > Still, I added a per-CB counter to keep track of how many packets this CB has in the retransmit buffer at any point. Like tcp_timeruse except per CB instead of total. Using this, a buildup of unacked packets can trigger a temporary increase in RTO for the CB.
> 
> This sounds like a potentially good idea. If the problem is that the retransmit buffer is unloading everything all at once when a single RTO expires, perhaps the first retrans packet should be sent, and then either the remaining RTT/RTOs adjusted and/or the send window itself decreased so that the retrans packets go out more slowly. That is, when the machine-to-machine communication "clogs up" and effectively halts, awaiting a retransmit (or provoking other problems with a faked double ACK), the sender might want to start sending quite differently than it otherwise would have, by starting quite slow with a single packet, then speeding up slowly. This could possibly be done using RTT/RTO fixups like you're thinking but applying them to all the packets in the retransmit queue as well, and adjusting their first/next retrans start time.
> 
> I hope I'm making sense here, I probably need to dive in to help more specifically. I sure wish there was an easy way to duplicate this using emulators, as of course localhost connections don't even hit the NIC cards.

You're definitely on to something, @ghaerr. Got me thinking through the whole thing again. It's tempting now that the problem is (more ) understood, to throw in a fix and move on. Only to find that the not-so-well understood consequences bite back later.

I've rigged my test environment to have a machine in Amsterdam log into the local TLVC system to get a picture of 'normal' jitter, delays and occasional retransmits. ktcp handles it well and this becomes - to me - a baseline, as in 'it ain't broken, don't fix'.

Btw - from the above, ktcp does not really dump all the packets in the retrans buffer on the recipient, it just looks that way. They all time out, because the timeout is too short. So the idea of letting the first retrans go and then adjust the timeout in the remaining packets  is in fact very viable. More about that later.

Here's the thing, the challenge: we're dealing with a special case, one that breaks the rules of 'normality' sufficiently to create real problems. How do we (1) recognize this particular situation and (2) handle it?

Before doing more experimentation (not that it has been in vain, I've learned a lot), those 2 problems  must be assessed, the first one by carefully studying the logs/traces (again). I'll have something to that effect tomorrow.

PR #72, opened 2024-08-02
[kernel] Add dynamic task table from ELKS (and more)

Mellvik:
This PR adds configurable task table sizing from ELKS PR#1839 (needs more testing with more than 17 tasks).
The number of tasks is specified in `/bootopts` using the `tasks=` keyword.

Also included is improved alignment control per the discussion in issue #71 and ELKS PR #1947. 

Out of necessity, some constants related to the upcoming LANCE network driver are also included (`init/main.c` and `arch/ports.h`).


Mellvik:
Thanks for the input on these, @ghaerr.

It is interesting - in the context of alignment - that while the word alignment now forced via the `p2align 1` directive improves processor fetch timing on the 8086 (and to some extent on newer processor), it does nothing for the 8088, which is presumably is more widespread. The 8088 on the other hand takes a hit every time a word is accessed (as opposed to a byte) for obvious reasons.

The question that comes to mind is whether 'struct packing' and our general use of `unsigned char` to save RAM is good or bad. I'm leaning towards 'good' ...

ghaerr:
> The question that comes to mind is whether 'struct packing' and our general use of unsigned char to save RAM is good or bad.

It's two old stories - CISC vs RISC, and the evolution of cheap RAM. Nowadays, its all about speed and gigabytes of RAM so  they dropped all the extra transistors and (in x64 long mode) everything must be aligned and the CPU'll trap instead of using extra fetching logic.

For our purposes here, you make a good point that structures should still be visually checked, for manual alignment, as now we know that all older CISC processors take considerable timing hits otherwise. I normally check automatically out of habit, but I'm going to do that again for kernel structures.

Regarding which way to lean - I guess I can't say good or bad since as you mention, the 8088 takes extra cycles on words, and the 8086 takes extra cycles on odd bytes... so much for ancient CPUs :)

On the subject of speed - I'm not sure its a great idea to optimize all of TLVC or ELKS around 8086 and 8088-specific timings. The reason is that these older systems are in much smaller use than say 386/486/Pentium+, and they're slow as heck anyways. So while our discussion of 8086/8088 timings is great to learn from to get a feel for kernel and driver timings, not writing a routine because it's got five MULs may be going a bit far. We'll have to see, building a useful precision timer will probably take a few revisions anyways...

A related interesting side note on structure alignment is the `struct wait_queue { char pad; }`, used in various places. This structure takes up a byte, but the byte isn't used, as only the struct's address is needed - what we really want is a zero-length structure. GCC allows zero-length structures as an extension, but I've hesitated to add this into ELKS because I thought it too tricky: if two wait_queues were placed next to each other, the kernel wait mechanism would stop working!

Mellvik:
> > The question that comes to mind is whether 'struct packing' and our general use of unsigned char to save RAM is good or bad.

> 
> For our purposes here, you make a good point that structures should still be visually checked, for manual alignment, as now we know that all older CISC processors take considerable timing hits otherwise. I normally check automatically out of habit, but I'm going to do that again for kernel structures.

My comment was mostly rethorical - and admittedly somewhat selfserving in that I think you have be really old in the game to have these reflexes at all. We're lucky. That said, most of the time I think the packing is good because the environment is always memory strapped.
> 
> Regarding which way to lean - I guess I can't say good or bad since as you mention, the 8088 takes extra cycles on words, and the 8086 takes extra cycles on odd bytes... so much for ancient CPUs :)
> 
> On the subject of speed - I'm not sure its a great idea to optimize all of TLVC or ELKS around 8086 and 8088-specific timings. The reason is that these older systems are in much smaller use than say 386/486/Pentium+, and they're slow as heck anyways. So while our discussion of 8086/8088 timings is great to learn from to get a feel for kernel and driver timings, not writing a routine because it's got five MULs may be going a bit far. We'll have to see, building a useful precision timer will probably take a few revisions anyways...

I agree 100%. Fortunately, optimizations for one architecture will be an improvement for all of them - to some degree. Completely avoiding some instructions (like MUL or DIV) because they are slow on one or two variants, is definitely unsmart.

While it has been an interesting exercise to work with real XT/8088/V20 hardware, I do consider ATs to be the ideal platform for TLVC and ELKS, with either 286 or 386SX processors. They are true 16 bit (unlike the 386DX, which always takes a two-clock hit unless the address is 32bit aligned), thus lend themselves well to real mode and can address memory beyond the first meg. I know I keep repeating myself, but my 386SX 40MHz Dream System with CF disk appears as fast as QEMU except when accessing slow peripherals.

> 
> A related interesting side note on structure alignment is the `struct wait_queue { char pad; }`, used in various places. This structure takes up a byte, but the byte isn't used, as only the struct's address is needed - what we really want is a zero-length structure. GCC allows zero-length structures as an extension, but I've hesitated to add this into ELKS because I thought it too tricky: if two wait_queues were placed next to each other, the kernel wait mechanism would stop working!

How interesting! Actually, I don't understand this: How can a zero-length struct have any practical value? If there is no memory location assigned to it, how can it be addressed? And if there is a memory location assigned to it, how can it be empty? It would seem a `struct wait_queue { int pad;}` to get alignment would make sense. Wasting a byte being a calculated price...

Mellvik:
@ghaerr,

>> It would seem a struct wait_queue { int pad;} to get alignment would make sense.

Come to think of it - and in lieu of the discussion above - this is simply wrong. The purpose of the `wait_queue` struct is to provide a unique number for reference,  not a location to be accessed. Whether it's odd or even is moot because it's never referenced. 

So `{ char pad; }` is good, saving space and no actual penalty. And an empty struct (if it were possible) wouldn't cut it since the value is missing. Am I right?

ghaerr:
> And an empty struct (if it were possible) wouldn't cut it since the value is missing.

No - an empty struct will work fine. It just won't work if the programmer happens to place to `struct wait_queue`s directly next to each other, in either another struct or as globals. Which brings us back to my reasoning above for not adding the empty struct feature to save space in structs - it's too tricky!

PR #73, opened 2024-08-08
[kernel] Precision timer code, printk updates from ghaerr, cleanups

Mellvik:
Add @ghaerr's precision timer code (from ELKS #1953) as discussed in #71 and an expanded `printk` also from @ghaerr, which - among other things - eliminates the need for the `kdevname()` call.

Also, some cleanups.

As to how to use the timer routines, take a look at the examples in the `testloop()` routine at the end of `tlvc/init/main.c`.


ghaerr:
prectimer.c also needs to be pushed.

Mellvik:
Rushing things is never good...

PR #78, opened 2024-08-26
More leftovers from previous checkins

Mellvik:
Also includes `ne2k` support for setting the global `running_qemu`.

### DIFF
--- a/tlvc/arch/i86/drivers/block/blk.h
+++ b/tlvc/arch/i86/drivers/block/blk.h
@@ -79,8 +79,6 @@ struct drive_infot {            /* CHS per drive*/
 };
 extern struct drive_infot *last_drive;	/* set to last drivep-> used in read/write */
 					/* used to handle QEMU bugs and quirks */
-extern int running_qemu;	/* for directhd.c and directfd.c */
-
 #ifdef CONFIG_BLK_DEV_BHD
 extern unsigned char hd_drive_map[];
 #define MAX_BIOS_DRIVES 2
--- a/tlvc/arch/i86/drivers/block/init.c
+++ b/tlvc/arch/i86/drivers/block/init.c
@@ -42,7 +42,7 @@
 
 int boot_rootdev;	/* set by /bootopts options if configured*/
 extern int boot_partition;
-int running_qemu = 0;	/* for directhd/fd */
+char running_qemu;	/* for directhd/fd */
 
 void INITPROC device_init(void)
 {
--- a/tlvc/arch/i86/drivers/net/ne2k.c
+++ b/tlvc/arch/i86/drivers/net/ne2k.c
@@ -24,6 +24,7 @@
 #include <linuxmt/debug.h>
 #include <linuxmt/netstat.h>
 #include <linuxmt/heap.h>
+#include <linuxmt/kernel.h>
 #include "eth-msgs.h"
 
 // Shared declarations between low and high parts
@@ -546,8 +547,7 @@ void INITPROC ne2k_drv_init(void)
 
 	j = k = 0;
 
-#define PRINT_EPROM
-#ifdef PRINT_EPROM
+#if 0
 	for (i = 0; i < 32; i++) printk(" %02x", cprom[i]);
 	printk("\n");
 #endif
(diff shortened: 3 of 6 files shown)


ghaerr:
The main.c overwrite bug fix looks good, but I found a calculation problem in exactly figuring when the envp_init array gets overwritten, changed (again) in https://github.com/ghaerr/elks/pull/1976. I also decreased the size of the very-seldomly seen panic text strings.

This won't affect the rabbit we're chasing on TLVC, so go ahead and commit this.

Mellvik:
Thank you, @ghaerr.
About the panics, I'm not sure a panic is the right choice when encountering bootopts errors. A panic makes it impossible for the user to fix the problem 'locally' on the platform. If the process to get to that point was cumbersome, that's unfortunate. 

Of course it is also unfortunate with a warning that scrolls off the screen. Thus I've been pondering a different approach, a delayed error message emitted just before passing control to process 1. That way the handling of oversized bootopts files could be made more  lenient too.

what's your take?

ghaerr:
> what's your take?

I had almost the same sequence of thoughts as you're having on this. Actually, the data overwrite issue occurs extremely rarely, and only occurred because ELKS was seeing the long hdparms= and le0= strings from a TLVC (non-ELKS) bootopts as environment variables rather than parsing them. So on almost every system, these issues will never arise.

I don't like panic, and was thinking of writing a bunch of extra code to deal with getting a keystroke, etc but this occurs extremely early in the kernel init. The previous code checked for args or envs > 8 and silently ignored them (worse), but didn't check the args string length overrun. 


I finally decided on using as little code and data as possible (the code is INITSEG but the data arrays[] stay in the kernel data segment forever), since these three error messages will almost likely never happen again. I do realize that means one might have to rebuild the image on a host, but we seem to do that a lot nowadays anyways.

> That way the handling of oversized bootopts files could be made more lenient too.

Yes, I'm pretty sure that message has scrolled by many times unnoticed. However, people usually will continue to play with settings until they get them right. I initially thought about making this a panic too and realized that would be a disaster. 

Ultimately, I don't think adding a bunch of extra code is worth much. If the a BIOS character wait were added, then the system gains one more dependency and could also just sit and not remotely boot, requiring human intervention anyways. Having a timeout doesn't really notify the user if they're not staring at the console during boot. So I figured in the rare case that /bootopts would cause corruption or not work as intended, its better just to stop the boot so the user know for sure somethings broken.

On another note, ELKS and TLVC bootopts aren't compatible, ELKS uses task=, TLVC tasks=, etc etc. So the whole idea of sharing is lost and copying a /bootopts file from one system to another will produce undesired results with no error messages at all. We can consider more of this issue after the long-stacked problem of getting the NIC drivers compatible is resolved, I would say! :)

Mellvik:
> 
> I don't like panic, and was thinking of writing a bunch of extra code to deal with getting a keystroke, etc but this occurs extremely early in the kernel init. The previous code checked for args or envs > 8 and silently ignored them (worse), but didn't check the args string length overrun.
> 
> I finally decided on using as little code and data as possible (the code is INITSEG but the data arrays[] stay in the kernel data segment forever), since these three error messages will almost likely never happen again. I do realize that means one might have to rebuild the image on a host, but we seem to do that a lot nowadays anyways.

My idea is to skip both the panic and the message - assuming to scrolls off the screen anyway (the smart ones film the boot sequence with their phones to get everything :-) ), and instead see a simple global int bit vector where a bit is set depending on which error happened. That vector can then be simply decoded and messages emittelike just before or after kernel_banner, ensuring its visibility.

As to rebuild versus letting it boot: On real hardware the situation is entirely different. A new image may take 5 or more minutes to build, transfer, copy, swapping floppies back and forth between systems, and in my case (which admittedly is rare) a trip to the garage and back. Letting `bootopts` errors softfail into a booted system (of course the error may be too severe to boot, but in most cases it isn't) is actually quite a big deal.

> 
> On another note, ELKS and TLVC bootopts aren't compatible, ELKS uses task=, TLVC tasks=, etc etc. So the whole idea of sharing is lost and copying a /bootopts file from one system to another will produce undesired results with no error messages at all. We can consider more of this issue after the long-stacked problem of getting the NIC drivers compatible is resolved, I would say! :)

Yes, there are a few incompatibilities and it's worthwhile working on that. I chose the plural forms `bufs=` and `tasks=`, sacrificing a byte in both cases, because  it's easier to understand - to new users in particular. The plural form invites a number, the singular form is less obvious. I suggest doing the same with all `bootopts` keywords expecting a number (inodes=, files=, xmsbufs= etc.) . And yes, I agree that this belongs under the 'general' (wow, sounds serious!) compatibility tab of our worklist. Anyway, think about it.

ghaerr:
> and instead see a simple global int bit vector where a bit is set depending on which error happened. That vector can then be simply decoded and messages emittelike just before or after kernel_banner, ensuring its visibility.

The error checking occurs in different functions, but I like the idea of later visibility. I'll look into this good idea also, with an attempt at minimal code addition.

PR #80, opened 2024-09-09
[direct floppy] Add XT floppy type setting to bootopts

Mellvik:
The direct floppy driver is reasonably smart about figuring out correct floppy types, probing alternatives in a circular fashion on read failures. If a read fails and the drivers is in 'probe' mode, the next alternative is tried until success or the list is exhausted. This reliance on a read failure to check the next, means that the physical format has to be different between the alternatives. Not perfect, but works in most cases - on AT class and later systems with floppy drives that typically supports 2 or more different formats. Earlier XT class systems don't need probing because only 360k drives are supported. That said, 320k format could easily be supported using the same mechanism.

This mechanism fails if different drives have the same (media) format but different geometry. Which is what happens when a 720k 3.5" drive is connected to a XT class system. The drive may even be a 1.44M 3.5" - most of them support the 720k format, which uses the same low level encoding as the 360k drives. The  only difference between the two - seen from the controller - is the number of sectors per track. This data format has the highest data rate these systems (the CPU, not the controller) can support.

Taking advantage of this 720k alternative became an option in DOS  3.3, but requires an extra driver (`driver.sys`) or config.sys setting, depending on DOS version.

In TLVC such support has been missing. One would think extending the probing mechanism already in place could fix that, but as the discussion above describes, it can not. In its default (360k) configuration, the controller will read the first 9 sectors off of a 720k drive just fine and the driver will conclude that everything is fine. Then, when attempting to access a sector number above 9, a `sector not found` error will occur. This error can be caught and a probe launched,  but it doesn't fix the problem: Reading the first 9 sectors successfully is enough to mount a 720k Minix file system. When mounted, the next access may be beyond the end of a 360k disk, which will be caught in `ll_rw_blk.c`, and the request will never reach the driver.

The next alternative is to do probing in the driver open routine. A few if-statements and a test-read. Possible but more code, more complications. A better - more flexible, less code - option is to add floppy type setting a la CMOS to `bootopts`.

This PR adds `xtflpy=n,n` to the list of supported directives in bootopts. Type numbers match standard AT CMOS types, 360 being 1, 720k being 3. The driver reports 'bootopts' instead of 'CMOS' during configuration if the `xtflpy=` directive is used.

Also included are some cleanups in `init/main.c` and a still unfinished version of the `calibrate_delay` function.

### DIFF
--- a/tlvc/arch/i86/drivers/block/directfd.c
+++ b/tlvc/arch/i86/drivers/block/directfd.c
@@ -53,7 +53,6 @@
  *
  * 2023/3/10 -- helge@skrivervik.com
  * Rewritten for TLVC to replace BIOS based IO from ELKS. 
- * DMA setup from Minix for simplicity and compactness. 
  */
 
 /*
@@ -63,7 +62,6 @@
  *   the odd case of using BIOS HD + DIRECT FD + XMS buffers + TRACK cache, 
  *   which really should not happen. IOW - use either BIOS block IO or DIRECT block IO,
  *   don't mix!!
- * - Update DMA code
  * - Clean up debug output
  * - The driver has many provisions for 4 floppy drives, but except for the oldest PCs
  *   (PC and XT) there is no physical support more than 2 drives except via a 2nd controller.
@@ -157,15 +155,15 @@ void (*DEVICE_INTR) () = NULL;
 #endif
 
 
-static int initial_reset_flag = 0;
+static int initial_reset_flag;
 static int need_configure = 1;	/* for 82077 */
-static int reset = 0;		/* something went wrong, reset FDC, start over */
-static int recalibrate = 0;	/* less dramatic than reset, like changed drive
+static int reset;		/* something went wrong, reset FDC, start over */
+static int recalibrate;		/* less dramatic than reset, like changed drive
 				 * parameters or seek errors. Will (eventually)
 				 * trigger a call to recalibrate_floppy() */
-static int recover = 0;		/* signal that we're recovering from a hang,
+static int recover;		/* signal that we're recovering from a hang,
 				 * awakened by the watchdog timer */
-static int seek = 0;		/* set if the current operation needs a track
+static int seek;		/* set if the current operation needs a track
 				 * change (seek) */
 static int nr_sectors;		/* # of sectors to r/w, 2 if block IO */
 static int raw;			/* set if raw/char IO	*/
@@ -231,7 +229,7 @@ static unsigned char reply_buffer[MAX_REPLIES];
 #define CMOS_2880k  5
 #define CMOS_MAX    5
 
-/* indices into minor_types[], used for floppy format probes */
+/* indices into floppy_type[], used for floppy format probes. Must match the table below */
 #define FT_360k_PC  1           /* 360kB PC diskettes */
 #define FT_1200k    2           /* 1.2 MB AT-diskettes */
 #define FT_720k     3           /* 3.5" 720kB diskette */
@@ -255,7 +253,7 @@ static struct floppy_struct floppy_type[] = {
 };
 
 /* floppy probes to try per CMOS floppy type */
-static unsigned char p360k[] =  { FT_360k_PC, FT_360k_PC, 0 };
+static unsigned char p360k[] =  { FT_360k_PC, FT_720k, 0 };
 static unsigned char p1200k[] = { FT_1200k,   FT_360k_AT, 0 };
 static unsigned char p720k[] =  { FT_720k,    FT_720k,    0 };
 static unsigned char p1440k[] = { FT_1440k,   FT_720k,    0 };
@@ -276,9 +274,14 @@ static unsigned char *base_type[4] = { NULL, NULL, NULL, NULL };
 /*
  * The driver is trying to determine the correct media format
  * while probing is set. rw_interrupt() clears it after a
- * successful access.
+ * successful access. This works when the formats are distinct,
+ * not when the format is the same and the # of sectors differ.
  */
 static int probing = 0;
+
+/* On an XT with a 720k drive, probing will not work because the first 9 sectors
+ * read correctly anyway, we need to fake the CMOS types via bootopts */
+extern int xt_floppy[];
 
 /* Prevent "aliased" accesses. */
 
@@ -1615,7 +1618,7 @@ static unsigned char * INITPROC find_base(int drive, int type)
     if (type > 0 && type <= CMOS_MAX) {
 	if (type == CMOS_2880k) type--;	/* force 2.88 to look like 1.44 */
 	base = probe_list[type - 1];
-	printk("df%d is %s (%d)%d", drive, floppy_type[*base].name, type, *base);
+	printk("df%d is %s (%d)", drive, floppy_type[*base].name, type);
 	return base;
     }
     printk("df%d is unknown type %d", drive, type);
@@ -1629,17 +1632,27 @@ static unsigned char * INITPROC find_base(int drive, int type)
 static void INITPROC config_types(void)
 {
     int at = sys_caps & CAP_PC_AT;
-    printk("Floppy drive(s) [%sCMOS]: ", at ? "" : "no ");
+    printk("Floppy drive(s)%s ", at ? " [CMOS]: " : "");
     if (at) {
 	base_type[0] = find_base(0, (CMOS_READ(0x10) >> 4) & 0xF);
 	if ((CMOS_READ(0x14) >> 6) & 1) {
 	    printk(", ");
 	    base_type[1] = find_base(1, CMOS_READ(0x10) & 0xF);
 	}
-    } else {	/* No CMOS, force 2 type 1 drives - for convenience. 
+    } else {
+	if (xt_floppy[0]) {	/* floppy types from bootopts */
+	    printk("[bootopts]: ");
+	    base_type[0] = find_base(0, xt_floppy[0]);
+	    if (xt_floppy[1]) {
+		printk(", ");
+		base_type[1] = find_base(1, xt_floppy[1]);
+	    }
+	} else {
+		/* No CMOS or bootopts, force 2 type 1 drives - for convenience.
 		 * Neither may exist, we have no way to find out. */
-	base_type[0] = base_type[1] = probe_list[FT_360k_PC - 1];
-	printk("df0, df1 set to 360k/PC (type 1)");
+	    base_type[0] = base_type[1] = probe_list[FT_360k_PC - 1];
+	    printk("df0, df1 set to 360k/PC (type 1)");
+	}
     }
     printk("\n");
 }
--- a/tlvc/arch/i86/drivers/block/directfd.c
+++ b/tlvc/arch/i86/drivers/block/directfd.c
@@ -1787,7 +1787,6 @@ void INITPROC floppy_init(void)
 #endif
     blk_dev[MAJOR_NR].request_fn = DEVICE_REQUEST;
 
-    config_types();
     err = request_irq(FLOPPY_IRQ, floppy_interrupt, INT_GENERIC);
     if (err) {
 	printk("Unable to grab IRQ%d for the floppy driver\n", FLOPPY_IRQ);
@@ -1812,7 +1811,7 @@ void INITPROC floppy_init(void)
 #else
     fdc_version = FDC_TYPE_STD;	/* force std fdc type; can't test other. */
 #endif
-    printk("%s: Direct floppy driver, FDC (%s) @ irq %d, DMA %d", DEVICE_NAME, 
+    printk("%s: Direct floppy driver, FDC %s @ irq %d, DMA %d", DEVICE_NAME, 
 	    fdc_version == 0x80 ? "8272A" : "82077", FLOPPY_IRQ, FLOPPY_DMA);
     configure_fdc_mode();
 #ifdef USE_DIR_REG
@@ -1831,6 +1830,7 @@ void INITPROC floppy_init(void)
 	reset_floppy();
     }
 #endif
+    config_types();
 }
 
 #if 0
--- a/tlvc/init/main.c
+++ b/tlvc/init/main.c
@@ -27,13 +27,11 @@
 #define ENV		1		/* allow environ variables as bootopts */
 #define DEBUG		0		/* display parsing at boot */
 
-#define MAX_INIT_ARGS	8
-#define MAX_INIT_ENVS	8
 #define MAX_INIT_ARGS	6	/* max # arguments to /bin/init or init= program */
 #define MAX_INIT_ENVS	12	/* max # environ variables passed to /bin/init */
 #define MAX_INIT_SLEN	80	/* max # words of args + environ passed to /bin/init */
 
-#define	__STRING(x)	"x"	/* TODO: move these to header files later */
+#define	__STRING(x)	#x	/* TODO: move these to header file */
 #define STR(x)          __STRING(x)
 
 /* bootopts error message are duplicated below so static here for space savings */
@@ -63,7 +61,6 @@ unsigned sys_dly_base;
 #ifdef CONFIG_PREC_TIMER
 void calibrate_delay(void);
 void u10delay(unsigned);
-unsigned int hptimer;
 void testloop(unsigned);
 #endif
 
@@ -72,6 +69,7 @@ void testloop(unsigned);
 int hdparms[CHS_ARR_SIZE];	/* cover 2 drives */
 
 int netbufs[2] = {-1,-1};	/* # of network buffers to allocate by the driver */
+int xt_floppy[2];	/* XT floppy types, needed if XT has 720k drive(s) */
 static int boot_console;
 static seg_t membase, memend;
 static char bininit[] = "/bin/init";
@@ -91,9 +89,13 @@ static int args = 2;	/* room for argc and av[0] */
 static int envs;
 static int argv_slen;
 /* argv_init doubles as sptr data for sys_execv later */
-static char *argv_init[80] = { NULL, bininit, NULL };
+#ifdef CONFIG_SYS_NO_BININIT
+static char *argv_init[MAX_INIT_SLEN] = { NULL, binshell, NULL };
+#else
+static char *argv_init[MAX_INIT_SLEN] = { NULL, bininit, NULL };
+#endif
 #if ENV
-static char *envp_init[MAX_INIT_ENVS+1];
+static char *envp_init[MAX_INIT_ENVS];
 #endif
 static unsigned char options[OPTSEGSZ];
 
@@ -144,6 +146,11 @@ static void INITPROC early_kernel_init(void)
     setup_arch(&membase, &memend);  /* initializes kernel heap */
     mm_init(membase, memend);       /* parse_options may call seg_add */
     tty_init();                     /* parse_options may call rs_setbaud */
+
+/*** Expermiental: Testing the upper 1k of memory for BIOS modifications ****/
+    byte_t __far *upper = _MK_FP(0x9fc0, 0); 
+    int i = 0;
+    while (i++ < 1024) upper[i] = i;
 
 #ifdef CONFIG_TIME_TZ
     tz_init(CONFIG_TIME_TZ);        /* parse_options may call tz_init */
@@ -198,13 +205,13 @@ void INITPROC kernel_init(void)
 #endif
 
     kernel_banner(membase, memend, s, e - s);
+#define CALIBRATE_DELAY
 #ifdef CALIBRATE_DELAY
     calibrate_delay();
 #endif
 #ifdef CONFIG_PREC_TIMER
-#ifdef TIMER_TEST
-    unsigned x = hptimer?hptimer:10000;
-    //timer_test();
+#if TIMER_TEST
+    unsigned x = 5000;
     testloop(x);
     testloop(x/10);
     testloop(x/100);
@@ -560,11 +567,10 @@ static int INITPROC parse_options(void)
 			parse_parms(2, line+8, netbufs, 10);
 			continue;
 		}
-#ifdef CONFIG_PREC_TIMER
-		if (!strncmp(line,"hptimer=",8)) {
-			hptimer = (unsigned int)simple_strtol(line+8, 10);
+		if (!strncmp(line, "xtflpy=", 7)) {
+			parse_parms(2, line+7, xt_floppy, 10);
+			continue;
 		}
-#endif
 		if (!strncmp(line,"TZ=",3)) {
 			tz_init(line+3);
 			/* fall through and add line to environment */
@@ -594,12 +600,11 @@ static void INITPROC finalize_options(void)
 	int i;
 
 	/* set ROOTDEV environment variable for fsck in /etc/mount.cfg */
-	if (envs < MAX_INIT_ENVS)
-		envp_init[envs++] = root_dev_name(ROOT_DEV);
-	if (running_qemu && envs < MAX_INIT_ENVS)
-		envp_init[envs++] = (char *)"QEMU=1";
-	if (envs >= MAX_INIT_ENVS)
+	if (envs + running_qemu >= MAX_INIT_ENVS)
 		panic(errmsg_initenvs);
+	envp_init[envs++] = root_dev_name(ROOT_DEV);
+	if (running_qemu)
+		envp_init[envs++] = (char *)"QEMU=1";
 
 #if DEBUG
 	printk("args: ");
@@ -669,7 +674,7 @@ static char * INITPROC option(char *s)
 }
 #endif /* CONFIG_BOOTOPTS*/
 
-#ifdef TIMER_TEST
+#if TIMER_TEST
 void testloop(unsigned timer) 
 {
 	unsigned long pticks;
@@ -679,7 +684,7 @@ void testloop(unsigned timer)
 	get_ptime();
 	while (x--) outb(x, 0x80);
 	pticks = get_ptime();
-	printk("hptimer %u: %lk (%lu)\n", timer, pticks, pticks);
+	printk("timer %u: %lk (%lu)\n", timer, pticks, pticks);
 #else
 	while (x--) {
 		outb(0, TIMER_CMDS_PORT);       /* latch timer value */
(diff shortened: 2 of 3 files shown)


ghaerr:
> The only difference between the two - seen from the controller - is the number of sectors per track.

> In its default (360k) configuration, the controller will read the first 9 sectors off of a 720k drive just fine and the driver will conclude that everything is fine.

Can't this problem be solved by just probing for the 720k disk first in the probe list, which will try reading the higher sector number and fail or not, producing the right outcome?

Or is this something to do with the CMOS indicating 360k only and no 720k probe is taking place?

Mellvik:
That's the alternative I discussed in the text, mentioning we can do that in the open routine. It's not all that complicated although creating a read (the probe needs a read failure to trigger the change) inside the driver is nontrivial, but it's quite a bit of extra code. Thus the chosen path which to me is simpler and more flexible.

Also of some importance is the fact that this combination is rare. I cannot remember having seen it in a commercial system.

ghaerr:
Forgive me, but I'm not quite following your description of how the driver works. I'm looking at ELKS directfd.c, but I'm pretty certain TLVC is almost exactly the same: The floppy_init never does any probing or floppy I/O, all probing and floppy I/O requests are done during the next I/O request *after* floppy_open().

> although creating a read (the probe needs a read failure to trigger the change) inside the driver is nontrivial

There's no need to "create a read", as the probing occurs during a normal driver read request whenever probing =1 and current_type[drive] == NULL. (If current_type[drive] were always set to 0 in the close routine, then the driver would always auto-probe again every time it were opened).

So I don't know if the problem you're describing is because a 720k floppy was inserted after a 360k was previously used, or whether the probe routine just won't recognize the 720k in the first use since boot.

My understanding is that a probe starts only when the a block I/O read/write request is queued, and the global `probing` is > 0.
The "probing" then proceeds to try different floppy formats but doesn't actually "probe" using the highest allowed sector number for that format. Instead, it seems to just initialize the controller for the probing floppy type then continue with the already-queued I/O request on the actually requested block, after converting it to CHS. But even this normally results in a conversion to full track read - which could compound the issue at hand but also require that the floppy formats be ordered by highest max sector first. So it seems that the DF probing is quite a bit different than the BIOS floppy driver probing, and doesn't actually ever issue a special probe I/O request!

I'm writing this just to try to fully understand how I think the DF driver works and understand the real root of your issue.

Thanks!

Mellvik:
Hi @ghaerr, thanks for the interest. 
Your description of how it works is correct. I'm rereading the description provided with the PR, and I still think it details why 'regular probing' doesn't work (and cannot work). Having written it, that's no surprise, but I urge you to read it again. The short explanation being there is no error to trigger a probe. Thus a different approach is required.

Like a read initiated from `open` (complicated) or `bootopts` (simple) ...

Unless, of course, I'm missing something.

Mellvik:
@ghaerr, maybe I misunderstood your first post. Setting the default type to 720k for pre-AT machines (instead of the current 360k) and let it fail 'downwards' is an interesting idea. I didn't think about that one, it should work.  Since we're talking different physical drives, not different media densities in the same drive, there will never be a need to 'probe' back up to 720k if 360k was detected Kinda ugly since the 720k variant is so rare, but ditching the auto-detect message would (for this machine type) hide it. I'll look into that. Thanks.

Mellvik:
OK, I did some testing and thinking about this alternative, which sort of works but it does require some redesign that may not be desirable.
- Mounting and using a 360k file system works fine as it should even though the system thinks the device is 720k.
- Reading a 360k disk fails eventually but not like I expected. The driver is reading full tracks and not noticing that half the track is missing until later. Subsequent runs of the same `cat /dev/df1 > /dev/null` deliver errors on different sectors, and I don't yet quite understand why.
- Deciding to investigate further, the raw driver is conveniently at hand, which delivered a different kind of surprise:
```
tlvc16# dd if=/dev/rdf1 of=/dev/null bs=10b

panic: Divide by zero

SYSTEM HALTED - Press CTRL-ALT-DEL to reboot:df1: Auto-detected floppy type 720k
```
... which started a different rabbit hunt (still hunting).

Anyway, what I'm seeing (and forgot before) is that unless the driver is in probe mode, errors will be reported as what they are, not as a trigger for new probes. This can be fixed, but it needs more complicated logic - i.e. code. This is a different type of situation than the probe mechanism was made for, as I also pointed out on the initial PR description. The probe mechanism detects format, not geometry. The difference is profound in that in the first case, access will fail until it's right, in the second case it doesn't fail unless something is out of bounds, which means it fails later, maybe never.

As to the divide by zero error, I'll be back on that. Fascinating. What did we do before we had the trap installed? Probably a couple of bugs in the driver.

ghaerr:
> but I urge you to read it again. 

> This mechanism fails if different drives have the same (media) format but different geometry.
> The only difference between the two - seen from the controller - is the number of sectors per track. 

After rereading your well-written initial explanation, and also realizing I keep forgetting that the DF driver probe routine is entirely different than the BIOS probe in that it does not perform max sector number probing, I think I confused the issue by suggesting to reorder the "probe" list. However, as I had thought possibly might happen, and you later tested, the driver probe routine actually ends up reading full tracks, with an error thrown only if/when an I/O request comes in with an invalid sector number for the drive having been selected. I'm not sure I like this full sequence of probe events, but the probe routine is tied into doing an initial probe on the back of a submitted I/O request, rather than performing a read on its own during driver open (as you've also noted).

In short, the DF probe routine is somewhat brittle, and also doesn't "probe" in the way the BIOS driver does, which allows the kernel to think it's got a floppy CHS that actually isn't valid. The allowed CHS mismatch compared to what is guaranteed by the BIOS driver is probably not good, and fixing that would require reworking the entire probe code, which probably isn't worth it at this time. You came to this conclusion in your first comment.

I completely agree with all your recent comments, especially the summary that the driver is probing format, not geometry, which delays errors. While developing my own version of the driver, I had also considered ditching the auto-probe message for a clean display, but decided against it as I thought better to see what the driver thinks is happening.

I would guess the reason for `cat` displaying different sector errors has to do with the async nature of of the DF driver combined with track reading, but not sure. I don't think the driver is buggy, but I have noted that errors can happen because the driver doesn't actually do a max sector probe.

Anyways - sorry the confusion, but nice discussion - and can see why a simple /bootopts solution is preferable at this point.

Mellvik:
The story goes on - new chapter. Turns out there IS an elusive bug in the driver, related to reading 360k drives. Easy to reproduce, hard to identify. Any sequence of reads with a physical block size that is not a multiple of the track size will eventually fail. Sometimes early, like after a few reads, some times later, and rarely in the same spot. Using the block device always works thanks to the track cache - all reads are track size, so the raw driver (and `dd bs=` ...) is required to reproduce the error.

the error manifests itself as a hung read. Dma setup completes, the read initiated, but the operation complete interrupt never arrives. Eventually the watchdog timer kicks in and aborts the transfer. 

Seems I need to reread the fine print in the nec765 controller docs again, and refresh the story behind the tubo/no tubo modes on vintage systems. It did have to do with floppy compatibility. Fortunately the drive is beyond suspicion - brand new as in previously unused, a Qume identical to the one used in IBMs 5155 and 5160 models. 

The problem may be a interrupt (protection) issue, the operation being somehow disturbed during initiation. Or a delay loop running too fast ... The system is my V20 8mhz xtclone.

ghaerr:
> Any sequence of reads with a physical block size that is not a multiple of the track size will eventually fail. 

Other than a bug in your char driver, could this be attributed to the inability of the FDC to switch tracks when the final 1K block in the multi-sector request that overlaps tracks? There may be some incompatibility in that operation between new an old FDCs, and a lot of effort was added in the BIOS floppy driver to never attempt a read where the track increments (along with a DDPT update to the BIOS didn't think it needed to if it allowed such operations).

Mellvik:
Unfortunately, those alternatives are in the clear (the char driver is only a shell calling the block driver).

I believe I've run the same tests on all other drive types, but I need to redo that ti make sure. If so, it's timing...

ghaerr:
> those alternatives are in the clear (the char driver is only a shell calling the block driver).

Well, you said the block driver never fails but the char driver does. The difference is that the block driver will always start on a block beginning, whereas the char driver can be asked to start on an arbitrary sector boundary. This might cause a different start/end sector difference since sectors per track (9?) is not an even number. It might be worth running tests on another drive type with an odd sector number count?

Does this ever happen on QEMU?

ghaerr:
> tlvc16# dd if=/dev/rdf1 of=/dev/null bs=10b
> panic: Divide by zero `

BTW, I executed the following script to find every case where a `divw` or call to __{u}divsi3 or __{u}modsi3 is called in the kernel source directory:
```
find . -name "*.o" -print0 | while read -d $'\0' file
do
    ia16-elf-objdump -D -r -Mi8086 $file 2>/dev/null | grep "divsi\|modsi\|divw" && echo $file
done
```
(It displays the filename *below* the matched instructions), but interesting to see where the kernel needs help! I plan on looking at this in more detail to speed up other kernel operations if it makes sense. Printk.c has five (!) calls itself, which once eliminated will definitely speed things up!

Then looking at the directfd.c driver, there are only two DIV instructions in the entire driver, which are used to compute the following expressions:
```
    sector = start % floppy->sect;
    tmp = start / floppy->sect;
    head = tmp % floppy->head;
    track = tmp / floppy->head;
```
So this should give you a starting point, that could be related to the driver bug you've just found - as there should be no divide by zero unless floppy->head or sect is getting trashed...

It was nice to see that GCC is producing code that uses both the quotient and the remainder (modulo) from a single DIV instruction, thereby making the above code pretty fast, all-in-all. Of course, should the values become unsigned longs, that's a different story!

Mellvik:
> It was nice to see that GCC is producing code that uses both the quotient and the remainder (modulo) from a single DIV instruction, thereby making the above code pretty fast, all-in-all. Of course, should the values become unsigned longs, that's a different story!

Agreed, that's encouraging. And thanks for the help with mapping out the DIVs. I've spent some more time on the issue, including remounting the floppy drive which - for reasons unmentionable - was mounted upside down. This changed the behaviour somewhat, different frequencies etc. but the pattern is the same. One exception: Raw block reads may fail even when using 9 or 18 sector block sizes. 'May' is important, it's rare, but it happens. While it never happens when using the track cache (block driver), which may be a characteristic of the bug I'm on the trail off, or because there is a significant delay between each read operation. Watching the traces roll over the screen, it's easy to get a feel for the timings of some of these operations. Raw reads are back-to-back, block reads are fill-buffer-then-read-buffer-until-its-empty-then-new-read.

It seems to me we have a memory leak/stack corruption issue (wow does that sound familiar), which makes some of the rather odd behaviours explainable. One of them is the Div by zero trap when certain block sizes are used, another that when the timeout occurs and calls reset, the wrong interrupt handler gets called. So there is a pattern even to this, which I'm tracing. 

What I'm thinking is that the leak is in error processing somewhere. As long as everything is fine, we're good. When there are errors (the less banged-up-area of the driver), we may trigger the error. And 360k subsystems (5.25in drives and media) always have errors and retries. 3.5in drives (almost) never do.

To be continued. Another rabbit hunt possibly, unless it's just a bug...

Mellvik:
@ghaerr - your morning coffee update (entertainment?) is late, but then again ...

Unsurprisingly, it turns out I'm dealing with at family of rabbits, or at least more than one. The first is in custody - as it turns out, an old friend, or at least acquaintance: the DMA 64k boundary. Detected by the driver, but only for 1k (block mode) transfers. So what happened when the DIV0 trap popped up was that lots of data poured into a 1k bounce buffer, overwriting anything in its way. Fixed by taking advantage of the fact that the track buffer was hitherto unused for raw access.

What's interesting is why the problem appeared only when using 360k floppies. I went so far as to redefine the 720k drive to 360k (by simply changing the drive type in bootopts), and voila, now this drive had the problem too. Why would a change that essentially shifts an index in a driver table cause an application data segment to shift that much (or at all), and consistently? I don't have the answer to that one, maybe you have an idea. 

Unfortunately, the fix somehow affected the loading address of the `dd` application (again), so I haven't been able to recreate the 64k wrap scenario. That opportunity will eventually come while chasing the rest of the (rabbit) family. Like the hang problem, which turns out to be unrelated to the DMA wrap issue. 

The chase continues.

ghaerr:
> Fixed by taking advantage of the fact that the track buffer was hitherto unused for raw access.

Are you saying that the problems resulting from DMA wrap are now fixed, for both 360 and 720k (all) drives, and that just the hang problem remains?

I ended up fixing this issue on ELKS by separating out the lowest level (BIOS, in this case) floppy read/write routine, which always handle DMA wrap by using DMASEG, and always calling that routine from any upper levels, including the track cache read routine(s). The good old mandatory hierarchy of routines.

> Why would a change that essentially shifts an index in a driver table cause an application data segment to shift that much (or at all), and consistently? 

Could that have something to do with the DF driver probe routine always initially reading a full track, rather than just a sector? So the first element in the probe array would have caused the overwriting just like it did when 360k was the first probe drive type.

> the fix somehow affected the loading address of the dd application (again), so I haven't been able to recreate the 64k wrap scenario

I see - because `dd` is requesting raw char driver I/O to its own buffer? That can likely be easily addressed by adding other char arrays either before or after the main buffer in dd.c, although it might take a moment to discover what size is needed to get the buffer in the wrong place again.

Mellvik:
> > Fixed by taking advantage of the fact that the track buffer was hitherto unused for raw access.
> 
> Are you saying that the problems resulting from DMA wrap are now fixed, for both 360 and 720k (all) drives, and that just the hang problem remains?

Yes. Although there are other problems than the hang (which, for the record, is a transaction hang, not a system hang). Like 'random' (and 'unrecoverable') 'sector not found' errors on sectors that read fine the next time. 
> 
> I ended up fixing this issue on ELKS by separating out the lowest level (BIOS, in this case) floppy read/write routine, which always handle DMA wrap by using DMASEG, and always calling that routine from any upper levels, including the track cache read routine(s). The good old mandatory hierarchy of routines.

Good point. The 'extra' bounce buffer is a leftover from whatever back in the day and no longer meaningful since the track buffer is always available in the cases where the bounce buffer is being used. Which brings up an unrelated point about the usefulness of the track buffer in general, more about that below.
> 
> > Why would a change that essentially shifts an index in a driver table cause an application data segment to shift that much (or at all), and consistently?
> 
> Could that have something to do with the DF driver probe routine always initially reading a full track, ...

There are no independent probe actions. The driver does nothing until requested to do so from above. If the first access fails, it probes the next alternative and so on. Which brings up a different point, discovery actually. The motor startup time is set to 0.5secs. This is less than required for many drives, old 360k drives in particular, which require more like a second. This is a calculated 'error', as in 'let it fail, it will succeed the second time' (per Hermann Sieb, the floppy guru). Which is exactly what happens, but it doesn't go all that well with the probing: If the driver setup is for the 360k to switch to 720k as the next in line, the latter will always be selected. For newer and higher capacity drives, the .5 secs seems to work fine, so instead of changing the probe, the easy way out is to not have a second alternative for the 360k to switch to. What's left (still) is the 'Auto-detected floppy type'. To be fixed.

> 
> > the fix somehow affected the loading address of the dd application (again), so I haven't been able to recreate the 64k wrap scenario
> 
> I see - because `dd` is requesting raw char driver I/O to its own buffer? That can likely be easily addressed by adding other char arrays either before or after the main buffer in dd.c, although it might take a moment to discover what size is needed to get the buffer in the wrong place again.

This is what raw access is - copy data directly to the application data space, so that part has nothing to do with `dd` really. `dd` is just the perfect tool to test IO because block sizes can be chosen freely on the command line. Up to whatever is available in its heap, typically 11k with the current settings.

About the track buffer: I've been a strong supporter of the track buffer since forever, or at least since the days of the creation of the `fdtest` utility. To the extent that I implemented full cylinder buffering in the directed driver. Only to disable it because it was slowing down the system. Now I'm on a path to change my mind entirely. Having looked at tons of debug traces from the driver running 'regular usage' of the floppy, it seems to me the track buffer - just like the full cylinder buffer - is lowing down the system rather than helping. Because discards the buffer without using more than the initial 2 sectors more often that it's benefitting from them. (The physical system in question was booting and running off of a 720k floppy for weeks until I installed an IDE-CF card in it, an educating experience in several ways.)

This, of course, is mounted file system usage. For archival usage and data copying with lots of sequential IO, the situation is very different, so it ends up being a question of device usage.

Continuing along the same chain of through, my thinking is that adding (more like activating, since the code is there) buffer read-ahead may be an interesting exercise. Possibly not useful unless the IO drivers are interrupt driven, but TLVC is close, with `directfd` and MFM (`xd`) in place and `athd` close.

Raw/char IO is per definition as direct as physically possible between device and application, and likely used by people who know why they're using it. Which means choosing optimal block sizes for the occasion, such as accessing floppies a track at a time.

What do you think?

ghaerr:
> because dd is requesting raw char driver I/O to its own buffer? 

> copy data directly to the application data space, so that part has nothing to do with dd really.

My point was that the raw char driver I/O directly to an application buffer might have triggered the DMA wrap issue since the normal system buffers would not be used if file offset and sector alignment permitted it.

> What do you think?

WRT the track buffer? I had thought we showed, way back then, that it did, in fact, provide a speed improvement while booting. And there was some discussion about whether the extra track read would actually take up "much" extra time, since the head was already in position. Admittedly though there are a number of issues which have not been measured: the time it takes for the sector to become under the read head while positioned on the track, versus reading sector 1-max, whether the extra track cache data is ever used, and whether there are intervening writes, which of course invalidate the cache entirely.

So agreed, its performance is quite dependent on what the user is doing. But fast boot is nice. Perhaps now that we have working precision time measurement capabilities, this could be timed, starting from start_kenrel just before /bin/init is exec'd through (somehow) the shell prompt being displayed. And of course all this begs the measurement of our buffering system, and what the optimal number of system buffers should be. (Remember that the track cache itself never makes it into the system buffers unless the blocks are actually used, and the track cache is using a valuable 9k (19k on TLVC?) of main memory.

For the read-ahead idea, that might be useful, but I would suspect it would be limited on floppies - and also potentially destroy any benefit of the track cache, when a read-ahead is outside track cache. And would a full track be read on a single readahead buffer request? Lots to think about... but better to measure first so we know what we're up against!

> it seems to me the track buffer - just like the full cylinder buffer - is slowing down the system rather than helping. 

I can see how if a track buffer might slow the system down, a full cylinder cache could add to the problem. Perhaps the solution, now that you're running entirely async I/O, would be to not schedule the I/O complete based on an entire track or cylinder, but instead start the I/O on the requested block only, then when completed allow the application to run, while additionally scheduling the remainder of the track (or cylinder) to be read asynchronously. With DMA, it would seem the CPU load would be minimal, and any application waits negated. This could also work as a replacement for adding read ahead at the upper buffer level layer. With some doing, the track cache buffers could be reconfigured as EXT system buffers, which could be locked into place for the track I/O (or perhaps just schedule 9 separate I/O requests to each buffer, using the multi-bh code that isn't quite working in ll_rw_blk.c).

Mellvik:
> So agreed, its performance is quite dependent on what the user is doing. But fast boot is nice. Perhaps now that we have working precision time measurement capabilities, this could be timed, starting from start_kenrel just before /bin/init is exec'd through (somehow) the shell prompt being displayed. 

That's a good idea. It seems to me the easy way to do this is to print jiffies as start_kernel is called, and the - from `init` just before `getty` is called, simple now that the jiffies ioctl is in place. On a XT that will be quite a few seconds. I agree that fast boot is important, so this is an important parameter to consider.

And of course all this begs the measurement of our buffering system, and what the optimal number of system buffers should be. (Remember that the track cache itself never makes it into the system buffers unless the blocks are actually used, and the track cache is using a valuable 9k (19k on TLVC?) of main memory.

9k it is on both systems, and that was the idea behind this line of reasoning, including mentioning the read-ahead: getting rid of the track cache, saving the RAM. The cylinder cache was implemented, then discarded for performance reasons. It's still in there, but commented out.

> Perhaps the solution, now that you're running entirely async I/O, would be to not schedule the I/O complete based on an entire track or cylinder, but instead start the I/O on the requested block only, then when completed allow the application to run, while additionally scheduling the remainder of the track (or cylinder) to be read asynchronously. With DMA, it would seem the CPU load would be minimal, and any application waits negated. This could also work as a replacement for adding read ahead at the upper buffer level layer. 

An interesting idea (although killing the discarding of the track buffer). It's also a low threshold thing, I might just test that. Of course, if the track buffer isn't all that useful in the first place, this isn't very useful either, but if it comes as 'almost for free' (performance wise and implementation wise) it may just be worthwhile.

> With some doing, the track cache buffers could be reconfigured as EXT system buffers, which could be locked into place for the track I/O (or perhaps just schedule 9 separate I/O requests to each buffer, using the multi-bh code that isn't quite working in ll_rw_blk.c).

Again, that should be workable (on AT system and up). As to scheduling requests, we're running low on request headers already, we'll have to look at the consequences of that.

Thanks - great discussion.

ghaerr:
BTW, on another matter of floppy boot speed and general performance - some time ago, I think when fixing one of your reported `mkfs` issues, I happened to notice that the exact order of creating of /dev/ entries as well as the order of copying files (creating inodes) when creating the boot floppies mattered. That is, if the inode is in one of the "early" filesystem blocks, it got buffered and was found in the L1/L2 cache for subsequent boot activities. I noticed that one of the /dev/tty entries had been created with a high inode number - meaning the floppy had to seek tracks just to open the console! I think I fixed that, but can't remember. We have the same problem with various primary utilities and files opened (and intermediate directories searched)  including /bin, /bin/init, /etc, /etc/inittab /etc/rc.sys, /dev, /dev/tty1, /dev/console, etc etc. Getting rid of /usr/bin in the early days helped because accessing /usr caused another 1K disk read (possibly flushing the track buffer).

So as always, the devil's in the details! Track buffer or file read-ahead may not be nearly as important as the ability to create the inodes and file data near each other as well as possibly flagging a directory or file for increased priority in system caches.

Mellvik:
Good point, @ghaerr. And by the same token, booting fast is important, fast regular operations is even more important.So finding the balance etc. ...

About inodes and placement of files: Very good reminder. IIRC, this is probably why early unixes has system and startup utilities in /etc along with the configuration files - I'm sure they had a means of optimizing this particular directory for fast boot, leaving the rest of the standalone (as in runlevel 1) programs in /bin etc etc etc.

We're up for some fun figuring this out. After I find the last (ha..) rabbit.  BTW what I said about motor startup time was not correct. When making reads very slow (like pipe dd through hd) the motor stops between reads and there are no errors. There is more to learn.

Mellvik:
Rabbit update: The final (!) rabbit was on that has escaped from custody, IOW self-inflicted but (aren't they all really?): I had increased the step rate parameter (from 0xD to 0xF milliseconds) early on as an experiment with no effect at all. Reversing that change eliminated the read error, the read hang and the sector 0 re-read problems. Weird. Why would increasing it have such effect?

The bible on FDC timing issues claims 8ms should be safe for all drives. I haven't tried that value on the 360 drive, but it works well for the 720k drive (which is really a 1.44M drive). BTW the same bible claims 3ms should be fine for modern 1.44M drives. Whether this has any effect on access times, I don't know. Probably not or unmeasurable. 

Anyway, it means other more experimental issues we discussed above are coming up. About time.

ghaerr:
> increased the step rate parameter (from 0xD to 0xF milliseconds) 
> Why would increasing it have such effect?

I'm a bit confused by exactly what the Step Rate Time SRT) actually does on the FDC, but i think it specifies the time to wait between stepper motor pulses. However, the comment in directfd.c as well as the NEC 765 datasheet shows that the SRT is in the reverse (i.e. 0xF -> 0x01) order:
SRT


If this is a wait time instead of a speed, then changing the SRT from 0Dh to 0Fh actually decreases the wait time, increasing the step speed. So your drive probably couldn't handle the increased step speed (lower wait time).

> The bible on FDC timing issues claims 8ms should be safe for all drives.

So that value happens to be the same regardless of SRT interpretation: 8. All others are 16 - SRT.

</rabbit>

> it means other more experimental issues we discussed above are coming up. About time.

Definitely!

Mellvik:
I agree, it's somewhat confusing - even the name - Step Rate Time, wtf is that? Its either step rate or step delay, right? However, since the metric is ms, I figure it's reasonably clear - we're talking delay (at least that's the way I've always read it). So, the older the drive, the slower it is and the longer delay. Thus my surprise that an increase (increased slowdown/delay) (by a meager 2ms) would have such effect. 

BTW - the ulitmate floppy driver source puts it this way  ( https://wiki.osdev.org/Floppy_Disk_Controller )
```
SRT = "Step Rate Time" = time the controller should wait for the head assembly to move between successive cylinders. A reasonable amount of time to allow for this is 3ms for modern 3.5 inch floppy drives. A very safe amount would be 6 to 8ms. To calculate the value for the SRT setting from the given time, use "SRT_value = 16 - (milliseconds * data_rate / 500000)". For a 1.44 MB floppy and 8ms delay this gives "SRT_value = 16 - (8 * [500000 / 500000](tel:500000 / 500000))" or a parameter value of 8.
```
I've changed the value for the 3.5in drives to 8 and it works just fine (and left the 360k drive at 0xd). I doubt that it makes any practical difference (the reduction) though.

Mellvik:
I'm merging this PR now, to create a new one for the upcoming fixes and optimizations. There are some interesting numbers in the benchmarks.

Mellvik:
Oh, I see what you mean, that the setting should be (0xf-delay[ms]). It's really hard to read the 765 docs that way, but the 'bible' may be read that way. Certainly the experience with the 360k drive setting points in that direction. OTOH - why is the setting (in the driver) lower for the 1.44M drive, which is definitely faster than the 5.25in drives? And lower again for the 2.88 (which I have removed btw)? Even more confusing - like I said I changed the setting to 8 for the 720k drive (really a 1.44M drive) and it works fine, but it's making more noise than before when seeking (I just noticed, I'm normally not in that room...).

To be sure, I guess I have to step through each value and see what works and what doesn't (while listening).

ghaerr:
> Oh, I see what you mean

Yes - that's why I included the screenshot of the 765 data sheet: it doesn't explain it well, but does show that the F = 1ms, while D = 3ms (delay). So at least this answers your question of why your drive had failures when you decreased delay by increasing the SRT actual value.

Try not to wear out your stepper motors lol!

ghaerr:
> OTOH - why is the setting (in the driver) lower for the 1.44M drive, which is definitely faster than the 5.25in drives? And lower again for the 2.88 (which I have removed btw)?

Geez - I hadn't noticed that. I'm going to leave that one for you (especially since I'm definitely not going to chase down SRT compatibility on QEMU!!!! :)

Mellvik:
> Yes - that's why I included the screenshot of the 765 data sheet: it doesn't explain it well, but does show that the F = 1ms, while D = 3ms (delay).

I missed that and can't find it, where is it?

ghaerr:
Screen Shot 2024-09-19 at 11 02 07 AM

Mellvik:
Sh.., blindness come in many variants. I have the same document, apparently different version, the parens missing. No wonder I couldn't find it. 
IMG_0435

ghaerr:
Wow. So it looks like keeping multiple copies/versions of chip data sheets is a good idea (after all). I just happened to quickly find that version on the web. Note the confusion between your and my version "must be defined for each of the four drives" vs "stepping rate applies to all drives".

With a bit more research, the [82077AA Datasheet](http://www.osdever.net/documents/82077AA_FloppyControllerDatasheet.pdf) section 5.2.6 has both text and a table for the required SRT values based on drive type. As I suspected, some magic: "Note that the spacing between the first and second step pulses may be shorter than the remaining step pulses􏰃." So at least for some controllers, its not actually a stepping rate but a special number for each drive type!

Mellvik:
Ahh, the beauty of old stuff. i really need to clean my docs archive! 

All the more important (or at least interesting) to run through the values on my 4 physical drives - while listening carefully of course:-) the 1.44 and 2.88 settings in the driver are obviously wrong.

Mellvik:
The 82077AA data sheet is enlightening. Finally the formula in the 'OSdev wiki' is understandable - and the SRT values in the driver drive table make (some) sense. Who would have guessed that the SRT value has a different metric depending on the data transfer rate? Although that's what the formula indicates ...

Anyway, using the formula and the data rate, the current setting for the XT drives (720k, 360k), 0xD, means 6ms, which is good. For the 1.2M and 360k in 1.2M drive, the rates 0xD adds up to mean 5ms, also fine but a little high. For the 1.44M drive (data rate 500k), the 0xC value means 4ms which is also high, the recommendation is 3ms, so 0xD would seem better. For 2.88M and 1M data rate, the number (0xA) adds up to 3ms, which is good.

Given this understanding I'm not going to spend much time testing various values, just change the 1.44M SRT to 0xD (3ms) and test 3ms with the 720k drive - which ironically is impossible since the SRT value would be 14.5. So I tested both 0xF (ca. 2.5ms) and 0xE (ca. 3.5ms): Both work well (almost no sound from the drive while seeking) - and provided a chance to see if the difference affects performance in a measurable way. It does! The difference in boot time, measured as suggested earlier in this thread, is 3% - 80 jiffies. Practically insignificant but still interesting as in demonstrating that tuning - even at this level - has some effect. I'll keep the slow setting though, taking into account that a 'real' 720k drive (not a 1.44M like in this case) may need a little extra.

More about the boot time and other measurements/benchmarks later.

PR #82, opened 2024-09-21
[sys_utils] Fixed getty host name display

Mellvik:
In order to display the host name as part of the login prompt, `getty` used to pick the host name from the first line of `/etc/hosts`, a relic from before TLVC/ELKS had TCP/IP and a 'real' `/etc/hosts` file.

The fix deletes the old code and uses the HOSTNAME environment variable instead. More reliable and simpler than using the 'new' `/etc/hosts`.

Also included is a compile time option to print the jiffies value at startup time, a part of a time measurement mechanism (debug) in which the kernel init (`init/main.c`) prints the starting value, `getty` the ending value.

### DIFF
--- a/tlvccmd/sys_utils/getty.c
+++ b/tlvccmd/sys_utils/getty.c
@@ -47,6 +47,9 @@
 #include <stdarg.h>
 #include <errno.h>
 #include <paths.h>
+#include <linuxmt/mem.h>	/* /dev/mem */
+#include <linuxmt/memory.h>	/* _MK_FP */
+#include <sys/ioctl.h>		/* for jiffies */
 
 #define DEBUG		0	/* set =1 for debug messages*/
 
@@ -56,7 +59,8 @@
 /* For those requiring a super-small getty, the following define cuts out
  * all of the extra functionality regarding the /etc/issue code sequences.
  */
-//#define SUPER_SMALL		/* Disable for super-small binary */
+#define SUPER_SMALL	0	/* Enable for super-small binary */
+#define BOOT_TIMER	0	/* Enable to print jiffies at startup */
 
 #if DEBUG
 #define debug		consolemsg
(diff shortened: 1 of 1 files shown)


ghaerr:
You might also be interested in [this bug fix](https://github.com/ghaerr/elks/pull/2021).

Are you getting rid of the /etc/hosts functionality in total, or just using HOSTNAME= instead of reading /etc/hosts? I'm not clear on exactly what the benefit is here, as I think /etc/hosts is still required in certain libc routines for converting hosts to IP  addresses or vice versa.

Mellvik:
Thanks for the link/fix, I'll certainly include that!

This is a bugfix and an optimization. Got rid of a big (!) static buffer and a bunch of code in addition to the fix itself. Getty made wrong (ancient) and clumsy assumptions about the hosts file. This is simpler, and /etc/hosts lives on.

tlvccmd/sys_utils/getty.c
```diff
@@ -86,28 +90,16 @@ void consolemsg(const char *str, ...)
...
-	ptr--;
-    *ptr = '\0';
-    if (!*Host)
-	strcpy( Host, "LocalHost" );
+    char *host = getenv("HOSTNAME");
+
+    if (host == NULL)
+	strcpy(Host, "LocalHost");
+    else
+	strcpy(Host, host);
```
ghaerr, line 102: If you're trying for min executable size, replace line 100 strcpy with initialization of char Host[] with Localhost via declaration statement. 

Line 102 strcpy is dangerous as length is not checked. Recommend `strncpy(Host, host, sizeof(Host)-1)`.

Mellvik: Thanks, this is good. And yes, definitely going for min size. Along with `init`, `getty` is the one program always running. A byte here, a byte there...

PR #86, opened 2024-09-23
[directhd] Fixes and updates

Mellvik:
A number of updates to the `directhd` driver:
- probe enhancement
- cleaned, consistent and fewer boot messages
- improved XTIDE support
- moved driver specific header file from `linuxmt` to `arch`
- changed bounce buffer allocation from static to heap
- removed zero initializers from static variables

### DIFF
--- a/tlvc/arch/i86/drivers/block/directhd.c
+++ b/tlvc/arch/i86/drivers/block/directhd.c
@@ -666,9 +666,10 @@ int directhd_open(struct inode *inode, struct file *filp)
 	return -ENXIO;
 
 #if defined(USE_LOCALBUF) || defined(CONFIG_FS_XMS_BUFFER)
-    if (!(localbuf = heap_alloc(BLOCK_SIZE, HEAP_TAG_DRVR))) {
+    if (!open_drv_count)
+	if (!(localbuf = heap_alloc(BLOCK_SIZE, HEAP_TAG_DRVR))) {
 	printk("athd: cannot allocate bounce buffer\n");
-            return -EBUSY;
+            return -ENOMEM;
     }
 #endif
     if (!S_ISCHR(inode->i_mode)) { 	/* Don't count raw opens */
--- a/tlvc/arch/i86/drivers/block/directhd.c
+++ b/tlvc/arch/i86/drivers/block/directhd.c
@@ -50,13 +50,12 @@
 #include <linuxmt/string.h>
 #include <linuxmt/mm.h>
 #include <linuxmt/heap.h>
-#include <linuxmt/directhd.h>
 #include <linuxmt/debug.h>
 #include <linuxmt/errno.h>
 #include <linuxmt/config.h>
 #include <linuxmt/stat.h>	/* for S_ISCHR() */
 
-#include <arch/hdreg.h>
+#include <arch/directhd.h>
 #include <arch/ports.h>
 #include <arch/io.h>
 #include <arch/segment.h>
@@ -86,7 +85,7 @@ __asm__("cld;rep;outsw"::"d" (port),"S" (buf),"c" (nr))
 //#define USE_LOCALBUF		/* For debugging: use local bounce buffer instead of */
 				/* direct buffer io via far pointers.
 				 * When XMS buffers are active, the same 1k buffer
-				 * is used for bouncing. Price: 1k bytes */
+				 * is used for bouncing. Price: 1k bytes from heap */
 #define MAJOR_NR ATHD_MAJOR
 #define MINOR_SHIFT	5
 #define ATDISK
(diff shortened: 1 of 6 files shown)


tlvc/arch/i86/drivers/block/directhd.c
```diff
@@ -680,14 +665,25 @@ int directhd_open(struct inode *inode, struct file *filp)
     if (((int) hd[minor].start_sect) == -1)	/* FIXME is this initialized */
 	return -ENXIO;
 
-    if (!S_ISCHR(inode->i_mode)) 	/* Don't count raw opens */
+#if defined(USE_LOCALBUF) || defined(CONFIG_FS_XMS_BUFFER)
+    if (!(localbuf = heap_alloc(BLOCK_SIZE, HEAP_TAG_DRVR))) {
+	printk("athd: cannot allocate bounce buffer\n");
+            return -EBUSY;
+    }
```
ghaerr, line 672: Doesn't this alloc need to be happen only once when open_drv_count == 0? This seems to allocate a new bounce buffer into a single global each time the driver is opened - which occurs on separate raw opens and possibly other opens, correct?

No mem should return -ENOMEM, is -EBUSY for another reason?

A heap allocated bounce buffer has a (small but real) chance of major DMA overwrite bug if/when the kernel DS segment spans a 64k boundary, so this method is unsafe on early machines.

tlvc/arch/i86/drivers/block/idequery.c
```diff
@@ -74,7 +74,7 @@ int INITPROC get_ide_data(int drive, struct drive_infot *drive_info) {
 	word_t port = io_ports[drive >> 1];
 	int retval = 0, i, timeout = 10000;
 
-	word_t *ide_buffer = (word_t *)heap_alloc(512, 0);
+	word_t *ide_buffer = (word_t *)heap_alloc(512, HEAP_TAG_DRVR);
 
 	out_hd(drive, IDE_DRIVE_ID);
     	while (WAITING(port) && timeout--);
```
ghaerr, line 80: This code was changed in ELKS to a real time 30ms timeout which will work on very fast machines:
```
#define WAIT_READY      (3*HZ/100) /* 3/100 sec = 30msec - should be instantaneous */
        unsigned long timeout = jiffies + WAIT_READY; /* 30 ms */
        out_hd(drive, IDE_DRIVE_ID);
        while ((STATUS(port) & 0x80) == 0x80) {     /* wait 30ms until not busy */
        if (time_after(jiffies, timeout))
            goto out;               /* timeout */
        }
```

tlvc/arch/i86/drivers/block/directhd.c
Mellvik: Thanks, rushing things is never good. Appreciated. 
There is no DMA at work here, but the xd driver has borrowed some code from the floppy driver and needs some work in the dma dept.

tlvc/arch/i86/drivers/block/idequery.c
Mellvik: Thanks, I'll take advantage of that. Appreciated.

tlvc/arch/i86/drivers/block/directhd.c
ghaerr: Oh, I had thought you mentioned that your HD driver used DMA in a previous PR deleting some DMA stuff.

Mellvik: I did, although not this driver but the older (xt generation) mfm (xd) driver.
 Most ide drives and controllers can do dma but thanks to on-drive buffers, pio is faster on old hw.

tlvc/arch/i86/drivers/block/idequery.c
Mellvik: Actually @ghaerr, the `goto out` might as well be a `break`, right?

Also, further down in the ELKS code, I think the `if (drive == 2) {` is probably superfluous since the routine will be called again for the 4th drive.

Mellvik: @ghaerr, 
not part of a plan, but I happened to use the compaq286 with its original 40m wide drive to test the idequery changes. It turns out that 30ms wait doesn't cut it. Changed it to 300ms and we're fine. Since the loop will break as soon as the drive is ready, the time doesn't really matter, so a high number is fine.

A different question is whether the more accurate timing is meaningful in this case.

ghaerr: Wow, changing the timeout by a factor of 10 seems huge. How do you know the actual time accuracy when you're using `while(timeout--);` instead of jiffies?

> A different question is whether the more accurate timing is meaningful in this case.

I would say its always better to use a jiffies timer than otherwise, if for no other reason than to avoid problems with high speed emulators. Busy loops in general are bad practice for kernels, but we get away with it since we're non-reentrant, single CPU (at least for now!).

I'll take a look at the `if (drive == 2)`, thanks.

Mellvik: Well, in this case there is no need for any accuracy, it's just a matter of waiting for however long as it takes. Since the loop exits whenever the drive is ready, it is more like a watchdog timer than a specific wait. This being in the probe phase of system startup, it's not critical.

PR #90, opened 2024-10-09
[kernel,directhd] Add complete XT/IDE support to the directhd driver

Mellvik:
This major update to the `directhd` driver replaces the preliminary 'XT/CF' support added early this year with broad support for the XT/IDE class of devices. TLVC may now be compiled with XT/IDE support and work seamlessly across XT and AT class devices. XT/IDE support is activated via a new `bootopts` setting:
```
xtide=ioaddr, irq, flags
xtide=0x300,5,5        # IO address 0x300, IRQ 5, flags: no_interrupts, use_cflite
```
If the io-address parameter is missing, standard AT IDE is assumed.

The 'flags' field is extremely important as it is the only way to tell the driver what kind of XT/IDE card is installed. This is a messy area to say the least, described in detail in this new WIKI entry - 'The XTIDE catastrophe' : https://github.com/Mellvik/TLVC/wiki/TLVC-Developer-Notes#tlvc-ide-and-the-xtide-catastrophe

As discussed in the Wiki, there are 3 different 'types' of cards supported. The type is set in the boots flag like this (from `directhd.h`):
```
/* xtide configuration data & flags, for bootopts. Note that CFLITE and
 * HISPEED are mutually exclusive, setting both will create havoc */
#define XTIDE_CFG_CFLITE        1       /* Use the XT-CF Lite register addressing
                                         * scheme - A0 = 0, registers at even addresses only. Ctrl reg is base + 0x1c */
#define XTIDE_CFG_HISPEED       2       /* For cards supporting 'high speed PIO' which
                                         * means that address lines A0 and A3 have been swapped.
                                         * This places the two data registers (lo and hi byte) adjacent, 
                                         * and they may be accessed using word instructions.
                                         * If HISPEED is set, the address of the control
                                         * register block is base+16 instead of base + 8 */
#define XTIDE_CFG_NO_INT        4       /* Use PIO mode only, interrupts not supported */
                                        /* unused, set the irq value to zero instead */

```
If bits 0 and 1 of the flag are both 0, 'compatibility' mode is assumed.

Given the size (code and data) of the addition to support XT/IDE, the configuration option `CONFIG_IDE_XT` has been retained. Turning it off will save approximately 530 bytes of code and slightly above 100 bytes of data. It may be desirable to add the opposite CONFIG option too, making available a stripped down driver for XT class systems only. This idea is pending.

The updated driver has been extensively tested with 3 different cards/configurations. However, many variants and combinations have obviously not been check out - IOW bugs and subsequent updates are to be expected.

It has been tempting to add support for two controllers, while at it - a minimal effort in time and code. Adding this is particularly tempting from a development point of view. The second controller can (must) have BIOS code turned off and the configuration switches may be changed freely. This allows for more flexible testing. Pending.


Mellvik:
The final commit adds support for 2 controllers (expands the size of the `bootopts` `xtide` setting accordingly). 

Also finally tested the XT-IDE hi-speed mode - which didn't work for writing. Somehow, with word writes, the high byte ends up in the next word instead of the current. Apparently some kind of timing problem, need to take a close look at the UniversalBIOS sources to figure out how they handle this. Reading works fine, making hi-speed mode still worth it - reading is what we do most of the time anyway.

PR #91, opened 2024-10-13
[cmd] Update fdtest to use the directfd driver and raw IO

Mellvik:
The `fdtest` needed a long overdue update to work properly with TLVC and the `directfd` driver. Most of the old code (using direct BIOS calls) is now commented out: Access to raw IO allows a utility to do physical IO which is what this utility needs to do its testing.

Actually, the `dd` utility can do most of what `fdtest` does - with the help of the external `time` program. However, if a 'skip-length` - blocks to skip between each operation, is added, all kinds of new possibilities opens up for this program.

The timer needs to be fixed to report higher time resolution than 1 second.

### DIFF
--- a/tlvccmd/misc_utils/fdtest.c
+++ b/tlvccmd/misc_utils/fdtest.c
@@ -6,10 +6,15 @@
 #include <signal.h>
 #include <stdlib.h>
 #include <unistd.h>
+#include <fcntl.h>
+#include <sys/types.h>
+#include <sys/stat.h>
 #include <linuxmt/bioshd.h>
 #include <linuxmt/biosparm.h>
 #include <linuxmt/memory.h>
 
+#define USE_DIRECTFD
+#ifndef USE_DIRECTFD
 struct biosparms bdt;
 
 /* Useful defines for accessing the above structure. */
@@ -23,14 +28,16 @@ struct biosparms bdt;
 #define BD_DI bdt.di
 #define BD_ES bdt.es
 #define BD_FL bdt.fl
+#endif
 
 #define MAX		30		/* max # of sectors per read */
 #define SECSIZE		512		/* sector size*/
 
-unsigned char kernel_buf[MAX * SECSIZE];
+unsigned char iobuf[MAX * SECSIZE];
 int verbose = 0;
 int out = 0;
 
+#ifndef USE_DIRECTFD
 int read_disk(unsigned short drive, unsigned short cylinder, unsigned short head,
 	unsigned short sector, unsigned short count)
 {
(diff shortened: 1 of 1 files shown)


Mellvik:
About this update of ´fdtest´:
- Now uses jiffies for timing, temporary because even 10ms resolution isn't good enough for neither step rate nor average single sector read times on fast machines and 1.44M drives. An implementation used the precision timer is forthcoming.
- Uses the raw device - default /dev/rdf0, use `-d 1` to get to the second floppy drive.
- Without options, `fdtest` assumes 18 sector per track and runs a
    - Individual sector reads on one track (by default cylinder 3, head 0)
    - 1k block reads on the same track
    - Complete track read in one op
    - … all of these with timings.
- Use the -h, -s and -c options to specify a different starting point. `-s` is particularly useful when measuring head changes. Like, if sectors per track is 9, use `-t 9 -b 2 -s 8` to get timing for a read of the last sector on one side and the first on the other.
- Likewise, use `-t 9 -b 18` on a 9 sectors per track floppy to get timing for a complete cylinder read.
- The new `-S x` option may be invoked to determine 'real' track-to-track stepping time, where x is (preferably) the number of cylinders on the disk. Since there is no way to seek without reading or writing (which greatly affects timing), the higher number of cylinders, the more accurate the timing (even though the average sector read time is subtracted from the measured seek time before calculating the average step time). The accuracy and thus usefulness of these measurements will improve significantly with the introduction of the precision timer instead of jiffies.
- A 36 sector (18k) buffer is statically allocated and may cross 64k boundaries. This will affect the timings (split IO into several operations) and a warning is issued. To get details about memory allocation, use the `-v` option. Experience shows that when this happens, what it takes to get rid of it is just starting another program - like another login from a different terminal.
- The old code using direct BIOS calls has been commented out and not maintained during the code overhaul, and should probably be deleted.

There are many interesting observations to be made when using the utility on real hardware. For example, when reading blocks requiring a head change, we see that in most cases, the read is continuous, even on slow machines like the 8MHz V20. 
Even more interesting is average sector read times and how they differ depending on system speed and how they're read. The following loop when run on a single track - 1.44m drive, fast system, delivers 20ms access time average.
```
        lseek(fd, 0, SEEK_SET);
        for (i = 0; i < b; i++) {
                t = *jp;
                read(fd, iobuf, SECSIZE);
                a += *jp - t;
        }
        a /= b;         /* average */
```
With this loop, and otherwise identical parameters, the number more than doubles, 42-45ms.
```
        start = *jp;
        for (i=0; i < bs; i++) {
                j = read(fd, iobuf, SECSIZE);
                if (j != SECSIZE)
                        fprintf(stderr, "short read, expected %d, got %d\n", (int)SECSIZE, j);
        }
        end = *jp;
        j = (int)(end-start)*10;        /* turn into ms */
```
While on a slow (XT class) system and thus a lower data rate (half data speed), the difference between the two loops is minimal, typically 210ms in the first case, 240ms in the latter.
I'm thinking these observations are the beginning of understanding why the sector cache doesn't contribute positively on fast systems.

PR #93, opened 2024-10-19
[cmd] Prevent fdisk from creating partition beyond end of medium.

Mellvik:
Apparently since the beginning of time, `fdisk` has allowed creation of partitions 1 cylinder beyond the end of medium (by equating cylinder count with cylinder numbers). This minor update fixes that.

### DIFF
--- a/tlvccmd/disk_utils/fdisk.c
+++ b/tlvccmd/disk_utils/fdisk.c
@@ -166,8 +166,8 @@ void add_part(void)
     oset = MBR + PARTITION_START + ((part - 1) * 16);
 
     printf("Total cylinders: %d\n", geometry.cylinders);
-    for (scyl = geometry.cylinders + 1; scyl < 0 || scyl > geometry.cylinders;) {
-	printf("First cylinder (%d-%d): ", 0, geometry.cylinders);
+    for (scyl = geometry.cylinders; scyl < 0 || scyl > geometry.cylinders-1;) {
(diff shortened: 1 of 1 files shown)


ghaerr:
Wow, nice find! I'm surprised this has been in there so long, as it would seem many users might have selected max cylinders when creating a partition. Did this come to light by using an image that actually ran out of cylinders?

Mellvik:
What happens is that when initializing a minix fs in a partition that goes past physical, it tries to write (inodes presumably) beyond the end of the device and dies. So it has been an irritation (aka itch) for some time and I finally decided to get rid of it - having a blank 256M CF card available to play with..

Mellvik:
Completely off topic, @ghaerr: I finally got a complete from scratch copy of ELKS downloaded and cross tools compiled. It turns out the problem with Apple M(1-4) architecture and cross tools is still there (I had to fix that when I started with TLVC almost 2 years ago). Maybe you could get @tkchia to fix it:
https://review.haiku-os.org/c/buildtools/+/4329/2/gcc/gcc/config/aarch64/aarch64.h#1003

Mellvik:
Update - this turns out to be the right link for the fix: https://review.haiku-os.org/c/buildtools/+/4329

Mellvik:
Thanks for the heads up @ghaerr. I've been fortunate not to encounter this one (I just recently started using hd images in qemu). I'll be sure to add the fix. (The list is becoming dangerously looong, looks like I really need a cleanup weekend).

PR #95, opened 2024-10-22
[kernel] directfd: Rewritten autoprobe, many cleanups and fixes

Mellvik:
NOTE: This PR will prevent the system from compiling properly until the next PR - debug-infrastructure update - has been merged.

Important changes in this PR:
- The MINOR_SHIFT has been eliminated, devices are now literally (minor) 0-3, while forced floppy formats are 4 and up, per device. Also, the forced format # has been corrected in Make.devices.
- The open routine has been simplified and corrected. In the process a new per-device struct `fdev_s` has been created to hold per device status variables that could otherwise create race conditions if one drive was opened or closed while another was doing IO. 
- Auto-probing has been simplified and corrected - using some ideas from @ghaerr's ELKS version of the driver. The device size as reported via the inode is now correctly set after autoprobing. 
- The cylinder span check has moved from `redo-fd-request` to `setupDMA`.
- XT class systems will no longer attempt to (meaninglessly) autoprobe.
- A new `define` `MAX_FLOPPIES` has been added to keep the static data structures small. Set to 2 by default.
- Some cleanups in variable use: `dev` and `drive` were used inconsistently through the code.


ghaerr:
I must say it is informative each of us (re)writing and updating our own versions of this complicated driver, especially given the different focus each of us have when testing it. And in a driver like this, writing (or rewriting) code helps learn what's actually going on.

After looking at your comments and code in this PR, it appears that the mess that is the mass of globals in both our drivers may now be at it's design breaking point when attempting to operate two floppy drives at once. Can you elucidate more on what you've found that's requiring `struct fdev_s` and such? I had previously thought that the globals mess would still work, since the driver will only handle each I/O request (even for separate drives) only sequentially - each request has to completely finish before the next one is started, right (or have you found an exception)?

If the above does not hold, then we're going to have to put most everything into drive-indexed variables, including `track`, `sector`, `probing`, etc. Or is the problem only relating to `open/close` code, where each routine could be reentered (as a result of an I/O sleep etc). I don't (yet) see anything that would cause a process sleep during open though, as the first probe I/O waits until an actual I/O request to start.

On another point, can you explain a bit what the XT vs AT hardware difference(s) is regarding checking sys_caps for the CMOS settings?

Thank you!

Mellvik:
> I must say it is informative each of us (re)writing and updating our own versions of this complicated driver, especially given the different focus each of us have when testing it. And in a driver like this, writing (or rewriting) code helps learn what's actually going on.

Touché! and the floppy driver is either the worst or best of examples in that regard.  So much asynchonicity and interrupt handlers and jumping back and forth. It's indeed an interesting piece of work and like you point out, there is always more to learn...

> 
> After looking at your comments and code in this PR, it appears that the mess that is the mass of globals in both our drivers may now be at it's design breaking point when attempting to operate two floppy drives at once. Can you elucidate more on what you've found that's requiring `struct fdev_s` and such? I had previously thought that the globals mess would still work, since the driver will only handle each I/O request (even for separate drives) only sequentially - each request has to completely finish before the next one is started, right (or have you found an exception)?
> 
> If the above does not hold, then we're going to have to put most everything into drive-indexed variables, including `track`, `sector`, `probing`, etc. Or is the problem only relating to `open/close` code, where each routine could be reentered (as a result of an I/O sleep etc). I don't (yet) see anything that would cause a process sleep during open though, as the first probe I/O waits until an actual I/O request to start.

As I think the comment in the PR or the code mentions, the globals are safe except in open, release and ioctl. So that's what I fixed with the struct. The rest is protected by the busy test in early `redo_fd_request`. This about it, there is a 2minute fsck -vva going on on fd0 and the fd1 being opened ....

> 
> On another point, can you explain a bit what the XT vs AT hardware difference(s) is regarding checking sys_caps for the CMOS settings?

is this about config_types or in _open? If it's the latter, it just makes sure there is never a probe on XT class systems.

We spend too much time on floppies. I have noted that before. Interesting how addictive some challenges are! Let's continue to have fun!

ghaerr:
>  the globals are safe except in open, release and ioctl.

Ok, I see now how `open_inode` and possibly some others globals would be trashed (in my driver). I'll fix that.

> The rest is protected by the busy test in early redo_fd_request.

That "protection" is unneeded and has been removed in my driver - the ll_rw_blk.c `add_request` ensures that there will never be two concurrent I/O requests at the same time, instead queuing any new requests to the directfd request queue, and only removed on completion of the current I/O in progress.

For POC of the above, in your driver there is a "df: request while fdc busy\n": have you ever seen it? No. 

I will prepare some shell scripts for (mutual?) testing of overlapping I/O and open requests. Its definitely time to start thinking about putting together some regression testing of some sort, given all we've been through.

> is this about config_types or in _open? If it's the latter, it just makes sure there is never a probe on XT class systems.

I was wondering about both. I assume you're saying that there's no need to probe on XT systems because only a single floppy type can be installed? And that installation in turn won't be reported in CMOS? I thought that XT systems had CMOS?

Can you help me a bit by clarifying a bit more what the hardware is for XT systems (CMOS and floppies)? I'm still not completely there on the need for the sys_caps possibly as well as the xtflpy= /bootops requirement. Thank you!

> We spend too much time on floppies.

That's for sure. I have recently changed my mind in regards to the DF driver and now think that, regardless of its size, it needs to move to be the default driver in ELKS, if for no other reason than to make async I/O become standard, thus elevating ELKS (and TLVC of course) to being truly capable multitasking operating systems in their own right. 
And why not?! :)

Mellvik:
> > The rest is protected by the busy test in early redo_fd_request.
> 
> That "protection" is unneeded and has been removed in my driver - the ll_rw_blk.c `add_request` ensures that there will never be two concurrent I/O requests at the same time, instead queuing any new requests to the directfd request queue, and only removed on completion of the current I/O in progress.
> 
> For POC of the above, in your driver there is a "df: request while fdc busy\n": have you ever seen it? No.

That's right, I have't seen it. I have kept it, pending some more research into the possibility that a driver internal call to redo_fd_request (like from rw_interrupt after a request_done(1)) may collide with a call coming down from ll_rw_blk.

> 
> I will prepare some shell scripts for (mutual?) testing of overlapping I/O and open requests. Its definitely time to start thinking about putting together some regression testing of some sort, given all we've been through.

Great - that's a good idea. `fdtest` may possiby be useful in that context.

> 
> > is this about config_types or in _open? If it's the latter, it just makes sure there is never a probe on XT class systems.
> 
> I was wondering about both. I assume you're saying that there's no need to probe on XT systems because only a single floppy type can be installed? And that installation in turn won't be reported in CMOS? I thought that XT systems had CMOS?

OK, here's the deal
- XT has no clock, no cmos and can handle only one floppy density (the latter not strictly correct since 320k floppies may be supported, but in practice it's true).
- `xtflpy` replaces CMOS, if there is no `xtflpy`, 2 x 360k drives are configured regardless of what the hardware may look like. `xtflpy` enables the use of 720k drives.
- Only one density supported -> probing is a waste. The drives may be 720k or 360k but the format is the same. We can test for tracks, but that's an entirely different type of probe than the one we have - as discussed a couple of weeks back.
- With the XT class out of the picture probing wise, it's easy to reintroduce the DIR (media change bit) if desired (xt systems don't have that). For probing though I think what we have is good enough (and simple). It could be made even simpler if we didn't have to take 2880k drives into account, but it makes sense to keep that.

> 
> > We spend too much time on floppies.
> 
> That's for sure. I have recently changed my mind in regards to the DF driver and now think that, regardless of its size, it needs to move to be the default driver in ELKS, if for no other reason than to make async I/O become standard, thus elevating ELKS (and TLVC of course) to being truly capable multitasking operating systems in their own right. And why not?! :)

Indeed - that's good news. Great choice. !!

ghaerr:
Thanks for your explanation on the XT, CMOS and probing - that helps. I'll add the appropriate changes to the ELKS DF driver.

> the possibility that a driver internal call to redo_fd_request (like from rw_interrupt after a request_done(1)) may collide with a call coming down from ll_rw_blk.

Yes, it would be good for you to fully understand how this works. I ended up removing all this cruft after rewriting the SSD driver (see ssd.c for reference) as a basic async I/O driver to fully test the block I/O async subsystem. All async drivers must following the following simple rules: 1) only accept new requests through `add_request`, which protects the request queue from interruption via clr_irq/set_irq, 2) the driver must call end_request() to (re)set CURRENT to the possible next I/O request, and 3) the driver request entry point (xxx_request()) must be coded as a for(;;) loop and continue processing the CURRENT request (both when called when the queue is empty by add_request and when the next request is dequeued by end_request) until there are no more requests. The CHECK_REQUEST macro double-checks the queue integrity and panics when garbage is found. This is turned off for production use since I've never seen one.

Mellvik:
Yes, @ghaerr - this is async 101. This is how all async  drivers in TLVC work - network drivers, this driver, the xd driver.

What we're discussing is the possibility of a race condition. And now that we're looking at it, I recall that that risk was carefully considered when the driver was turned into a working piece of code for TLVC. Looking at it again I find that a race condition is unlikely - as experience proves - but not impossible, so I'm keeping it.

Whether the loop in the request function is a `for(;;)` or something else is a code design issue, not a functional one. The 'founding fathers' of Linux apparently liked `goto`, my preference is `while(1)`, I've left this one alone this far.

ghaerr:
> Whether the loop in the request function is a for(;;) or something else 

Sorry for the confusion! My comment was nothing to do with for(;;) vs while (1). I was trying to say that the async driver design requires a for/while loop to continue to empty the queue, since without that (e.g. by executing `return` before the queue is empty), will cause queue processing to necessarily stall.

> This is not an async driver - and turning it into one is quite a bit down on the list

(from previous comment on directhd.c): I see. I thought I saw a `USE_INTERRUPTS` and mistakenly thought that when set the HD driver became interrupt-driven. (I had to comment on the first PR that originally changed the `continue` to `return` because Github seemingly disallows comments on unchanged code?)

It is interesting your note that PIO is as fast or faster than interrupt driven (and can I assume using DMA in the interrupt case, or just using the interrupt signal completion then use PIO to read the HD data?)

But isn't having even a I/O-compleition-interrupt driver superior to a non-async (PIO-only) driver? With an async driver, other processes continue to run while the HD disk spins waiting for the sector to come under the head. That's an advantage, regardless of whether PIO is later used to read the data, right? Or are there other considerations here that I'm missing?

At some point when you think directhd.c is ready for export, I would like to use it in ELKS :) This may allow the conversion, as you've done with TLVC, to an async-only platform and the retirement of BIOS. Pretty cool really, even if the HD driver isn't fully async, as there are still currently some big issues with mixing BIOS and direct drivers.

> this is async 101. 
> What we're discussing is the possibility of a race condition.

Don't get me wrong - I don't care whether the check is left in or not - I'm commenting only because of the mutual desire for education on how stuff really works, (which admittedly can be quite tricky). There is no race condition here, because the OS is non-reentrant, single CPU and I/O requests are *only* inserted with `add_request`. Separate processes still bottleneck through `add_request` for all block I/O (not for opens though, of course). Since interrupts of the same IRQ level are controlled from kernel reentry through the PIC, there is no possibility of the I/O request handler being reentered, and no race condition.

I'm a big believer in removing cruft, and have been doing so for the past four years in ELKS, with the intent to produce understandability and probably more importantly simplicity through removing code (crazy, construction through deletion, huh?!). My take for the reason this fdc_busy/sleepon mutex was added in the first place is that our driver code comes from around the pre-Linux 2.0 days, when multi-CPU support was being added to Linux. In a concurrent-CPU kernel design, the kernel is reentrant, and any I/O request queue could be in fact scheduled to be handled by multiple CPU's at once. This removes the single-threaded add_request protection commented on above, and thus a multi-CPU mutex was added to the driver interrupt handler to protect multiple CPUs from talking to the FDC simultaneously. This mutex looks exactly like that kind of thing used in other parts of Linux 2.0. At this point, ELKS is very unlikely to support multiple CPUs, largely because of the huge changes imposed on the kernel.

My $0.02!!

Mellvik:
> > Whether the loop in the request function is a for(;;) or something else
> 

> (from previous comment on directhd.c): I see. I thought I saw a `USE_INTERRUPTS` and mistakenly thought that when set the HD driver became interrupt-driven. (I had to comment on the first PR that originally changed the `continue` to `return` because Github seemingly disallows comments on unchanged code?)

Well, the USE_INTERRUPTS wasn't intended for 'public' consumption, it was and still is an early test that the interrupts work on this device and when they kick. They do nothing except print a char on the console.

> 
> It is interesting your note that PIO is as fast or faster than interrupt driven (and can I assume using DMA in the interrupt case, or just using the interrupt signal completion then use PIO to read the HD data?)

While DMA is possible with (many maybe most) IDE systems, it's generally discouraged except on XT class PCs for the simple reason that the data is coming from/going to a buffer on the drive (it's memory-to-memory via the ISA bus), and the processor does that faster than DMA if the bus is 16 bit and the processor - I'm not sure where the borderline is here - is at least an 286@12MHz. The line where traditional ISA DMA becomes slower is in this area.

> 
> But isn't having even a I/O-compleition-interrupt driver superior to a non-async (PIO-only) driver? With an async driver, other processes continue to run while the HD disk spins waiting for the sector to come under the head. That's an advantage, regardless of whether PIO is later used to read the data, right? Or are there other considerations here that I'm missing?

That depends. If the drive is old enough, access times probably makes it worth it. But there is overhead added with interrupt handling, and the drive came with fast increasing cache sizes, so it's a complicated equation. I intend to pursue it though - if just for the heck of it, it's just not high up the list at the moment. The distinct advantage is the added asynchronicity, which may be beneficial even if it comes at a price.

The XD (MFM drives) driver is fully DMA/IRQ based and runs great. Like floppies there is no way to run these drives w/o DMA and thus interrupts on the XT machines the 'belong' to. It's interesting to look at the REQ queue when running such drives and floppies concurrently. Unfortunately my Compaq Portable Plus' Power supply conked again, so I haven't been using it for a while. There is a (well known) 64k DMA boundary problem in the driver that I haven't fixed pending getting the PS back amongst the living.

> 
> At some point when you think directhd.c is ready for export, I would like to use it in ELKS :) This may allow the conversion, as you've done with TLVC, to an async-only platform and the retirement of BIOS. Pretty cool really, even if the HD driver isn't fully async, as there are still currently some big issues with mixing BIOS and direct drivers.

Well, that interest may yank the directhd driver development up above the LANCE driver :-)

> 
> > this is async 101.
> > What we're discussing is the possibility of a race condition.
> 
> Don't get me wrong - I don't care whether the check is left in or not - I'm commenting only because of the mutual desire for education on how stuff really works, (which admittedly can be quite tricky). There is no race condition here, because the OS is non-reentrant, single CPU and I/O requests are _only_ inserted with `add_request`. Separate processes still bottleneck through `add_request` for all block I/O (not for opens though, of course). Since interrupts of the same IRQ level are controlled from kernel reentry through the PIC, there is no possibility of the I/O request handler being reentered, and no race condition.

When I looked at it again this morning, I still could not prove that impossibility to myself. The critical path in the driver is `rw_interrupt` calling `request_done(1)` which in turn calls `end_request()` which before returning will advance the CURRENT request pointer. Then it calls `redo_fd_request()` which checks if that pointer is zero. If it was zero by the time `end_request` returned, can `add_request` have run in between and changed it so that `redo_fd_request()` continues while `fd_request()`is called? That depends on what the `wake_up()` call in between triggers. That to me makes the next step unpredictable. If the req queue is empty, which is a requirement for this situation to be 'interesting' (dangerous), there is no chance that someone is waiting for a req and the `wake_up()`is a null call? Except all req's may be taken by some other device(s). The distance (code wise) in the driver, from the return from `request_done()` to the test for CURRENT != 0 in `redo_fd_request()` is really small, but may it still be interrupted?

I write down these questions as I run through cases like this and unless or until I can give the right answer with confidence, the protection stays. Getting rid of cruft is always high on the list, and I think we've done quite a bit of that in this driver (and many other places). But there have also been cases where my cleanups have cost me dearly - causing caution...

> 
> I'm a big believer in removing cruft, and have been doing so for the past four years in ELKS, with the intent to produce understandability and probably more importantly simplicity through removing code (crazy, construction through deletion, huh?!). 

Spot on, I've been a firm believer - for decennia - in the saying that a good programmer can be gauged not by codelines produced but by codelines deleted. 

> My take for the reason this fdc_busy/sleepon mutex was added in the first place is that our driver code comes from around the pre-Linux 2.0 days, when multi-CPU support was being added to Linux. In a concurrent-CPU kernel design, the kernel is reentrant, and any I/O request queue could be in fact scheduled to be handled by multiple CPU's at once. 

Quite possible. At times I'm looking at Linux v1 code for references, and the 'protection' is there - comments match our version up until 1993/4/29 - and no mention of SMP, but that may not mean much. One of these days the `get_request/add_request` code may need to be streamlined because it becomes an increasingly important part of our systems - and the question/issue will disappear.

> 
> My $0.02!!

...And now we're up to $0.04 !!

ghaerr:
Just for kicks, I looked back at the original directfd.c driver code from 20 years ago. It appears that the fdc_busy/sleep_on hack was there specifically for guaranteeing that the driver queue was emptied before processing a format floppy request. Since the format floppy request came in through a user-mode `format` program via `ioctl`, the ioctl code slept the `format` process until the I/O queue was empty, then set fdc_busy, then proceeded with the format. After the floppy format completed, fdc_busy was set to 0. In this case, the I/O request handler would receive requests via an additional function outside of add_request, and used a global mutex to keep the I/O requests separate.

All the formatting code has been deleted in the ELKS version of the driver.

My ~$0.02!!~$0.05, a nickel's worth!

ghaerr:
> While DMA is possible with (many maybe most) IDE systems,
> The XD (MFM drives) driver is fully DMA/IRQ based and runs great.

I'm glad I asked - I hadn't realized that you have seperate XD and IDE drivers, and somehow thought they were the same. Do they both live in directhd.c, or is the XD driver somewhere else?

> Well, that interest may yank the directhd driver development up above the LANCE driver :-)

I'm probably more interested in IDE support than MFM, due to availability of working hardware. Unfortunately, the user base for ELKS and TLVC seems to be quite very small, so I wonder sometimes what really matters... After I understand more about TLVC HD vs XD driver(s), I can make a more informed decision. IMO, having both interrupt driven would be worth it to remove all cases of synchronous I/O behavior (and thinking) within the OS(es).

>  If it was zero by the time end_request returned, can add_request have run in between and changed it so that redo_fd_request() continues while fd_request()is called? That depends on what the wake_up() call in between triggers. 

If you want an explanation of exactly how that works - the gist of it is that wake_up doesn't actually reschedule anything at the moment - ever. It just sets a bit in a task structure and always immediately returns to the caller. Understanding this about wake_up is key to understanding the way scheduling actually works in ELKS. schedule() is explicitly disallowed to be called (and checked for that) by an interrupt handler and will never be called on interrupt handler return. The rescheduling allow/disallow magic is all in the wonderful _irqit routine, which guarantees no kernel reentrancy via the _gint_count variable. 

> The distance (code wise) in the driver, from the return from request_done() to the test for CURRENT != 0 in redo_fd_request() is really small, but may it still be interrupted?

No. This is because the interrupt handler C code itself is in fact controlled also by _irqit: the IRQ 6 (or whatever) actually calls _irqit which then determines exactly how to call the C redo_fd_request interrupt handler code, but then most importantly, whether to allow a context switch when returning from the C code (and interrupt). In the case of kernel code being interrupted by an interrupt handler, rescheduling is never allowed. Rescheduling *only* occurs after a user-mode syscall or during a hardware timer interrupt during user mode execution, never (automatically) otherwise. Otherwise, an explicit call from the kernel to schedule() is required. (Note that wake_up) never calls schedule, and can't since wake_up is callable from an interrupt handler).

So you can rest assured that the kernel is never reentered. Interrupt handlers, part of the kernel, are technically "reentering" the kernel, but not the main kernel routines. Only a certain subset of kernel functions are allowed to be called from an interrupt handler, and then the interrupted kernel code is returned to.

Hope all that makes sense.

> But there have also been cases where my cleanups have cost me dearly - causing caution...

Yeah, I get it. 

> One of these days the get_request/add_request code may need to be streamlined 

That code is already extremely fast - not sure there's any need to streamline it, and not sure what could be done differently?

Anyways, let me know what you think the readiness of the XD and/or HD driver code is, I'm looking forward to integrating that into ELKS, which will be a very nice addition, following your excellent work getting the DF driver working! :)

Mellvik:
Check out the xd PR for details.
https://github.com/Mellvik/TLVC/pull/56#issue-2220227249

keep in mind, this is xt-class only, mfm on AT and up is nonsensical because IDE is built in and orders of magnitude better.

Mellvik:
Admittedly, it's easy to get confused here: the xt-ide stuff (or mess) is in the IDE driver.

And as an apropos to our discussion, XT/CFlite cards don't even have an interrupt line (CF Cards can do interrupts, but many do not support dma). To add to the confusion, most (probably all) XT-IDE cards handle CF cards well, have interrupts and other features discussed in the developer notes and PR notes.

https://github.com/Mellvik/TLVC/pull/90#issue-2576414011

ghaerr:
> While DMA is possible with (many maybe most) IDE systems, it's generally discouraged except on XT class PCs

Thanks for the explanations on IDE and MFM interrupts and DMA. 

> But there is overhead added with interrupt handling

What I'm thinking for ELKS is that it would be really nice to have a floppy and HD driver as default, and both interrupt-driven, if for no other reason than having a fully working async kernel w/drivers, with nothing dependent on the BIOS. That would show off ELKS and TLVC as behaving in a truly Linux-like fashion, which has yet to be fully accomplished by ELKS.

Given your understanding and comments about MFM, I would likely consider adding that to ELKS later and separately, as there just isn't that much demand for it and the system can always be configured for BIOS access for those that really want or need MFM. As far as DMA vs PIO, keeping DMA only in the DF driver simplifies things (for ELKS, since it's not dragging in the overkill DMA "library"), and even if the IDE interrupt handler only busy loops reading via PIO, that still allows for async floppy operation (or at least should, depending on the IRQ configured for both). At the end of the day the system is fully asynchronous and the process of getting there would further flush out thoughts and theories regarding interrupts and I/O like we're discussing now.

What do you think? TLVC is almost there with the IDE driver it would seem :)

> it's easy to get confused here: the xt-ide stuff (or mess) is in the IDE driver.

Yes, I read your developer notes (very nice and quite descriptive and complete) and am just venturing in to the whole XT-IDE thing. (BTW, there have been a number of very strange bug reports regarding XTIDE on ELKS, most of which turned out to be issues with the CF card itself, but with operational errors that seem to have nothing to do with it).

> XT/CFlite cards don't even have an interrupt line (CF Cards can do interrupts, but many do not support dma). To add to the confusion, most (probably all) XT-IDE cards handle CF cards well, have interrupts 

Well, now I'm getting more confused - can an interrupt-driven XTIDE driver not function for certain ISA cards or CF cards?

Also, can XTIDE be separated from a HD IDE driver, or does that just not make sense?

ghaerr:
> and even if the IDE interrupt handler only busy loops reading via PIO, that still allows for async floppy operation (or at least should, depending on the IRQ configured for both)

Looking a bit further into this, it appears that the default HD IRQ is 5, and DF is 6. This means that, if the IDE driver has an interrupt handler for IRQ 5 that performs PIO, the FDC IRQ 6 DMA complete interrupt won't be handled until the PIO is fully completed. Thus, any floppy I/O won't get to complete and/or schedule the next operation until the HD I/O is finished, if the I/O is completely overlapping. I don't know if this is good or not, but wanted to bring it up for discussion.

The alternative, handling the HD I/O request purely in normal kernel mode would allow for the FDC IRQ 6 to interrupt PIO processing and complete/start another (slow) FDC operation. Not sure which is better, I suppose it depends on how many microseconds are actually required (on either 8088 vs 386+) for reading 512 bytes PIO vs the floppy I/O.

Should DMA be used/considered instead, does IDE and/or XTIDE use a separate DMA channel than the FDC DMA? Depending on this and the PIO/FDC issue above, perhaps DMA is a superior solution?

Having said all this, perhaps it doesn't really matter given the relative frequencies of HD vs FD I/O requests and/or copying from HD<->FD etc. But it would be a bummer if a lot of development time went into getting the HD driver interrupt-driven, only to find it actuality either it's or the FDC's I/O ends up being synchronous with respect to each other. 

Comments? Do you have any knowledge about how DOS (or even early Linux) worked with respect to this? I would guess that Linux would have been much more concerned with HD performance than floppy, as few Linux systems ran on floppy for very long. I don't know if the same can be said for ELKS and/or TLVC however!

Mellvik:
> Just for kicks, I looked back at the original directfd.c driver code from 20 years ago. It appears that the fdc_busy/sleep_on hack was there specifically for guaranteeing that the driver queue was emptied before processing a format floppy request. Since the format floppy request came in through a user-mode `format` program via `ioctl`, the ioctl code slept the `format` process until the I/O queue was empty, then set fdc_busy, then proceeded with the format. After the floppy format completed, fdc_busy was set to 0. In this case, the I/O request handler would receive requests via an additional function outside of add_request, and used a global mutex to keep the I/O requests separate.
> 
> All the formatting code has been deleted in the ELKS version of the driver.

Thanks!! This lifted the curtain completely AFAIK - makes complete sense. I recall we discussed keeping or ditching the formatting code in the driver a while back, and happily decided to go different ways. Adding formatting is still on my list, and armed with this added knowledge, I'll put the fdc_busy mutex into the FORMATTING conditional compile.

> 
> My ~$0.02!!~$0.05, a nickel's worth!

I'll add another $0.05 to that (I cannot find the 'cent'- character on my keyboard...)

Mellvik:
> 
> > Well, that interest may yank the directhd driver development up above the LANCE driver :-)
> 
> I'm probably more interested in IDE support than MFM, due to availability of working hardware. Unfortunately, the user base for ELKS and TLVC seems to be quite very small, so I wonder sometimes what really matters... After I understand more about TLVC HD vs XD driver(s), I can make a more informed decision. 

Agreed, you need a user with MFM hardware to qualify doing anything with the XD driver, but know that it's ready and able so to speak. It would require - as you point out - reviving the original DMA code, but now that I'm familiar with it, it's no big deal and the code is small.

> IMO, having both interrupt driven would be worth it to remove all cases of synchronous I/O behavior (and thinking) within the OS(es).

Again agreed. It's an interesting project regardless of what we (or a user) may end up using since we get to measure the effects of one versus the other in different settings.

> ... 
> So you can rest assured that the kernel is never reentered. Interrupt handlers, part of the kernel, are technically "reentering" the kernel, but not the main kernel routines. Only a certain subset of kernel functions are allowed to be called from an interrupt handler, and then the interrupted kernel code is returned to.
> 
> Hope all that makes sense.

It does, thanks - that was useful.


> > One of these days the get_request/add_request code may need to be streamlined
> 
> That code is already extremely fast - not sure there's any need to streamline it, and not sure what could be done differently?

What I'm anticipating is that when several drive types (and units) start competing for req headers, some mechanism may be required to handle contention - since they will have vastly different speed. Like, if floppy operations ate all req headers, the XD or IDE drives would be waiting, hat in hand, for 200m rotational delays --- etc. To be handled when we get there.

> 
> Anyways, let me know what you think the readiness of the XD and/or HD driver code is, I'm looking forward to integrating that into ELKS, which will be a very nice addition, following your excellent work getting the DF driver working! :)

Pending the DMA 64k boundary fix in xd.c, it ready. Directhd - that's work in progress. DF - well, thanks, that seems to be a forever-work-in-progress case, doesn't it?

BTW - since the LANCE ethernet card requires DMA, that's another reason to consider the revival of the native DMA code. It's the weirdest piece of hardware ...

Mellvik:
> > But there is overhead added with interrupt handling
> 
> What I'm thinking for ELKS is that it would be really nice to have a floppy and HD driver as default, and both interrupt-driven, if for no other reason than having a fully working async kernel w/drivers, with nothing dependent on the BIOS. That would show off ELKS and TLVC as behaving in a truly Linux-like fashion, which has yet to be fully accomplished by ELKS.

Of course the HD driver provides BIOS independence already, but I agree with the desirability of using interrupts for both, and as commented above, get a feel for the difference. It's kind of like the network drivers - when writing, the data is PIOd to the NIC buffer and an interrupt comes when it has been processed.

> 
> ... and even if the IDE interrupt handler only busy loops reading via PIO, that still allows for async floppy operation (or at least should, depending on the IRQ configured for both). At the end of the day the system is fully asynchronous and the process of getting there would further flush out thoughts and theories regarding interrupts and I/O like we're discussing now.

Agreed.

> 
> What do you think? TLVC is almost there with the IDE driver it would seem :)

Not quite - but I do have a skeleton interrupt driven variant stashed, but I do need to get back to and complete the ktcp enhancements before fixing it. The project got sidetracked by our work on the precision timers ...

> 
> > it's easy to get confused here: the xt-ide stuff (or mess) is in the IDE driver.
> 
> Yes, I read your developer notes (very nice and quite descriptive and complete) and am just venturing in to the whole XT-IDE thing. (BTW, there have been a number of very strange bug reports regarding XTIDE on ELKS, most of which turned out to be issues with the CF card itself, but with operational errors that seem to have nothing to do with it).

Yes, that's common and I've seen my share of it. Cheap cards are like cheap floppies, and some times the problem actually is the CF-to-IDE adapter (even my Sarle CP/M system works with THIS particular adapter but not THAT one - it cost me weeks a few years back). Then there is 8bit vs 16 bit modes: Some cards simply don't do 8 bit, other do 8bit fine, but save it (past power cycling, which they should not). So a card that has been used in 8bit mode will not work in a 16 bit setting until reprogrammed, which is very simple but many (most?) drivers don't do it.

> 
> > XT/CFlite cards don't even have an interrupt line (CF Cards can do interrupts, but many do not support dma). To add to the confusion, most (probably all) XT-IDE cards handle CF cards well, have interrupts
> 
> Well, now I'm getting more confused - can an interrupt-driven XTIDE driver not function for certain ISA cards or CF cards?

Confusion is warranted - I dived into doing the 'complete' XTIDE support expecting a pice of cake, not a pile of rotten tomatoes (exaggerating of course). What I'm saying is that while the current HD driver works fine with XT/CFlite cards, an interrupt driven version would not support these cards - unless we have a way to turn interrupts off. Which we'll probably have, let's face that one when it hits.

> 
> Also, can XTIDE be separated from a HD IDE driver, or does that just not make sense?

No, it doesn't make sense. However, XTIDE support is conditional (menuconfig), and I've contemplated making regular IDE support conditional too, to enable a (somewhat) smaller XTIDE only driver for XT class systems. Most of the code is common for both groups.

Mellvik:
> > and even if the IDE interrupt handler only busy loops reading via PIO, that still allows for async floppy operation (or at least should, depending on the IRQ configured for both)
> 
> Looking a bit further into this, it appears that the default HD IRQ is 5, and DF is 6. This means that, if the IDE driver has an interrupt handler for IRQ 5 that performs PIO, the FDC IRQ 6 DMA complete interrupt won't be handled until the PIO is fully completed. Thus, any floppy I/O won't get to complete and/or schedule the next operation until the HD I/O is finished, if the I/O is completely overlapping. I don't know if this is good or not, but wanted to bring it up for discussion.

I think this is OK if not ideal. A typical HD operation would be 2 sectors and the time it takes to complete the (effectively) mem-to-mem move via the ISA bus is very short compared to the floppy transactions (depending on the speed of the processor of course, but I suspect even on XT class it would be 'reasonable' - we'll find out).
> 
> The alternative, handling the HD I/O request purely in normal kernel mode would allow for the FDC IRQ 6 to interrupt PIO processing and complete/start another (slow) FDC operation. Not sure which is better, I suppose it depends on how many microseconds are actually required (on either 8088 vs 386+) for reading 512 bytes PIO vs the floppy I/O.
> 
> Should DMA be used/considered instead, does IDE and/or XTIDE use a separate DMA channel than the FDC DMA? Depending on this and the PIO/FDC issue above, perhaps DMA is a superior solution?

Here's a useful quote from os2museum.com http://www.os2museum.com/wp/the-danger-of-datasheets/
```
...
I hadn’t thought of that but it’s true that in the PC and XT, the 8237 ran faster (4.77 MHz) than in the PC/AT (4 MHz), which even further disadvantaged 8237 DMA transfers in the PC/AT. The maximum transfer rate they give for the 8237 in the PC/AT with 16-bit transfers is 1.6 MB/s, which is a lot less than a bus-mastering adapter can transfer on 8-MHz AT bus (4 MB/s and very likely 5.33 MB/s or more).

The bottom line is that on a PC/AT, peripherals that needed to transfer data quickly (hard disks, network cards) did not use 8237 DMA because it would be counterproductive. But peripherals with limited bandwidth, tight timing constraints, and minimal buffering (floppies, sound cards) benefited from 8237 DMA.
```

IOW - DMA is good on XT class machines running at 4.77MHz - and it's good on later/faster machines if the peripheral BW demands are low. As they are with floppies. Which means that DMA + IDE (and XTIDE) is good on XT class, but not anything faster (including my V20 system running at 8MHz). 16 bit wide PIO at 8MHz (AT class, say 286@12MHz) beats DMA by a wide margin.

Another interesting consideration is that DMA always competes with the CPU for bus cycles. This is noticeable on slow/old systems because there is minimal if any instruction & data cache for the CPU to lean on when 'disconnected' from the bus (can't get to instructions or data) and because the CPU gets only a few cycles in between the DMA cycles (depending on the speed/BW of the peripheral at hand). The faster the cpu, the less noticeable the DMA activity is, which means that floppy activity is hardly noticeable on a 386 while it may take 50% of the bus bandwidth or more on an XT. This must be part of the reason the floppy sector cache doesn't deliver much if anything at all on fast systems.

As to your question, yes, IDE w/DMA uses a different DMA channel than the FDC. Typically DMA3 - I don't remember whether that is programmable or hardwired.

> 
> Having said all this, perhaps it doesn't really matter given the relative frequencies of HD vs FD I/O requests and/or copying from HD<->FD etc. But it would be a bummer if a lot of development time went into getting the HD driver interrupt-driven, only to find it actuality either it's or the FDC's I/O ends up being synchronous with respect to each other.
> 
> Comments? Do you have any knowledge about how DOS (or even early Linux) worked with respect to this? I would guess that Linux would have been much more concerned with HD performance than floppy, as few Linux systems ran on floppy for very long. I don't know if the same can be said for ELKS and/or TLVC however!

It's easy to forget that Linux got started in the 386 age, 32 bit, virtual and all. Looking at the early code there are no traces of using BIOS for anything except booting. The early (v1) hd driver is interrupt driven, and one of two I'm leaning on when doing the `directhd` 'rewrite'. I do not consider this 'rewrite' a lot of work - and as stated earlier in this thread, I believe it's worthwhile for a number of reasons although it may not deliver benchmark style performance increases.

Mellvik:
>>> and even if the IDE interrupt handler only busy loops reading via PIO, that still allows for async floppy operation (or at least should, depending on the IRQ configured for both)

One more thing about this: when an IDE read or write operation has started - one or several sectors, there is no busy loop involved, and string operations may actually be used. What I'm saying is that there is no test for a ready bit or anything between each byte. When the transfer starts, both ends know by many bytes to read or write, and the rest goes at CPU or bus speed whichever is slowest. This way the PIO beats (old style isa) DMA hands down. [PCI bust mastering DMA is a different story of course... and not of our concern]

ghaerr:
> What I'm anticipating is that when several drive types (and units) start competing for req headers, some mechanism may be required to handle contention - since they will have vastly different speed. Like, if floppy operations ate all req headers, the XD or IDE drives would be waiting, hat in hand, for 200m rotational delays --- etc. 

I see. I haven't got a handle yet on how many request headers the system needs in general, even during normal operation. In my last trace update, I added code that showed the max inuse count of request headers, which during a `cp /bin/vi /foo` copy showed a quick increase to IIRC 7 or so headers, then stopped, even with more copying later. I'll have to trace through this because I'm not sure this changes much with the size of the application buffer, as all application read/writes bottleneck through the filesystem read/write code which then uses the buffer system and finally block I/O (request headers). So it depends a lot on the system buffers setting(s) as well and what's in them.

I would say in general that request headers are cheap (small at ~20-22 bytes/each) and it would be in the system interest to never sleep a task (under "normal" system operation) waiting for request header. In order to determine this, I think adding a printk in `__get_request_wait` before `schedule()` to show when max headers are in use. I'll do that and perform some more testing. The ELKS SSD driver can also operate in async I/O mode for testing to compete with DF to test this.

> since the LANCE ethernet card requires DMA, that's another reason to consider the revival of the native DMA code. 

Thanks for the reminder on DMA for that. ELKS can always bring back the DMA library, I looked at it and discarded it for size reasons when porting over your DF driver, but frankly as long as the channels are different (they must be), using a small snippet of code like that in the ELKS DF driver would still work and might be lots smaller. IIRC the original library concerned itself a bit too much about whether or not another driver might be using the same DMA channel, etc.

> 16 bit wide PIO at 8MHz (AT class, say 286@12MHz) beats DMA by a wide margin.
> This way the PIO beats (old style isa) DMA hands down.

Ok, so DMA is out for the HD driver, good points. As you mention, we'll just have to see whether any issues turn up with an IDE IRQ 5 blocking out a floppy complete interrupt, but no data will be lost, so worst case would be waiting just a bit longer for the floppy I/O interrupt handler to get scheduled. I had originally worried that the floppy IRQ 5 had to be handled in real time, which isn't the case since the DMA transfer is already complete. I suppose that's why the DMA chip was added in the first place, now that I think of it.

Mellvik:
> Thanks for the reminder on DMA for that. ELKS can always bring back the DMA library, I looked at it and discarded it for size reasons when porting over your DF driver, but frankly as long as the channels are different (they must be), using a small snippet of code like that in the ELKS DF driver would still work and might be lots smaller. IIRC the original library concerned itself a bit too much about whether or not another driver might be using the same DMA channel, etc.

there are complications that may require a common 'driver' like the dma library for dma control. The (probably) most important being the flipFlop used when accessing 16bit registers in the chip. I think this snippet describes it well:
..........
Many devices on a PC (e.g. ATA disk drives) use 8 bit IO Ports to receive 16 bit values. This is done using a flip-flop. The device expects the low byte first. As soon as it receives a byte, the flip-flop changes state and then the device expects the high byte. When the high byte is received, the flip-flop changes state again, and the device expects a new low byte. Usually, each 16 bit 'register' will have its own flip-flop, but the ISA DMA controller has a problem regarding this.

On a 8237 chip, there is only one flip-flop. And there are eight of the 16 bit registers. And there can be up to three device drivers all competing for the use of that one flip-flop simultaneously.

This creates two serious problems. One is "contention issues". The other is that it is difficult to be sure what state the flip-flop is currently in. The standard solution for dealing with the flip-flop state issue is to reset the flip-flop to "low byte" state every single time you want to use it, just so you can be certain it is in the proper state before sending bytes. There are only two solutions to "contention": either use a lock, or allow only one ISA DMA driver, so that contention is impossible.
https://wiki.osdev.org/ISA_DMA
..........

ghaerr:
Super information on the 8237, thank you. I was just reading about programming the 8237, but missed the part about the single flip-flop. However, you have (finally) convinced me to leave in the CLR/SET interrupts around my non-library DMA code, as this information now shows precisely how programming the chip can fail through multiple access, unless locked (by interrupt disable/enable in my case around the direct DMA chip programming).

Interestingly, however, the DMA library code doesn't do any explicit locking around the flip-flop, which is its job. Instead, the library concerns itself with using an `xchg` macro to "protect" access to dma_request in the first place, which is pretty much unnecessary. The `xchg` macro, which I noticed was missing and you supplied, has to use a non-atomic implementation for our 8086 systems, since the actual/original `xchg` was required as a GCC hack to emit an 386+ XCHG register-to-memory instruction which atomically provides a real lock - all unavailable and thus useless on our systems!

Thus - the DMA library still isn't really required nor actually protecting anything - only real locking around the outb instructions is doing that outside the library!

(As you can see, I'm still not a fan of the DMA lib! :)

Mellvik:
> 
> Interestingly, however, the DMA library code doesn't do any explicit locking around the flip-flop, which is its job. Instead, the library concerns itself with using an `xchg` macro to "protect" access to dma_request in the first place, which is pretty much unnecessary. The `xchg` macro, which I noticed was missing and you supplied, has to use a non-atomic implementation for our 8086 systems, since the actual/original `xchg` was required as a GCC hack to emit an 386+ XCHG register-to-memory instruction which atomically provides a real lock - all unavailable and thus useless on our systems!
> 
> Thus - the DMA library still isn't really required nor actually protecting anything - only real locking around the outb instructions is doing that outside the library!
> 
> (As you can see, I'm still not a fan of the DMA lib! :)

This is a great catch! Thank you - I need to figure out a smart way to fix that. I'm not all that fond of the current dma code either, and it does need some simplification in addition to this fix. What I do like is having all the DMA stuff in one place. With (soon to be) 3 drivers using it, it seems to make my life easier ...

Mellvik:
> > What I'm anticipating is that when several drive types (and units) start competing for req headers, some mechanism may be required to handle contention - since they will have vastly different speed. Like, if floppy operations ate all req headers, the XD or IDE drives would be waiting, hat in hand, for 200m rotational delays --- etc.
> 
> I see. I haven't got a handle yet on how many request headers the system needs in general, even during normal operation. In my last trace update, I added code that showed the max inuse count of request headers, which during a `cp /bin/vi /foo` copy showed a quick increase to IIRC 7 or so headers, then stopped, even with more copying later. I'll have to trace through this because I'm not sure this changes much with the size of the application buffer, as all application read/writes bottleneck through the filesystem read/write code which then uses the buffer system and finally block I/O (request headers). So it depends a lot on the system buffers setting(s) as well and what's in them.
> 
> I would say in general that request headers are cheap (small at ~20-22 bytes/each) and it would be in the system interest to never sleep a task (under "normal" system operation) waiting for request header. In order to determine this, I think adding a printk in `__get_request_wait` before `schedule()` to show when max headers are in use. I'll do that and perform some more testing. The ELKS SSD driver can also operate in async I/O mode for testing to compete with DF to test this.

I recall from the testing when the FD driver was rewritten (and the only device using req's), that writing to a floppy would take the available req's fast: System running out of available buffers, calling sync which in turn sends as many buffers as possible to the req queue. That is, IIRC, taking 2/3s of them. The final 1/3rd being reserved for reads, and reads would hardly ever sleep, but writes would while the slow device fought to catch up. This experience triggered the statement above that got us started on the issue.

I hope to take a new look at this shortly, preferably with the an XD drive active - with a trace of what's actually going on in various situations. There's a lot of inspiration (and knowledge of course) in traces. Agreed, req headers are cheap but I don't think we can avoid running out of them, certainly not when using floppies.

ghaerr:
>  that writing to a floppy would take the available req's fast: System running out of available buffers, calling sync which in turn sends as many buffers as possible to the req queue.

Yes - for WRITEs only. When writing, the application can be allowed to continue as the user buffer has been copied to a system buffer, without any I/O scheduled. When there are no more easily released system buffers, sync starts and I/O is scheduled, as you say. For READs, the application has to wait for the each I/O request to complete before continuing. So we need to keep track of which side of the copy operation is actually being observed...

> That is, IIRC, taking 2/3s of them. The final 1/3rd being reserved for reads, and reads would hardly ever sleep, but writes would while the slow device fought to catch up.

Ah yes, I remember that code now. That's a whole 'nuther issue, determing how many read vs write I/O requests might want to be allowed for best system tuning. I'll make a note of that and try to measure it somehow.

> req headers are cheap but I don't think we can avoid running out of them, certainly not when using floppies.

Agreed and good points - the question might be how many read vs write vs total we should have on a tuned system. More measurements needed! I also plan on using your PT (precision timer) test results to calibrate my QEMU IODELAY-based emulation delay so that I/we can more accurately emulate floppy I/O without having to use real hardware for some of these tunings we're now talking about.

ghaerr:
> This is a great catch! Thank you - I need to figure out a smart way to fix that

Actually, I think the following would work quite well for you: instead of having a dma.c "library" with excess code, checking, and missing IRQ disable/enable, I'm thinking that converting the entire "API" (minus the open/close routines) to just a set of defines in dma.h would be far better. That is, just define routines like:
```
#define disable_dma(chan)    outb(FLOPPY_DMA | (chan), DMA_INIT)`
```
etc. In this way there is no function call overhead, no dumb open/close routines, and interrupts can be explicitly disabled and enabled around the DMA programming sequence, for maximum visibility. This gives a lightweight wrapper around what is being used now in the ELKS driver, but would match the source in the TLVC driver (almost). The only complication might be issues when the channel is <= 3 but could be made to work with some fancier GCC macro handling which optimizes away constant if expressions.

Mellvik:
Very interesting, @ghaerr -thanks.

I'll take a close look at that!

PR #97, opened 2024-10-24
[lib] Added @ghaerr's precision timer user space library routines

Mellvik:
The import of the precision timer routines from @ghaerr is complete, and may be called equally from user space and the kernel. An upcoming version of `fdtest` will be the first user space utility to use this valuable tool. Thanks @ghaerr.


Mellvik:
@ghaerr, what are the formatting options if any for the %k entry in printf? I'm looking to get rid of the decimals when unneeded.

ghaerr:
As you probably saw, the %k conversion routine(s) got a bit complicated (and took much longer than I'd like to admit to get right).

There isn't currently an option for eliminating the decimal - I assume you're wanting to display something simpler, just the digits preceding the dot? If so, I'm thinking the following mod might work for you, using the "alt" formatting mechanism which would be `%#k` in this case. For kernel printk:
```
    if (Sign)
        kputchar('-');
    while (*p) {
        if (n-- == Decimal)  {                /* only for %k pticks */
            if (alt) break; // <--- change these two lines
            else kputchar('.');
        }
        kputchar(*p++);
    }
```
Try that out and see if that suits your needs. If so, I'll convert both the kernel and C lib version to a permanent fix. I'm seeing a couple of minor issues with "alt" that could use an update, unrelated to this request.

ghaerr:
Oops - I just noticed you're looking for printf, not printk. Let me take a quick look at that.

ghaerr:
For printf, try the following mod in libc/misc/ptostr.c:
```
    while (*p) {
        if (n == Decimal) {
            break; // <---- change these lines
            //*outbuf++ = '.';
        }
        --n;
        *outbuf++ = *p++;
    }
```
If that works for what you want, then ptostr and libc/stdio/vfprintf.c will have be changed to add "hash" (alt) to be passed to ptostr. I'll do that if you can verify this is what you want. Note that until vfprintf.c is modified, `%#k` won't likely work as `#` processing was removed due to no format using it - so fdtest.c would remain just using %k for the time being.

ghaerr:
Note: a problem with the above approach is that it strictly truncates the value displayed at the decimal point, instead of rounding (up at .5). Making rounding work will be a lot more work, and might better be done before passing to printf/printk, if desired.

Mellvik:
Thank you @ghaerr, wow that was fast. Appreciated.
Feedback in the morning.

Mellvik:
Thanks @ghaerr - I imported the updates from ELKS, but I get unexpected results - I may be using the %k wrong:
Does the %k need a 'l' prefix? as in '%lk'? And if yes, does that apply to the '#' case? Like '%#lk'? I assume you've run the rather rigorous printk test you created which has turned out incredibly useful before.

I'm trying to figure out whether the changes actually reveal a new rabbit hole on my system since the symptoms look like memory/stack corruption, or if it's simply a matter of wrong size function arguments somewhere. what's happening is that the printk changes (I'm testing using the calibrate_delay tests again) seem to affect the numbers coming back from `get_ptime()`.

Mellvik:
Update @ghaerr - 
I've got the kernel printk working (seemingly) correctly. For some reason the timings I get now are different from the ones I got before the printk change - the new numbers some 15% lower than the old. I need to look at that. 

More importantly right now though is that I'm seeing that getting rid of the decimals was a bad idea because of the autoscaling in %k. No decimals is fine when there are 2 or 3 digits before the decimal point. When there is only one, you definitely want the decimals.

For reference I'm including some console outputs from the calibrate_delay function, the first is before applying the changes, the latter is after. You can see wherein the # format specifier has been used. The numbers themselves aren't interesting in this case, just the difference between the two runs.
```
Delay calibration index: 30, skew: 0
temp=49; 
41.062us (41) diff -9, d1=50;
41.062us (41) diff -9 (18%);temp=549; 
460.062us (460) diff -40, d1=500;
460.062us (460) diff -40 (8%);temp=5555; 
4.655ms (4655) diff -345, d1=5000;
4.655ms (4655) diff -345 (6%);temp=55807; 
46.766ms (46766) diff -3234, d1=50000;
46.766ms (46766) diff -3234 (1%);
MINIX: inodes 21845 imap 3 zmap 8, first data 696
VFS: Mounted root 0x0503 (minix filesystem).

--------------------------------

Delay calibration index: 25, skew: 0
temp=42; 
35.196us (35) diff -15, d1=50;
35us (35) diff -15 (30%);temp=461; 
386.318us (386) diff -114, d1=500;
386us (386) diff -114 (22%);temp=4658; 
3.903ms (3903) diff -1097, d1=5000;
3ms (3903) diff -1097 (8%);temp=46758; 
39.183ms (39183) diff -10817, d1=50000;
39ms (39183) diff -10817 (0%);
MINIX: inodes 21845 imap 3 zmap 8, first data 696
VFS: Mounted root 0x0503 (minix filesystem).
```

ghaerr:
> Does the %k need a 'l' prefix? as in '%lk'? And if yes, does that apply to the '#' case? Like '%#lk'? 

Yes.

> I've got the kernel printk working (seemingly) correctly. 

What was the problem? Was it the %#lk or source code port-over issues?

> More importantly right now though is that I'm seeing that getting rid of the decimals was a bad idea because of the autoscaling in %k. No decimals is fine when there are 2 or 3 digits before the decimal point. When there is only one, you definitely want the decimals.

I was afraid of that - welcome to the hairy business of massaging numeric output. Solving the problem in a way that works well is akin to rewriting floating point `ecvt` routine (take a look at libc/misc/ecvt.c if you want in on that). 

There isn't an easy answer to messing with the decimal point in the output routine itself, if this idea doesn't work well I can revert the %#k change and we'll need to think harder about exactly what is required and likely use routine like ecvt etc instead. I had originally thought you wanted this change for displaying within a known range of ptick outputs, but I can see that might be hard to guarantee. You may want to wrap the ptick results in another procedure before passing to %#k but that gets complicated as well.

>  For some reason the timings I get now are different from the ones I got before the printk change - the new numbers some 15% lower than the old.

Is output with *only* the %#k change, or with the ldivmod changes also? Given your initial porting problem, I'm not certain the exact code has been ported over? From the above, it looks like you are not using the %#k, just getting lower numbers? Certainly printk itself was heavily rewritten with the ldivmod routines and is much faster. and of course not displaying the numbers after the decimal point would result in faster timings, but that's not apparently what's happening above.

> 3.903ms (3903)

What is this output exactly? Is that supposed to be %k vs %#k? If so, that is incorrect. I can't really comment since I can't see the code producing the numbers.

It sounds like there are too many problems with the %# change - perhaps revert it and continue with testing with the real numbers.

Mellvik:
> > Does the %k need a 'l' prefix? as in '%lk'? And if yes, does that apply to the '#' case? Like '%#lk'?
> 
> Yes.
> 
> > I've got the kernel printk working (seemingly) correctly.
> 
> What was the problem? Was it the %#lk or source code port-over issues?

Both actually. History now. :-)

> 
> > More importantly right now though is that I'm seeing that getting rid of the decimals was a bad idea because of the autoscaling in %k. No decimals is fine when there are 2 or 3 digits before the decimal point. When there is only one, you definitely want the decimals.
> 
> I was afraid of that - welcome to the hairy business of massaging numeric output. Solving the problem in a way that works well is akin to rewriting floating point `ecvt` routine (take a look at libc/misc/ecvt.c if you want in on that).
> 
> There isn't an easy answer to messing with the decimal point in the output routine itself, if this idea doesn't work well I can revert the %#k change and we'll need to think harder about exactly what is required and likely use routine like ecvt etc instead. I had originally thought you wanted this change for displaying within a known range of ptick outputs, but I can see that might be hard to guarantee. You may want to wrap the ptick results in another procedure before passing to %#k but that gets complicated as well.

This is indeed a can of worms. The value span in the `fdtest` program is big, from a couple of ms to several secs, and what I'm seeing is that while clipping off the decimals was a bad idea, the ability to set the total number of digits should work well with this autoscaling (which is a great idea to begin with). Like `3.428s, 55.30us, 234.5ms` or 3: `3.43s, 55.3us, 234ms` etc. If it's a (relatively) easy fix.
> 
> > For some reason the timings I get now are different from the ones I got before the printk change - the new numbers some 15% lower than the old.
> 
> Is output with _only_ the %#k change, or with the ldivmod changes also? Given your initial porting problem, I'm not certain the exact code has been ported over? From the above, it looks like you are not using the %#k, just getting lower numbers? Certainly printk itself was heavily rewritten with the ldivmod routines and is much faster. and of course not displaying the numbers after the decimal point would result in faster timings, but that's not apparently what's happening above.
> 
> > 3.903ms (3903)
> 
> What is this output exactly? Is that supposed to be %k vs %#k? If so, that is incorrect. I can't really comment since I can't see the code producing the numbers.

This is debug output, and not understandable w/o the code. The first is `%lk`, the second is the 'control value' - 'manually' calculated from pticks.
The thing with the 15% difference bothers me because it doesn't make sense, and it should not have anything to do with this at all. The 'delay calibration index' changes from 30 to 25 between the two runs, and the only difference is some small changes to `printk`. I need to understand this one before letting the issue rest - and yes, It may have to do with `divmod` although I don't see how yet. If `divmod` does that, I'm incredibly impressed!!

> 
> It sounds like there are too many problems with the %# change - perhaps revert it and continue with testing with the real numbers.

That may well be the right choice. I'll be back when I figure this one out. BTW I haven focused out what's wrong with the libc version of `printf yet, the numbers coming out are an order of magnitude too low.

While on the subject of timing and measurements, there are all kinds of small and large surprises coming out of studying numbers from `fdtest`. Like this one from the fast 386 machine:
```
tlvc17# ./fdtest -S 80
/dev/rdf0: Size 1440k, using 18 sectors per track, 36 per cyl, start @ 3/0/1 (lba 108)

18 sectors (1-by-1):	350.147ms (19.452ms/sector)
 9 1k blocks:		417.022ms (46.335ms/block)
18 sectors in 1 op:	354.308ms
Average sector read time (no seek): 12.427ms
Seek time (80 steps*2): 763.226ms, step rate 4.769ms
```
What interesting is that reading a track sector by sector is slightly faster than reading the track continuously. The difference is within the margin of error, so let's say they are the same. Still very interesting (and explains the odd results from the cache tests). It's completely repeatable by the way - almost to the microsecond. Another one that needs some pondering to be explained, When compared to the numbers from the poor XT, this looks like a SpaceX rocket. The data rate difference is 1:2, the real speed diff is an order of magnitude. More about this later.

Mellvik:
The verdict on the kernel speed issue: The changes in `kernel/printk.c` - minimal as they are - somehow cause a 10% system speed reduction as measured by the `calibrate_delay` routine - not from 30 to 25 as I measured before, but from 28 to 25.
@ghaerr - you'l recognize the diff:
```
➜  kernel git:(misc) ✗ diff printk.c printk.c.mystash 
22a23
>  *              %#k     pticks truncated at decimal point
157,158c158,159
<         if (alt == ',' && ++i == 3) {
<             *--p = ',';
---
> 	if ((alt == ',' || alt == '\'') && ++i == 3) {
> 	    *--p = alt;
174,175c175,179
<         if (n-- == Decimal)                 /* only for %k pticks */
<             kputchar('.');
---
> 	if (n-- == Decimal) {               /* only for %k pticks */
> 	    if (alt)
> 	        break;
> 	    kputchar('.');
> 	}
202c206
<             if (c == '#' || c == ',') {
---
>             if (c == '#' || c == ',' || c == '\'') {
➜  kernel git:(misc) ✗ 
```
Quite possibly too small a number to be significant given the number of elements in the equation, and not the least the fact that there are no printk's in the code that calculates the index, but my curiosity is triggered: What or where is the connection?  I'm digging a little more before just shelving it as unsolved and unimportant.

BTW - the libc version is now working fine too! So - not that things work as expected, it's good time to revert, right? :-)

ghaerr:
> you'l recognize the diff:

I can't see how these changes help the timings? Probably not worth deep diving as its not a bug. Disasm the output would tell us more...

>  the libc version is now working fine too! So - not that things work as expected, it's good time to revert, right? :-)

Are you saying that you will not be using the %#k at all, and that I should revert the change? If you're not going to use it, we might as well take it back out, even though a mechanism to remove decimal points "sounds" like a good idea, it only actually works within small known ranges of values.

The above change also included a slight mod with the alternate thousands grouping character, that would remain.

Mellvik:
Small typo with big consequences: 'now' tuned into 'not':

"the libc version is now working fine too! So - NOW that things work as expected, it's good time to revert, right? :-)"

Thank you @ghaerr, and yes - this was a bad idea for reasons already discussed. I maintain that the ability to control the # of digits would be really useful, but only if the added complexity is low.

as to the unexplainable timings, definitely not a bug - if there is something there it'll pop up again at some point.

ghaerr:
> It may have to do with divmod although I don't see how yet. If divmod does that, I'm incredibly impressed!!

I see - you don't know whether or not you're testing against pre- or post- divmod printk kernel change(s). I can assure you that the previous version of printk was so slow that there's definitely a HUGE improvement. The reason for this is that the old printk performed a slow (non-divmod) DIV and separate MOD for every possible digit in the output *domain*. This is because it performed the integer-to-ascii in reverse order, starting with the 1 billion digit, then the 100 million, etc etc instead of doing it the new way, were the calculations are performed in reverse only for the displayable digits, then the string is reversed when output. Hard to describe, you'd have to look at old printk to see what I'm talking about. In the old printk, it took 10 DIV + 10 MODs just to output a '0' - ridiculously slow.

> I maintain that the ability to control the # of digits would be really useful, but only if the added complexity is low.

It is not easy to control the # of digits, as both the field width.precision %N.Mk would have to be parsed and passed, then calculated similar to the way that floating point precision is actually calculated. I'll wait to hear more about your results before deciding whether to revert %#k. Although it still sounds like a nice idea to "have", just truncated digits doesn't work well in practice (BTW, this is why a max field length is ignored on integer output using %3d etc, as us humans don't like dropped information without us knowing about it)

Mellvik:
> It is not easy to control the # of digits, as both the field width.precision %N.Mk would have to be parsed and passed, then calculated similar to the way that floating point precision is actually calculated. I'll wait to hear more about your results before deciding whether to revert %#k. Although it still sounds like a nice idea to "have", just truncated digits doesn't work well in practice.

Probably not. I have the benefit of not knowing how this magic works, which enables me to think outside the box. What I'm seeing is a string of digits, length determining the suffix (s,ms,us), ie a base 10 us number really. Then display the specified #of digits from the left. No calculations after the #of us have been determined. 
So simple when reality can be ignored...

Mellvik:
@ghaerr - sort of unrelated, but still related - and this one is **very interesting indeed**: 

One thing leads to another and my continuously revised `fdtest` program - now using the prec_timer package - delivered unexplainable results - on the 386. Occasionally the average_sector_access_time routine would return impossible results, like 3ms average sector access time whereas the normal and very reasonable number would be in the 40-60ms range. So I spent all morning digging into the mystery - including recording and printing the get_ptime() numbers for each access before averaging the result (which also provided some insight into my very primitive random number generator for the sector numbers to access - but that's a different story).

Heavy use of `get_ptime()` and not the least `printk` and `%lk`.

Partial output looks like below - the call to get the average and subsequently the track step speed, is duplicated for good measure - the output is from the compaq 386dx, 1.44M drive.
The first of the two runs looks fine - as in 'normal'. When sector 0 (actually 1) is accessed twice in a row, the access time is one revolution as expected. Looking at the sector # accessed and th previous sector number we can pretty much guess the time. Average access Tim 39ms looks reasonable. Then comes the second run - take a look.

```
legend: 
sequence # | sector # | access time | access time in pticks
             (start@0)
...
0 (14): 152.231ms 181661
1 (7): 112.940ms 134774
2 (13): 65.213ms 77820
3 (3): 80.238ms 95750
4 (0): 156.429ms 186670
5 (0): 199.078ms 237564
6 (4): 43.522ms 51936
7 (5): 693.026us 827
8 (13): 76.947ms 91823
9 (15): 11.664ms 13920
10 (7): 112.069ms 133734
11 (13): 55.158ms 65822
12 (3): 80.291ms 95813
13 (12): 97.875ms 116796
14 (7): 144.690ms 172662
15 (1): 123.682ms 147593
Average sector read time (16 repetitions)(1805165 1b8b6d): 39ms (3098935297)
Seek time (80 steps*2): 597.518ms, step rate 3.734ms
0 (1): 744.144us 888              <--------------- WTF?
1 (0): 188.251ms 224644
2 (0): 199.071ms 237555
3 (0): 199.081ms 237567
4 (4): 33.425ms 39887
5 (1): 166.401ms 198570
6 (8): 76.094ms 90805
7 (10): 11.658ms 13912
8 (10): 199.076ms 237562
9 (14): 33.457ms 39925
10 (7): 112.936ms 134769
11 (5): 177.298ms 211573
12 (13): 86.941ms 103749
13 (11): 177.349ms 211634
14 (14): 22.522ms 26876
15 (3): 79.397ms 94747
Average sector read time (16 repetitions)(2104663 201d57): 393us (30736386)
Seek time (80 steps*2): 576.150ms, step rate 3.600ms
```
... now this is interesting (and I have plenty more examples, most from the 386SX): A sector read in microseconds - not physically possible. Now look at the sector number: 1 - which is actually 2 or the second sector on the track, we're counting from zero in the program. Just before the test starts, there is a seek+read of sector 0 to initialize, so we've just requested the next physical sector on the medium. And it comes in a few microseconds. Well, not just a few, but less than a millisecond. Impossible, right?

Right. And here is the reason we're seeing such odd numbers when testing floppy cache on 'faster' (386) systems: The south bridge is always caching the next sector!!!! These machines don't have physical 765 or 8272 chips, they're integrated into a southbridge controller that is fully compatible seen from the CPU but does its own optimizations. EUREKA.

I first discovered this on the 386SX - which is a modern day (almost, 1999) machine with anything you can think of integrated into the southbridge. But I had to check the Compaq, 13 years older, and not likely to have a southbridge, right? At least I seemed to remember seeing the 765 chip on the mainboard. Wrong. There is indeed a southbridge chip.

This is output from the 386SX reading a 1.2M drive:
```
0 (9): 84.765ms 101152
1 (2): 83.191ms 99274
2 (12): 105.421ms 125801
3 (11): 156.400ms 186636
4 (6): 104.170ms 124309
5 (5): 156.415ms 186653
6 (1): 114.748ms 136931
7 (8): 63.704ms 76020
8 (10): 21.063ms 25136
9 (2): 72.615ms 86653
10 (8): 53.145ms 63420
11 (6): 145.926ms 174137
12 (13): 63.827ms 76166
13 (3): 51.448ms 61394
14 (4): 503.638us 601     <<-----------------
15 (1): 135.250ms 161397
Average sector read time (16 repetitions)(1685680 19b8b0): 33ms (2609577985)
Seek time (80 steps*2): 481.215ms, step rate 3.007ms
```
Again, reading sequentially - this time 3, then 4 - and the data is obviously coming from a buffer.  So what they have in their southbridges is a read-ahead buffer, which completely explains why we get the same or even slightly better timing when reading individual sectors sequentially than a full track, making the floppy sector cache is practically useless. And my sector access time routine needs rework....

Mystery solved. I figured you'd like to know.

ghaerr:
Wow @Mellvik, what a finding! This is an amazing result of your first real measurement use of the precision timing routines. I'm getting the feeling that our eyes will continue to be opened as we're now able to look at the reality of what's happening (vs what we may assume is happening).

My first thought after reading through this was that this confirms the need for adding `fdcache=0` (or automatically handle) turning cache off for south bridge machines, and somehow come up with a way to test for that capability.

So the new theory is that the Southbridge is performing a sector read-ahead in certain circumstances, and gets an internal cache hit when the FDC is sent a READ command requesting the very next sector from the one previously asked for? (The requests are through the raw driver, I am also assuming that the DF cache is OFF for these tests, right?) But when the FDC is sent a READ request that does not match the internal cache, the cache is invalidated, and the internal FDC waits for the sector then fulfills the read, then likely starts another read-ahead, is that it?

The above theoretical description doesn't really answer why the Southbridge machines perform *worse* when DF track caching is on, other than too many sectors are being read; but that same situation of reading "too many" sectors is also present with non-southbridge machines and the cache helps in that case. So it seems we need to figure out why the DF caching needs to be turned off for best performance. 

If best FDC performance is as simple as sector-readahead, then I figure: why not just emulate the Soutbbridge cache behavior in the DF driver itself (or at least for other physical FDC-chip machines)? Instead of adding read-ahead at the buffer layer, add it directly to the DF driver. The way this could work would be that instead of scheduling a full-, semi- or fixed-cache- track read, schedule only the first block read (at the ll_rw_blk.c layer), then, and only if there are *no more requests* in the request queue, schedule an additional (set) of I/O by creating new I/O request headers on the spot.

The cool thing about this new implementation idea would be that instead of scheduling a large I/O that can't be interrupted because of the DMA involvement, effectively schedule the same I/O (or just a single read-ahead, etc) as separate I/O requests, possibly fast enough such that the redo_fd_request could actually start the I/O before the physical sector passes. 

Another slightly different idea would be to always schedule a (block) read-ahead in the first I/O, but create additional I/O requests at the ll_rw_blk layer for the remainder of the track. With async handing, this could be measured and tuned.

Really great finding, lots of ideas coming from this! :)

Mellvik:
@ghaerr , I'll be back with a more detailed response i  the morning, but here's an important point (the way I understand the workings so far): the 'cache' is the sector currently under the head, not the sector after the last one requested. There is a big difference.

Mellvik:
> So the new theory is that the Southbridge is performing a sector read-ahead in certain circumstances, and gets an internal cache hit when the FDC is sent a READ command requesting the very next sector from the one previously asked for? 

Not exactly. What seems to be happening is that the sector under the head is being read continuously. If requested while valid (ie before the next one starts physically to pass under the head) you get it. From memory, no delay. Interesting- I've never even thought about that possibility.

> 
> The above theoretical description doesn't really answer why the Southbridge machines perform _worse_ when DF track caching is on, other than too many sectors are being read; but that same situation of reading "too many" sectors is also present with non-southbridge machines and the cache helps in that case. So it seems we need to figure out why the DF caching needs to be turned off for best performance.

Actually It does explain it. A single sector read is now faster than a multi sector read, and any multisector operation becomes a delay instead of a speedup, apparently (and surprisingly) even if we're talking about reading only the sectors needed.

I'm still no completely sure about the details, but playing more with `fdtest` tomorrow or Monday will help.

ghaerr:
> (the way I understand the workings so far): the 'cache' is the sector currently under the head, not the sector after the last one requested. 

> What seems to be happening is that the sector under the head is being read continuously. If requested while valid (ie before the next one starts physically to pass under the head) you get it. 

Got it, this is close but not exactly what I was thinking. So the updated/restated new theory is that the southbridge systems apparently (after track and head settling times) are streaming floppy media data continuously into from the current track into one of two buffers: the "current" buffer, which is the one being streamed to at the moment (or will be very shortly), or the "last" buffer, which is the last "current" buffer after the sector under the head has been fully read. These two buffers alternate each time a new sector passes under the head;

(continuing) if an FDC read request is made for a sector and its in the "last" buffer, the dma transfer is started immediately and the DMA complete is issued when the "last" buffer copy is completed. All the while the "current" data is being streamed into the "current" buffer... or something like this. I would assume that when a seek or FDC read request is received that is not in either the currently being received buffer or last buffer, both are invalidated and a new seek or sector wait process is initiated, and the cycle repeats.

This "theory", whether fully correct or not, is similar to a 765 FDC in that the process of programming the FDC is identical and DMA transfers are always performed (whether directly from streaming head on real FDC or possibly additional from the last buffer on SB. What is different is that SB systems will always be keeping the last sector seen in its last buffer (unless there is something more complicated about exactly which sector is buffered).

> A single sector read is now faster than a multi sector read

You mean because the SB system signals DMA complete after a single sector, but then continues to read the next sector?
I would assume so.

> and any multisector operation becomes a delay instead of a speedup,

It would seem this would only be true if reading more than two contiguous sectors. If two contiguous sectors read specified to be read, then that multisector read should be a bit faster than two single sectors, since the FDC doesn't need to be setup twice.

I type all this out so that we are very clear and what we think is happening, and to possibly flush out other possible theories of operation. Now, lets look at this with timings involved - does the following match at all what you've been seeing: ?

A 1440k floppy is spinning at 300rpm = 5 rps = 200ms/rotation. With 18 sectors, this is ~11ms per sector. So approximately 1 hardware timer interrupt occurs during every sector being read (amazing how slow?!) and 20 timer interrupts per full rotation. IIRC we figured the average instruction elapsed time was something on the order of 2.5us (~3 pticks) thus we're talking about ~4000 instructions between each timer interrupt, and the same 4000 instructions able to be executed between each sector as it passes under the head.

Given the above double-buffered streaming theory of operation, it would seem then that on SB systems, the kernel and application have ~1 timer tick = ~4000 instructions with which to schedule I/O, program the FDC, start DMA transfer from the "last" buffer, possibly copy the buffer to/from XMS, possibly copy the buffer to the application, have the application process the data, and repeat, in order to run at max speed. Of course, since the kernel requires two sectors/block, we're really looking at a full cycle time of 20ms / 2 timer ticks / 8000 instructions, with an average latency of half that.

Obviously the CPU speed is a big deal in measuring the elapsed time for the above full cycle, perhaps we should think of ways of measuring that, or that may be exactly what you're already doing in `fdtest`.

If the above description is anywhere's near close to reality, then I can see how for SB systems, it is probably best *not* to request any extra I/O from the FDC (as in caching), as that'll likely reset the "last" buffer and possibly interrupt the "current" buffer, unless those buffers happen to be the same as requested. For non-SB systems, the question remains as to whether the current caching is superior to the idea above where a read-ahead block is added to the I/O request queue to start FDC streaming into a DF driver buffer rather than waiting, only if there are no other I/O requests already queued.

Overall, sounds quite complicated to get right... per-system tuning may be the best answer and have lots of options available for doing so, kind of like we have now.

Mellvik:
> > (the way I understand the workings so far): the 'cache' is the sector currently under the head, not the sector after the last one requested.
> 
> > What seems to be happening is that the sector under the head is being read continuously. If requested while valid (ie before the next one starts physically to pass under the head) you get it.
> 
> Got it, this is close but not exactly what I was thinking. So the updated/restated new theory is that the southbridge systems apparently (after track and head settling times) are streaming floppy media data continuously into from the current track into one of two buffers: the "current" buffer, which is the one being streamed to at the moment (or will be very shortly), or the "last" buffer, which is the last "current" buffer after the sector under the head has been fully read. These two buffers alternate each time a new sector passes under the head;

I'm admittedly still speculating - as I have not done any more testing since the report from yesterday. The spec sheets don't really explain what I'm seeing, only referring to '16byte data FIFOs'. From the experience  it seems that we indeed have  a FIFO, but much more than 16 bytes in length, more like 512b. Given the difference in speed between filling the FIFO (from the floppy, slow) and emptying it (to memory via DMA), it seems possible that the FIFO is being filled from the next sector while we're reading (emptying) from the previous.

Important detail though - reading stops after the next sector, unless requested. There is no continuous read, that would interfere with 'normal operation' - head unload and motor stop etc.

> This "theory", whether fully correct or not, is similar to a 765 FDC in that the process of programming the FDC is identical and DMA transfers are always performed (whether directly from streaming head on real FDC or possibly additional from the last buffer on SB. What is different is that SB systems will always be keeping the last sector fully passed under the head in its last buffer (unless there is something more complicated about exactly which sector is buffered).

Actually - and in order to maintain compatibility, this IS the 765 (or later, as a licensed 'chip', in my 386SX case it's a 82077, maybe even later, I forget). Whatever the south bridge chips is doing, it's in front of the FDC, on the 'drive side', which is completely possible - this is there the decoding circuitry (analog to digital, digital to analog, PLLs and so on) resides anyway. We can 'trick' with the datastream but only within the physical restraints of the drive and medium.

> 
> > A single sector read is now faster than a multi sector read
> 
> You mean because the SB system signals DMA complete after a single sector, but then continues to read the next sector before issuing DMA complete? I would assume so.

Almost. Again still speculation but triggered by what I see: Reading one sector, handling it, then returning to catch the next even though the beginning of it may have passed by the head, and so on allows for more concurrency, less wait. When doing multisector IO, we wait for the physical duration of the spin. Did that make sense? It did to me, to a certain degree....

> 
> A 1440k floppy is spinning at 300rpm = 5 rps = 200ms/rotation. With 18 sectors, this is ~11ms per sector. So approximately 1 hardware timer interrupt occurs during every sector being read (amazing how slow?!) and 20 timer interrupts per full rotation. IIRC we figured the average instruction elapsed time was something on the order of 2.5us (~3 pticks) thus we're talking about ~4000 instructions between each timer interrupt, and the same 4000 instructions able to be executed between each sector as it passes under the head.
> 
> Given the above double-buffered streaming theory of operation, it would seem then that on SB systems, the kernel and application have ~1 timer tick = ~4000 instructions with which to schedule I/O, program the FDC, start DMA transfer from the "last" buffer, possibly copy the buffer to/from XMS, possibly copy the buffer to the application, have the application process the data, and repeat, in order to run at max speed. Of course, since the kernel requires two sectors/block, we're really looking at a full cycle time of 20ms / 2 timer ticks / 8000 instructions, with an average latency of half that.
> 
> It would be interesting to calculate elapsed time for a 1024-byte REP MOVSB with regards to the above, since this likely happens a lot, and is probably considerably slower on 8088 systems.

Since we're talking about a FIFO, the scenario changes quite a bit. Also, remember this is 386 systems only, the 286AT systems (I forget which processor the 2.5us was based on) have the physical NEC765 (or similar) chip.  Also, I suspect that this is a property of the 82077 (maybe even the 82072 which I think the first 386 systems had) - which makes much more sense than it being something the individual vendor adds in their southbridge ASICs.

> 
> Obviously the CPU speed is a big deal in measuring the elapsed time for the above full cycle, perhaps we should think of ways of measuring the full cycle elapsed time, or that may be exactly what you're already doing in `fdtest`. I am a bit surprised to hear an average of 40ms/sector, when one would think the average would require a full half rotation delay = 100ms + 11ms for the sector read. Any thoughts on that?

The 40ms is the type of number we get because of the FIFO. In most cases (and with recent changes to `fdtest`) the access numbers are 'sane', circling around 100ms. Take a look at the timings in the examples above.

I think the smart thing to do is to look at the numbers, check them as much as possible and act accordingly. I'm surprised there isn't more documentation out there about this, but then again how difficult is a FIFO? so my take is to use this knowledge for what it's worth, detect CPU and turn off FDCACHE if 386. It may not be a 100% sure hit, but exceptions may be handled via bootopts anyway.

ghaerr:
> The spec sheets don't really explain what I'm seeing, only referring to '16byte data FIFOs'. 

Yes, we're probably both speculating quite a bit. What exactly southbridge chipset number are you using (or datasheet)? I'm looking at several and it seems the newer units dropped support for an FDC entirely, although some hide the FDC support in the "Super I/O controller". I'd like to find the data sheets and/or application notes and read up further on the hardware. I'm not understanding how a 16-byte FIFO has anything to do with regards to floppy I/O, and if the FDC is part of the Super I/O controller they may be referring to the legacy (non-USB) serial also implemented within the Super I/O portion.

> use this knowledge for what it's worth, detect CPU and turn off FDCACHE if 386.

Yes, since you've found for sure that all the 386+ systems you've tested run faster without cache, just turning off caching based on CPU sounds like a good starting point.

Currently, cputype.S, called from setup.S, only checks and sets arch_cpu up to <= 6, which is 80286 or earlier. I could probably pull out the 386+ test routine from `check_unreal_mode` in unreal.S (without turning on unreal mode) and use that for the 386 cache disabling if you'd like to test that on real hardware. I'm not sure I can fully test that on QEMU at CPU <= 286. I could add the 386+ check to cputype.S from unreal.S and have everything in arch_cpu for use in the floppy driver(s).

By chance, have you (ever) tested TLVC w/BIOS driver to see whether a disabled cache is faster on real hardware? That would seem to be a double confirmation that our theories with regards to caching is correct also outside the DF driver. I would like to know whether we need to disable track caching on BIOS as well. Sometime if you ever boot ELKS that should be easy to test.

ghaerr:
> I'm not understanding how a 16-byte FIFO has anything to do with regards to floppy I/O,

I researched more on FDC controllers, and although unable to find anything (yet) on builtin FDC cache, I was able to find [more information on the floppy controller documentation on Linux regarding the FDC FIFO](https://www.kernel.org/doc/Documentation/blockdev/floppy.txt). It seems that later versions of some FDCs did have a programmable FIFO and this is used for non-DMA transfers: the FIFO queue length can be set to increase or decrease the amount of time required between FDC interrupts to empty the FIFO when used in non-DMA transfers, thus not having to busy-loop the CPU but instead relying on a FIFO for intermittent reading. I haven't yet found any documentation on actually programming the FIFO but it seems a good place to start might be (re)looking at the Linux floppy driver.

ghaerr:
Signing off for now. I've searched pretty hard but have not been able to come up with any documentation regarding FDC sector caching, although there is certainly FDC FIFO documentation and all of that seems to pertain only to non-DMA transfers and reducing the number of interrupts required to complete a non-DMA transfer, none of which apply to our drivers.

> Given the difference in speed between filling the FIFO (from the floppy, slow) and emptying it (to memory via DMA), it seems possible that the FIFO is being filled from the next sector while we're reading (emptying) from the previous.

Yes, it does seem possible, but we are both speculating, so here's some more: are we absolutely certain that the behavior you're seeing is in fact coming from an FDC READ then DMA operation and not actually a quick copy in the DF drive using `xms_fmemcpyw`? I took to reading tlvc/fs/block_dev.c's `raw_blk_rw` and it appears to always set req->nr_sectors to the requested byte count (which could be larger than a sector) if the offset is sector aligned and full sectors are being read. Assuming `fdtest` always reads only 512 bytes, this looks correct. However, the DF driver, while seemingly never setting `use_cache` on raw driver requests can in fact set `use_bounce` if the application buffer is not 64K aligned, setup_DMA will request a 1K (2 sector) read and then in `rw_interrupt` the variable `use_bounce` overrides `use_cache` and uses xms_fmemcpyw regardless from DMASEG for the transfer. While possibly not likely, it seems this could result in a fast copy rather than a sector wait/read. I may be misinterpreting the code as I have not studied it in depth, as the raw driver exists in TLVC only. This would also answer why a contiguous sector "readahead" is very quickly performed.

I mention this since we're pretty much speculating on everything with little to no documentation on southbridge operation. I've scanned the 82077AA reference and there's no mention of this, which would certainly qualify as a major feature, given the other "features" being touted of lesser value, it would seem. A fast DF driver xms_fmemcpy could be ruled out with debugging turned on in the driver, it would seem.  Obviously another test would be getting the `fdtest` numbers on non-386, non-southbridge systems, to see how they differ. Other thoughts would be to modify `fdtest` so that contiguous setons are always read to see how this affects our theory on 386 systems as well.

Mellvik:
> > The spec sheets don't really explain what I'm seeing, only referring to '16byte data FIFOs'.
> 
> Yes, we're probably both speculating quite a bit. What exactly southbridge chipset number are you using (or datasheet)? I'm looking at several and it seems the newer units dropped support for an FDC entirely, although some hide the FDC support in the "Super I/O controller". I'd like to find the data sheets and/or application notes and read up further on the hardware.

I've looked at quite a few over the past couple of days, like the SMC 37C669, they all have licensed FDC chips inside, typically the 82077, and mention only the intel specs which we get from the Intel data sheet anyway. 16byte FIFO. 'Southbrige' is a generic term for all these 'smart' (integrated) IO controllers btw, sitting on the south side of the CPU to handle slow IO (as opposed to the Northside which is memory, graphics, ...). 

> I'm not understanding how a 16-byte FIFO has anything to do with regards to floppy I/O, and if the FDC is part of the Super I/O controller they may be referring to the legacy (non-USB) serial also implemented within the Super I/O portion.

Check out the 82077 (or 82072) data sheet for some details on this. The FIFO alleviates the pressure to get the data coming from the floppy drive to the memory bus in real time, sitting on the 'floppy side' of the IO stream. it should be easy to increase the size of the FIFO, unless it's completely integrated into of the licensed 'die', making changes hard or legally impossible.

> 
> > use this knowledge for what it's worth, detect CPU and turn off FDCACHE if 386.
> 
> Yes, since you've found for sure that all the 386+ systems you've tested run faster without cache, just turning off caching based on CPU sounds like a good starting point.
> 
> Currently, cputype.S, called from setup.S, only checks and sets arch_cpu up to <= 6, which is 80286 or earlier. I could probably pull out the 386+ test routine from `check_unreal_mode` in unreal.S (without turning on unreal mode) and use that for the 386 cache disabling if you'd like to test that on real hardware. I'm not sure I can fully test that on QEMU at CPU <= 286. I could add the 386+ check to cputype.S from unreal.S and have everything in arch_cpu for use in the floppy driver(s).

Sounds good! I may be wrong but I seem to remember there is a complete list of CPUs recognized in cputype.S, commented out. If not, I have the code for such detection function in a book somewhere lees than 3 feet from my screen.

> 
> By chance, have you (ever) tested TLVC w/BIOS driver to see whether a disabled cache is faster on real hardware? That would seem to be a double confirmation that our theories with regards to caching is correct also outside the DF driver.

I haven't tested that specifically, but I did update the BIOS IO stuff to make sure TLVC compiles if it's selected in menuconfig. IOW it's a quick test to do, and I agree - it's worth it for knowledge.

> I would like to know whether we need to disable track caching on BIOS as well. Sometime if you ever boot ELKS that should be easy to test.

I did get ELKS to build and compile on an Intel Mac about a week ago (even after getting the cross tools problems out of the way, there were ELKS compilation problems when running Apple M hw and I didn't dive into it). I'll see what I get around to ...

Mellvik:
> > I'm not understanding how a 16-byte FIFO has anything to do with regards to floppy I/O,
> 
> I researched more on FDC controllers, and although unable to find anything (yet) on builtin FDC cache, I was able to find [more information on the floppy controller documentation on Linux regarding the FDC FIFO](https://www.kernel.org/doc/Documentation/blockdev/floppy.txt). It seems that later versions of some FDCs did have a programmable FIFO and this is used for non-DMA transfers: the FIFO queue length can be set to increase or decrease the amount of time required between FDC interrupts to empty the FIFO when used in non-DMA transfers, thus not having to busy-loop the CPU but instead relying on a FIFO for intermittent reading. I haven't yet found any documentation on actually programming the FIFO but it seems a good place to start might be (re)looking at the Linux floppy driver.
> 
> Additional information from the IBM PS/2 Hardware Interface Technical Reference Manual describing the FIFO mode and other interesting variations of the FDC controller is available [here](https://www.ardent-tool.com/docs/pdf/ibm_hitrc06.pdf).

Actually, this is all about the intel 82072/82077 and the IBM text seems to be more or less cut and paste from the Intel data sheet, describing what I mentioned above about the FIFO alleviating DMA constraints (and PIO, I didn't read that far). Our driver is already supporting this, turning on the fifo (if the 82077 is detected), and - IIRC - setting it to max. 

BUT - and I just realized this - all my testing has been with the 82077 support disabled in the driver (I have CONFIG_HW_XT on - which turns off 82077 support - so I can use the same kernel across all machines). So whatever the contribution from the 16byte FIFO is, it's not involved here.

Even if that was not the case (ie. 82077 support was on), what I'm seeing cannot possibly be explained by a 16byte FIFO, so there is something more to be understood here. I would have thought my readings from `fdtest` somehow were corrupted if it weren't for the fact that they corroborate the observations from the fdcache testing.

I don't think it will change anything, but I'll turn on 82077 support just for the heck of it. If nothing else, it will show what FDC Compaq is using in their southbridge on the Portable 386.

Mellvik:
@ghaerr, 
for the record, yes - I've been through the driver and ll_rw_blk and bock_dev backwards and forwards with printk verification - they're all good.

Anyway, based on the previous two posts, I've continued the testing and speculations and I believe to have found a reasonable explanation.

As expected, activating  support for the 82077 in the driver didn't change anything - except tell us that the COMPAQ portable 386 has a 765/8272A inside its ASIC (the 398sx system has a 82077AA).

I put the verbose output back in `fdtest` and wanted for the heck of it to see what the numbers were for a 360k floppy in the 1.2M drive.  The numbers immediately looked  out fine. When I looked closer though, I found the same pattern: Too low numbers when reading a sector adjacent to the previous one. Not extremely low as for the higher speed drives, but lower than (rotation speed/# sectors per track). (Notice that sector numbers start at 0, in column 2, and since this is a 9 sector per track drive, numbers above 8 are head 1).
```
/dev/rdf1: Size 360k, 9 sectors/track, 40 cylinders, start @ 3/0/1 (lba 54)
18 sectors (1-by-1):	458.386ms (25.465ms/sector)
 9 1k blocks:		536.857ms (59.650ms/block)
18 sectors in 1 op:	447ms
0 (0): 156.832ms 187151
1 (12): 42.091ms 50228
2 (7): 69.504ms 82941
3 (5): 132.060ms 157590
4 (1): 87.569ms 104498
5 (4): 52.019ms 62076
6 (1): 104.867ms 125140
7 (4): 42.021ms 50145
8 (1): 114.884ms 137094
9 (8): 111.560ms 133127
10 (10): 35.333ms 42164
11 (14): 69.272ms 82664
12 (7): 34.896ms 41643
13 (13): 104.684ms 124922
14 (15): 34.766ms 41488
15 (3): 104.947ms 125236
Average sector read time (16 repetitions)(1548107 179f4b): 81ms (96756)
Seek time (40 steps*2): 839ms, step rate 10.491ms (rough estimate)
0 (0): 166.875ms 199135
1 (8): 138.979ms 165847
2 (14): 104.616ms 124841
3 (15): 17.460ms 20836  <---------------
4 (7): 17.447ms 20821 <------------
5 (5): 122.058ms 145654
6 (5): 166.911ms 199178
7 (9): 70.216ms 83791
8 (14): 76.630ms 91444
9 (7): 34.918ms 41669
10 (9): 45.381ms 54154
11 (14): 76.596ms 91404   (sec 5, h 1)
12 (15): 17.445ms 20818  <--------------- (sec 6/h 1)
13 (7): 17.449ms 20823   <---------------- (sec 7/h 0)
14 (5): 122.058ms 145655
15 (1): 97.577ms 116441
Average sector read time (16 repetitions)(1542511 17896f): 80ms (96406)
Seek time (40 steps*2): 829ms, step rate 10.374ms (rough estimate)
tlvc17# 
```
Average 80ms adds up beautifully (being a 1.2m single speed drive it runs at 360rpm), 166ms/revolution.
Back to the (apparently) too-fast-sectors - consistently 17.5ms. Lower than the physical 22.2ms per sector, but that's a 300 rpm based number. The 360rpm number is 18.5ms, and given that we're measuring the transfer of data only, which is approximately 75% of the sector, this looks really good. Which is what makes it interesting in this context: All the numbers add up, confirming the timer, the test and the tooling setup.

So what is it that cause the higher speed drives, running at 500kbps (as opposed to the 250kbps format)? 500kbps means 2us per bit and 8ms per sector just to get the data across. That doesn't match with the measurements when reading adjacent sectors, consistently reporting between 400 and 800 us. Let's check the math: we have 5555 physical bits per sector (1.44M drive, 100k bits per track), the 26.5% 'overhead' is the same as for the lower capacity drive. Sector rotation time is 200ms/18=11.1ms less 26.5% = 8.1ms. So if extremely lucky with the timing of a read, 8ms is the absolute minimum time to get 512 bytes across from the drive through the controller and into RAM.

IOW there is no way 400 or 800us can happen (and we've already eliminated the possibility that there is a buffer or HUGE FIFO somewhere in the chain). So I changed the fdtest` setup to read consecutive sectors individually instead of random and headed back to the 1.44M drive. Wow - almost all reads are in the  600-900us range, average sector access time 2ms! Reading the same track, individual sectors, but calling get_ptime only before and after the loop delivers entirely different results:
```
18 sectors (1-by-1):	367.977ms (20.443ms/sector)
```
We're an order of magnitude off! It seems to me we may have a bug in get_ptime when reading values below (ca.) 10ms - possibly an interval in the 1-10ms range. @ghaerr - could you look into that?

ghaerr:
> for the record, yes - I've been through the driver and ll_rw_blk and bock_dev backwards and forwards 

Thanks for checking.

> (and we've already eliminated the possibility that there is a buffer or HUGE FIFO somewhere in the chain)

I wasn't aware that you've eliminated the possibility of the southbridge/FDC performing readahead into its own internal buffer, but frankly I'm not leaning in that direction, as nothing talks about such a feature.

> IOW there is no way 400 or 800us can happen
> almost all reads are in the 600-900us range, average sector access time 2ms! 

A couple ways that could happen are 1) a bug in get_ptime certainly, and 2) fmemcpy of cache somehow. It might nonetheless be interesting to calculate or measure how long a memcpy takes for 512 bytes, just to see if that matches up. (No, I don't believe this is the issue, but just trying to get my measurement units in a row).

> the 26.5% 'overhead' is the same as for the lower capacity drive. Sector rotation time is 200ms/18=11.1ms less 26.5% = 8.1ms.

Interesting - I was just reading about the overhead issue in the IBM tech ref. This "overhead" really only should be added for individual sector reads, because when reading tracks or multiple sectors the disk rotation over the unused bits still matters... right? (Again, trying to get my ducks in a row regarding measurements). This does answer why I thought it impossible to get lower sector read numbers than mathematically possible though.

> It seems to me we may have a bug in get_ptime when reading values below (ca.) 10ms - possibly an interval in the 1-10ms range.

Well - the sub-10ms interval IS treated quite differently than higher intervals, because the 10ms interval is the hardware timer tick interval that increments jiffies, which then is accounted for slightly differently. We also have to consider that this test is our first test of get_ptime in user mode, where interrupt rescheduling is handled quite differently than in kernel mode.

Let me take a quick look, I might post some code and it would probably be good for you to follow along. I had thought I had tested this very thoroughly, but user mode could be causing issues.

ghaerr:
I've checked and double-checked get_ptime and am not seeing anything. That doesn't mean its not there, but if we need to test get_ptime with sub-10ms intervals, we need to think up another way to do it from user space. Possibly using plain memcpy with a known buffer size, calculated as I talked about above to ensure we're close to what we think it should be.

> So I changed the fdtest` setup to read consecutive sectors individually instead of random and headed back to the 1.44M drive. Wow - almost all reads are in the 600-900us range, average sector access time 2ms! 

I would suggest we look further into the fs/block_dev.c code - since we're not actually dealing only with the DF driver and PT library here, there is a possibility something else is happening when application `read` requests are sent to the raw char device in order, with no seeks. For whatever reason, any filesystem code that somehow caches blocks unknowingly to us would also explain this problem. I'll take another look at your TLVC code. I suggest adding printk's within block_dev.c, possibly also using get_ptime (we can mix kernel and user get_ptime).

> I'm leaning towards keeping the extension to %k formatting.

Why? I was about to remove it. You're not using it and don't plan to, right?

Mellvik:
> I've checked and double-checked get_ptime and am not seeing anything. That doesn't mean its not there, but if we need to test get_ptime with sub-10ms intervals, we need to think up another way to do it from user space. Possibly using plain memcpy with a known buffer size, calculated as I talked about above to ensure we're close to what we think it should be.
> 
> > So I changed the fdtest` setup to read consecutive sectors individually instead of random and headed back to the 1.44M drive. Wow - almost all reads are in the 600-900us range, average sector access time 2ms!
> 
> I would suggest we look further into the fs/block_dev.c code - since we're not actually dealing only with the DF driver and PT library here, there is a possibility something else is happening when application `read` requests are sent to the raw char device in order, with no seeks. For whatever reason, any filesystem code that somehow caches blocks unknowingly to us would also explain this problem. I'll take another look at your TLVC code. I suggest adding printk's within block_dev.c, possibly also using get_ptime (we can mix kernel and user get_ptime).


Here's the thing: I'm reading individual consecutive sectors (full track), each read wrapped by get_ptime(), then doing the same but get_ptime outside the loop. Should get the same (approximately), but there is an order of magnitude difference.

I have checked block_dev.c using printk, it's doing what it's supposed to (AFAIKT)...

> 
> > I'm leaning towards keeping the extension to %k formatting.
> 
> Why? I was about to remove it. You're not using it and don't plan to, right?

That's the thing - I'm using it in a couple of places, where I know approximately what the value should be, and it's ok. Useful.
Example:
```
Average sector read time (16 repetitions)(2059836 1f6e3c): 107ms (128739)
Seek time (80 steps*2): 582ms, step rate 3.639ms (rough estimate)
```

ghaerr:
```
    save_flags(flags);
    clr_irq();                      /* synchronize countdown and jiffies */
    outb(0, TIMER_CMDS_PORT);       /* latch timer value */
    /* 16-bit subtract handles low word wrap automatically */
#ifndef __KERNEL__
    jdiff = *pjiffies - (unsigned)lastjiffies;
    lastjiffies = *pjiffies;        /* 16 bit save works for ~10.9 mins */
#else
    jdiff = (unsigned)jiffies - (unsigned)lastjiffies;
    lastjiffies = (unsigned)jiffies; /* 16 bit save works for ~10.9 mins */
#endif
    lo = inb(TIMER_DATA_PORT);
    hi = inb(TIMER_DATA_PORT) << 8;
    restore_flags(flags);
// the above code disables interrupts then reads current jiffies and PIT countdown register together.
// the last-read jiffies is saved (only 16-bits save is required for speed, good for ~10 minutes)
// the jdiff value is calculated which is the difference in jiffies between calls.

    count = lo | hi;
    pticks = lastcount - count;
    if ((int)pticks < 0)            /* wrapped */
        pticks += MAX_PTICK;        /* = MAX_PTICK - count + lastcount */
    lastcount = count;
// the above code detects a wrap through zero and adds 11932 to adjust to actual
// pticks (0..11931). This will always be in the sub-10ms range.

    if (jdiff < 2)                  /* < 10ms: 1..11931 */
        return pticks;
// if the jiffies difference is zero, then no wrap occurred, since the HW timer IRQ fires
// at 0. If the jiffies difference is one, then the wrap occurred, and the pticks value
// will have been updated from its negative value t by adding 11932.
// Thus effectively when jdiff is 0, the sub-10ms interval occurred without the PIT wrapping,
// and when jdiff is 1, the sub-10ms interval wrapped and crossed a HW IRQ timer interrupt boundary.

    if (jdiff < 4286)               /* < ~42.86s */
        return (jdiff - 1) * (unsigned long)MAX_PTICK + pticks;
// if jdiff >= 2 we have at least a 10ms interval plus the corrected pticks.
// When jdiff is exactly 2, we return 10ms + pticks, etc for increasing jdiffs

```
The basic operation of get_ptime is documented above. Do you follow it? Any comments? This code was extensively tested in kernel mode, including for 0, 10, 20 and 30+ms intervals. I can pull that code out again for testing in kernel mode, and possibly in user mode. But I want to make sure we're not dealing with a possible raw char driver bug... do we know for sure that the raw char driver always returns the correct data, i.e. has it been tested with `cmp` etc for various copies?

Mellvik:
I guess the most important point is that it all works fine on the slower drive when all the values are above 10ms. block_dev.c couldn't possibly distinguish between those I don't think.

ghaerr:
> I'm reading individual consecutive sectors (full track), each read wrapped by get_ptime(),

> block_dev.c couldn't possibly distinguish between those I don't think.

The issue for block_dev.c differentiation isn't the slower drive(s), but that the file position filp->f_pos isn't random anymore, but sequential. There may be another level doing some buffering (e.g. stdio definitely would). I still can't see your fdtest.c code, so guessing in the dark here.

ghaerr:
> I guess the most important point is that it all works fine on the slower drive when all the values are above 10ms.

Ok, I see what you mean. Its hard to follow all the info, but you're saying that multiple get_ptime's > 10ms seem to work fine but multiple get_ptime's < 10ms (only with the faster driver that seems process 8.1ms sector reads quickly) report very low values. Right?

Mellvik:
> > I guess the most important point is that it all works fine on the slower drive when all the values are above 10ms.
> 
> Ok, I see what you mean. Its hard to follow all the info, but you're saying that multiple get_ptime's > 10ms seem to work fine but multiple get_ptime's < 10ms (only with the faster driver that seems process 8.1ms sector reads quickly) report very low values. Right?

That's right. I was pondering the idea of adding a scaleable delay to find the exact point where the values change from low to correct, let me know if you think that is a reasonable step forward.

Mellvik:
> do we know for sure that the raw char driver always returns the correct data, i.e. has it been tested with cmp etc for various copies?

yes, the raw driver is used daily (df and dhd) for fdisk, mkfs, fsck and other purposes. It's mandatory in some of them, I forget which (including fdtest, should probably be renamed to dftest 😎

Mellvik:
@ghaerr - thanks for the link to the (big) linux floppy config doc. Wow, lots of people have really scratched their heads to tame this monster. We're in good company. Makes me feel slightly better when hitting the wall. Btw, I noticed there is no mention of a cache, and I'm pretty sure there ist't one. Might get tempted to scan through the code to find out.

Mellvik:
@ghaerr, 
Reading the code, isn't `pticks = lastcount - count;` always negative? Presumably `count` must be larger than `lastcount`?

ghaerr:
> I'm using it in a couple of places, where I know approximately what the value should be, and it's ok. Useful.

I've already scanned the entire (modern?) Linux floppy driver - to this day its eerily similar to ours, but of course w/32-bit address space lots of extra stuff added. I'm almost positive there's no cache - we invented it - likely because of the timing results, and if the raw driver is correct, then problem's probably in the get_ptime code and our rabbit search can come to an end.

The Linux floppy driver does have some cool stuff which I took note of - regarding the "broken" drive change line, there's a boot option to reset the drive media type every driver open (or not) versus using DCL hardware. 

I also ran into a (discontinued) Linux option in some other documentation called readahead which attempts to do what I started with the latest `mfs` enhancement. But the cool option was that the kernel could be passed a 'profile' boot option that actually has the kernel record which files are opened when during boot (into /etc/readahead/*) and later used to layout the boot image (along with some defrag stuff). I'm thinking we could have the same option that would allow for semi-automatic setup of images using the same technique, then passing that information to `mfs` to build a customized boot floppy :)

Another technique used in Linux (conceived for optimized page fault handling) is a `readahead` system call. For our purposes, readahead could be set at boot, then turned off, or vice versa or using sysctl, that would enable/disable certain useful floppy buffer system or cache optimizations when required for best performance, based on what we're learning here.

Agreed also that the Linux floppy config doc has tons of interesting stuff. Not that we have code space or desire for most of it, but yes - lots of options for floppy operation, way more than I had imagined!

ghaerr:
> Presumably count must be larger than lastcount?

No - exactly the opposite - the PIT counts *down* not up - so its a little confusing.

ghaerr:
@Mellvik: I think I have found the problem (thanks to your testing, as usual). When the two PIT timer values are compared (subtracted), the result is correct when less than 10ms has elapsed, but erroneous when > 10ms && < 20ms have elapsed. Not quite sub-10ms but in between that and sub-20ms.

I don't have a fix yet. Coming.

ghaerr:
> It seems to me we may have a bug in get_ptime 

You were right, get_ptime was buggy. Excellent debugging and testing work @Mellvik!! Very impressive!! I also apologize for the error. I had thought it to be well tested and had played with QEMU results for quite a while when developing the routine. What I missed, and won't miss again for this sort of thing is unit testing - instead of "eyeballing" the output it should have been verified with a unit test, either inside or outside the kernel. Not so much for the "exact" timing but the calculation itself. Going through writing the unit test this time really helped point out the problem and verify the solution was correct. There's a lot more that needs to be done on the unit test approach but we can talk about that another time.

The fix is https://github.com/ghaerr/elks/pull/2088 and contains the exact same file for both tlvc/arch/i86/kernel/lib/prectimer.c and libc/debug/prectimer.c. I also uploaded the unit test which shows the original calculation producing incorrect results and now correct results, for each case of count, lastcount and jiffies.

The fix contains both the corrected new and original bad code using `#if 1 / #else` for comparison. This will be deleted after your testing. I am also quite sure this fix will help with the problem(s) you've been having getting accurate, repeatable calibration and delay routines working. Also, this new version takes the same time in the routine regardless of the elapsed time, so results should be more consistent, if the time for executing the routine itself is subtracted.

Thank you!!

Mellvik:
Thank you @ghaerr and congratulations - this is really GREAT news. You know what it feels like when this itch (or worse) has been bugging you 24/7 for days, goes away. No apologies required - this is continuous learning (and the nature of software development). It's interesting that a timer bug turned us into southbridge experts, well, at least even more of FDC experts than we were before. Funny story really, and a great illustration of the unpredictable paths debugging (rabbit chasing) sends us onto.

I'm elated to report that the new version works great. Adjacent sectors on 1.44M and 1.2M drives now report 10.75ms and 10.5ms respectively. That's perfect. As to the kernel version - my calibrate timer routine now report reasonable numbers consistently. That's after a single reboot on one (of 4 active) machine(s), so there is more to be experienced. I have not taken the time to fully understand the bug and the fixes, but the results speak for themselves - and unit testing, well, what can I say?

As to the mystery that got me started on this path to begin with, the apparent inefficiency of a sector cache on fast machines, the conclusion is that that's what happens when the machines outpace the storage technology to the extreme. A 386 and later is actually better off reading a single sector, go do other stuff, come back and read the next, instead of starting a long (and automatic, multisector) read/write operation. Now that we have completely reliable timings to lean on I'm going back to confirm some of these numbers in order to establish the smart(test) thing to do with floppy IO.

Thanks again.

ghaerr:
> the apparent inefficiency of a sector cache on fast machines, the conclusion is that that's what happens when the machines outpace the storage technology to the extreme. A 386 and later is actually better off reading a single sector, go do other stuff, come back and read the next, instead of starting a long (and automatic, multisector) read/write operation. 

Well said, and new knowledge (for us). It might be worth considering using the *ratio* of the CPU speed (variable) to the floppy I/O speed (fixed) as a metric for optimizing expected cache performance. At some point, the CPU speed far outpaces the fixed I/O subsystem(s). By accurately measuring the I/O times like you're doing, they could be used as the denominator against calculated or measured times for CPU throughput (e.g. fmemcpy) and use the ratio as a metric for expected system performance. It seems to me we're just coming to fully appreciate the reality of the different execution timings, thanks to the precision timing being used.

On early legacy systems I now realize that the system is consistently heavily bottlenecked by the CPU, not I/O. A great example of this is in the conversations in https://github.com/ghaerr/elks/issues/2040, where the `sl` (train) program is timed drawn across the screen. On early 8088 systems it takes 1m57s while on a Pentium III the *same program* takes 0.09s. Wow!!!

The cost of copying buffers and other "mundane" code is barely noticeable on modern processors - but sometimes a heavy cost on legacy systems, and relatively unnoticed. This may open up another door for better performance on early systems.

Mellvik:
Thanks @ghaerr, 
I concur on most points. I wrote a summary after mote testing with the new `fdtest` this morning, let's continue the discussion after I post it.

With the new prectimer I need to redo some of the work on delay calibration, hopefully some of the 'magic' can be simplified. The calibration establishes a performance index for the system, and may come in handy in this context. However I do hope that determining processor will be enough to autoconfigure fdcache (for example).

Mellvik:
Floppy performance testing with the new fdtest program (29/10/24)
The testing confirm what we've suspected for a while. We now have real numbers to rely on when choosing optimal strategies for systems.
- There is a BIG difference between XT class and 386 class systems
- Pending: 286 based AT systems may belong to the XT category in this context.
- XT class:
    - Any multisector IO is VERY advantageous.
    - Reading/writing a full track (9 sectors) is faster than reading 2 (consecutive) sectors individually
    - Reading/writing 2 sectors in one op is only slightly slower than recording a single sector.
    - The explanation is that these systems will lose a revolution per sector every time when accessing individual sectors.
    - This is underscored by the fact that reading random single sectors on a track is almost twice as fast as reading them sequentially.
- 386Class - the numbers are almost the opposite (on an idle system)
    - Multisector IO differs very little from single sector IO. IOW, the system is fast enough to read single sectors on the fly.
    - Reading/writing 2 sectors in one op is - on average - 10% slower than reading them separately  - because we're reading time spent, not time elapsed.
    - Reading/writing a full track in 1 op clocks in at about the same time as doing single sector (sequential) IO (impressive, but unlikely to matched on a busy system)
    - Average random single sector IO OTOH is about the same as for XT class, as it must be given that there will be - on average - half a revolution between each.
    - There is minimal difference between the numbers from a 386DX/20MHz and a 386SX/40MHz, which is to be expected: we're doing IO at the physical speed of the floppy subsystem.  
    - It's important to keep in mind that the numbers are collected on a quiet system. It takes only a little additional activity to disturb this picture. In particular, believing that individual sector IO will always be faster than mulitsector,  is wrong. IOW - keeping double sector reads for block IO is definitely good regardless of system. And — as the numbers suggest, using full track continuous IO instead of individual sectors - when all those sectors are actually needed, is good, not only because the measured timing is about the same (on a quiet system) but because it allows the system to do other activities while the IO is going on.

Thus, the conclusion is
- Multisector floppy IO is good/useful for all system classes (IOW we're good already).
- Floppy cache is great for XT class systems, negative for faster systems. Although there will always be usage scenarios that deliver different conclusions,  this  holds for general usage (including system startup) . The TLVC directfd driver covers this but currently only manually. Automatic cache calibration is desirable (nice term for 'turning off the cache if fast cpu').
- The definition of 'faster systems' is still open, pending more 286 class testing.

Mellvik:
Update about numbers: It suddenly occurred to me that I don't need to pull out the 286 AT system again to determine the question above, about which systems need/need not fdcache. We have that already - in the system startup benchmarks from a few weeks back. The 12.5MHz 286 definitely benefits from the floppy cache and clocks in with best timings at 6k cache size.

So it seems the reasonable conclusion is to turn off the cache if 386CPU or above.
 
As to determining cputype  - @ghaerr, the book you probably have received by now, "The Processor and Coprocessor" has a listing (pages 76-81) of a rather (relatively) simple asm routine to determine type up to 486.

ghaerr:
Thank you for this well-written summary and conclusion. I will use it during subsequent research and development in the ELKS kernel and DF driver.

I'll go ahead and put together a PR for recognizing a 386(+) CPU from the tested XMS 386 cpu test code, which will run in setup.S/cputype.S and end up setting arch_cpu = 7 when the CPU is 386 or later. You can then use that for creating an automated effective fdcache=0 setting on 386 boot.

Mellvik:
> I'll go ahead and put together a PR for recognizing a 386(+) CPU from the tested XMS 386 cpu test code, which will run in setup.S/cputype.S and end up setting arch_cpu = 7 when the CPU is 386 or later. You can then use that for creating an automated effective fdcache=0 setting on 386 boot.

Thanks - appreciated. It wasn't mentioned. in the PR notes, but I've changed the default fdcache RAM allocation in `config.in` to 6k, which seems like a reasonable value - for now.

PR #100, opened 2024-10-30
[cmd] More fdtest updates/fixes

Mellvik:
Mostly display updates per comments from @ghaerr (thank you), and a couple of fixes.

Block size is now clustersize, the `-b` option for clustersize kept for lack of anything better.

```
tlvc17# ./fdtest -d 1 -D -c 50
Reading /dev/rdf1 from cyl 50, spt 15, clustersize 15 sectors
............................................................
Read 60 clusters in 15.0s (0.250s per cluster)
tlvc17# ./fdtest
/dev/rdf0: Size 1440k, 18 sects/trk, 80 cyls, clustersize 18 sects,
	   start: 3/0/1 (lba 108)
18 sectors (1-by-1):	365.509ms (20.305ms/sector)
18 1k blocks:		815.457ms (45.303ms/block)
18 sectors in 1 op:	365ms
tlvc17# 
```

### DIFF
--- a/tlvccmd/misc_utils/fdtest.c
+++ b/tlvccmd/misc_utils/fdtest.c
@@ -50,7 +50,7 @@ static unsigned long av_sec_time(int fd, int spt)
 		get_ptime();
 		if (read(fd, iobuf, SECSIZE) <= 0) return 0;
 		if (debug) {
-			sec_arr[i] = next;
+			sec_arr[i] = next+1;
 			time_arr[i] = get_ptime();
 		} else
 			a += get_ptime();
@@ -58,13 +58,13 @@ static unsigned long av_sec_time(int fd, int spt)
 	}
 	if (debug) {
 		for (i = 0; i < 16; i++) {
-			fprintf(stderr, "%d (%d): %lk %lu\n", i, sec_arr[i], 
+			fprintf(stderr, "%d (%d):\t%lk\t%lu\n", i, sec_arr[i], 
 				time_arr[i], time_arr[i]);
 			a += time_arr[i];
 		}
 		fprintf(stderr,
-			"Average sector read time (%d repetitions): %#lk (%lupt)\n", 
-			16, a>>4, a>>4);
+			"Average sector read time (%d repetitions): %#lk\n", 
+			16, a>>4);
 		return a>>4;
 	} else {
 		a >>= 4;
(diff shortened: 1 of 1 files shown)


ghaerr:
I can see `fdtest` is fast becoming a very useful tool for determining I/O bandwidth, pretty cool! It is quite nice to see how fast (or not) various combinations of floppy reading are, and I like the how it works with no arguments just to give a quick indication of how fast the disk I/O subsystem takes. I assume it takes a raw char device on the command line if no preceding '-' present?

It is however limited to raw devices only, but could be useful for comparing the older BIOS implemention as well as raw (no pun intended) speeds for block devices in general. Perhaps consider renaming to `iotest` and adding the ability to open any /dev device, including block devices. Of course, for block devices, the buffer system would be in use, and cluster size would always be 2, but still IMO useful, if for nothing else comparing raw vs buffered I/O. This should all be easily possible, since it's using standard syscalls for everything, right?

If you're not interested, I may take a look at that conversion and port over to ELKS. I remain interested in just how fast/slow the DF driver is vs BIOS, for a number of reasons, and this seems like it could become just the tool for that sort of thing.

If a block device were specified on the command line, the program could override the cluster size and set it to 2. I haven't yet looked at other considerations since I've not studied the source yet (lots of other things to do).

Mellvik:
As is, `fdtest` can access raw floppy devices only  (inheritance from your direct bios original implementation) and there are a number of floppy specifics built in. That's easy to change if deemed useful, but my view at this time is that `dd` can do most if not all testing/benchmarking of hard drives, block or raw. Admittedly with measurements limited to what the `time` command can provide. However, with the possible exception of xd (mfm) drives, doing low level benchmarking or measurements on hard disk (that is, IDE disks) appears much less useful than floppies because on-drive buffering and other 'smart' capabilities hide the physical drives and do optimizations on their own.

I would think that in order to make fdtest useful on ELKS, all you need to do is to add taking the device name from the command line. I would (of course) recommend though to add raw support to the ELKS floppy driver. With the exception of the 'raw front end' in block_dev.c it's a wrapper with few codelines. IMHO it's a great advantage to get the disk utilities away from the buffer system. 

Anyway, and on a different note, you may (not) have noticed that __divmod made it into the `fdtest` as the first user program on TLVC (outside printf). Thanks.

Mellvik:
This final commit adds another couple of bug fixes. More importantly support for providing the device to test on the command line has been added.

In order to use, `HDTEST` must be defined in the source and the program must be called by the name `hdtest`(as opposed to `fdtest`). Also, as is, the device name provided on the command line (must be the last argument) must contain `/dev/` and the subsequent 'r' identifying a raw device. This can easily be commented out if required.

@ghaerr, I hope this helps to make the tool useful on ELKS 'as is'.

PR #101, opened 2024-10-31
[kernel] Add cpu type detection for 32 bit CPUs (80386+)

Mellvik:
Add detection of 32 bit CPUs (80386 and higher) to the system in order to automatically scale the floppy sector cache (ref discussion in #97). From https://github.com/ghaerr/elks/pull/2089#issue-2625694587 --- Thank you @ghaerr .

Also includes @ghaerr's code for tracing (debug) system setup!

The SETUP_CPU_TYPE macro is used to get the numeric (byte size) cpu type, 7 is 386 and up, 6 is 286.

From boot messages - 386SX, 386DX, V20 respectively 
```
PC/AT class machine (type 7), syscaps 0xff, 640K base ram, 17 tasks.
PC/AT class machine (type 7), syscaps 0xff, 640K base ram, 17 tasks.
PC/XT class machine (type 2), syscaps 0x0, 640K base ram, 17 tasks.
```


ghaerr:
> Also includes @ghaerr's code for tracing (debug) system setup!

You'll have to enable the system setup trace code - a lot can be learned by tracing through through it. It's a bit complicated, but I tried to format it to the screen for readability. There's another option that allows for all the info to remain on the screen until a key is pressed.

Mellvik:
> > Also includes @ghaerr's code for tracing (debug) system setup!
> 
> You'll have to enable the system setup trace code - a lot can be learned by tracing through through it. It's a bit complicated, but I tried to format it to the screen for readability. There's another option that allows for all the info to remain on the screen until a key is pressed.

Thanks! I'll test it in the morning and make sure I got all the pieces in place.

Mellvik:
I planned the `directfd` update to go in its own PR, but missed. 

Anyway - the 'autoconfig' is very simple: Use `/bootopts` if present, otherwise either 1k (32bit system) or 6k. As I write this I'm thinking using the configured value from MENUCONFIG is probably better than just setting 6 arbitrarily.

PR #102, opened 2024-11-05
[kernel,fs] Udates xms support and various cleanups

Mellvik:
This PR adds `linear32_fmemset` to the xms code and introduces a number of cleanups from ELKS.

It also introduces an elusive bug - yet to be located: When XMS code is compiled in, part of the console output from the boot process is garbled - as if the console port serial speed gets disturbed. Fix for this is coming shortly.  

```
Direct console, polling kbd 80x25 emulating ANSI (3 virtual consoles)
xms: using unreal mode, 100 xms buffers (100K ram), 10K cache, 20 req hdrs
eth: ne0 at 0x300, irq 12, (ne2k) MAC 52:54:00:12:34:56 (QEMU), flags 0x80, bufs 2r/0t
eth: wd0 at 0x280, irq 2, ram 0xcc00 not found
eth: ee0 at 0x360, irq 11, (ee16) not found
athd0: AT/IDE controller at 0x1f0
athd0d0: IDE CHS: 63/16/63 , Multisector I/O, max 16 sects
athd1: Controller not found at 0x170 (376)
athd: found 1 hard drive 
df: Direct floppy driver, FDC 8272A @ irq 6, DMA 2
df0: 1.44M (type 4) [CMOS]
Partitions: dhda:(0,63504) no mbr (flat drive assumed)
?????1?1??߾?ُ?G?<Qˈ?????t???C??ó?U??PRQS????&??1?????????̈?1Ҋ?0???A??(ӋF?9?v??^?????u?8?v???PQR?F?^???s0???*?x??N?ZYX?u?X?O??F??.?a?ZYX?)F?tF???S????&??1?????????̈?1Ҋ?0???A??(ӋF?9?v??^?????u?8?v???PQR?F?^???s0???*?x??N?ZYX?u?X?O??F??.?a?ZYX?)F?tF??????̈?1Ҋ?0???A??(ӋF?9?v??^?????u?8?v???PQR?F?^???s0???*?x??N?ZYX?u?X?O??F??.?a?ZYX?)F?tF???
Delay calibration index: 115, skew: 0

w1??]_^á

ű???뼋?????F?tF???F??g???]?Running /etc/rc.sys script
Starting networking on ne0
ktcp -b -p ne0 10.0.2.15 10.0.2.2 255.255.255.0
ktcp: ip 10.0.2.15, gateway 10.0.2.2, netmask 255.255.255.0
ktcp: /dev/ne0 mac 52.54.00.12.34.56 mtu 1500
Tue Nov  5 15:46:31 2024
[35]

[tlvc15] TLVC 0.6.0

login: 
```

### DIFF
--- a/tlvc/arch/i86/lib/unreal.S
+++ b/tlvc/arch/i86/lib/unreal.S
@@ -40,6 +40,7 @@
 	.global	set_a20
 	.global	linear32_fmemcpyw
 	.global	linear32_fmemcpyb
+	.global	linear32_fmemset
 
 # Check if unreal mode capable. Currently requires 32-bit CPU (386+)
 # Returns 1 if OK, otherwise error code (-1=not 386, -2=in V86 mode).
@@ -402,6 +403,55 @@ linear32_fmemcpyb:
 	mov    %ss,%ax
 	mov    %ax,%ds
 	pop    %es
+	ret
+
+#
+# void linear32_fmemset(void *dst_off, addr_t dst_seg, byte_t val, size_t count)
+# WARNING: Requires 32-bit CPU, with unreal mode and A20 gate enabled!
+#          Trashes EAX, EBX, ECX, ESI and EDI without saving.
+#
+linear32_fmemset:
+	push   %si
+	push   %es
+	mov    %di,%dx
+
+	//pushf               // save interrupt status
+	//cli                 // uncomment if extended registers used in interrupt routines
+	xorl   %esi,%esi
+	mov    %sp,%si
+
+	xorl   %ecx,%ecx
+	mov    14(%si),%cx    // word count -> ECX
+
+	xorl   %ebx,%ebx
+	mov    6(%si),%bx     // word dest offset -> EBX
+	movl   8(%esi),%edi   // long dest address -> EDI
+	addl   %ebx,%edi      // EDI is linear destination
+
+	xorl   %eax,%eax
+	mov    12(%si),%al    // byte val -> EAX
+	mov    %al,%ah
+
+	xor    %bx,%bx        // ES = DS = 0 for 32-bit linear addressing
+	mov    %bx,%es
+	mov    %bx,%ds
+
+	cld
+	shrl   $1,%ecx        // store words
+	addr32 rep stosw      // word [ES:EDI++] <- AX, ECX times
+	addr32 nop            // 80386 B1 step chip bug on address size mixing
+
+	rcll    $1,%ecx       // then possibly final byte
+	addr32 rep stosb      // byte [ES:EDI++] <- AL, ECX times
+	addr32 nop            // 80386 B1 step chip bug on address size mixing
+
+	//popf                // restore interrupt status
+
+	mov    %dx,%di
+	mov    %ss,%ax
+	mov    %ax,%ds
+	pop    %es
+	pop    %si
 	ret
 
 # The GDT contains 8-byte DESCRIPTORS for each protected-mode segment.
--- a/tlvc/arch/i86/mm/xms.c
+++ b/tlvc/arch/i86/mm/xms.c
@@ -4,12 +4,12 @@
  * Nov 2021 Greg Haerr
  */
 
-#include <linuxmt/types.h>
+#include <linuxmt/config.h>
 #include <linuxmt/memory.h>
 #include <linuxmt/kernel.h>
 #include <linuxmt/init.h>
-
 #include <linuxmt/string.h>
+#include <linuxmt/debug.h>
 #include <arch/segment.h>
 
 /* linear address to start XMS buffer allocations from */
@@ -51,10 +51,10 @@ int xms_init(void)
 		return 0;
 	}
 #endif
-	printk("A20 was %s", verify_a20()? "on" : "off");
+	debug("A20 was %s", verify_a20()? "on" : "off");
 	enable_a20_gate();
 	enabled = verify_a20();
-	printk(" now %s, ", enabled? "on" : "off");
+	debug(" now %s, ", enabled? "on" : "off");
 	if (!enabled) {
 		printk("disabled, A20 error, ");
 		return 0;
--- a/tlvc/fs/buffer.c
+++ b/tlvc/fs/buffer.c
@@ -1,4 +1,5 @@
-#include <linuxmt/types.h>
+#include <linuxmt/config.h>
+#include <linuxmt/limits.h>
 #include <linuxmt/init.h>
 #include <linuxmt/sched.h>
 #include <linuxmt/kernel.h>
@@ -6,12 +7,9 @@
 #include <linuxmt/string.h>
 #include <linuxmt/mm.h>
 #include <linuxmt/heap.h>
-#include <linuxmt/config.h>
-#include <linuxmt/fs.h>
-#include <linuxmt/limits.h>
 #include <linuxmt/errno.h>
-#include <linuxmt/debug.h>
 #include <linuxmt/trace.h>
+#include <linuxmt/debug.h>
 
 #include <arch/system.h>
 #include <arch/segment.h>
@@ -20,19 +18,21 @@
 
 /*
  * Kernel buffer management.
+ *
+ * Aug 2023 Greg Haerr - Added dynamic L1 buffers and release L1 mappings during sync.
  */
 
-/* Number of internal L1 buffers/cache blocks,
- * either used as the system buffer pool (no L2) or
- * to map/copy external L2 buffers to/from kernel data segment,
- * default is 12, may be overridden by 'cache=' in /bootopts. */
-int nr_mapbufs = 12;
-#define MAX_NR_MAPBUFS 24
+/* Number of internal L1 buffers, used to map/copy external L2 buffers
+   to/from kernel data segment. */
+int nr_map_bufs = NR_MAPBUFS;                   /* override with /bootopts cache= */
+#define MAX_NR_MAPBUFS  20
 
-/* The L1 buffers/cache is allocated from heap later. */
-static unsigned char *L1buf;
-
-int L2_bufs;		/* /bootopts # buffers override */
+#ifdef CONFIG_FS_EXTERNAL_BUFFER
+int nr_ext_bufs = CONFIG_FS_NR_EXT_BUFFERS;     /* override with /bootopts bufs= */
+#endif
+#ifdef CONFIG_FS_XMS_BUFFER
+int nr_xms_bufs = CONFIG_FS_NR_XMS_BUFFERS;     /* override with /bootopts xmsbufs= */
+#endif
 
 /* Buffer heads: local heap allocated */
 static struct buffer_head *buffer_heads;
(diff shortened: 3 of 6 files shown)


ghaerr:
> When XMS code is compiled in, part of the console output from the boot process is garbled

Never seen that before. I would guess problem has to do with ramdesc_t being 32 bit rather than 16, and/or %s taking a bad pointer and spitting out garbage. Unlikely anything to do with baud rates. It'll be the printk right after the "Partition" output.

I see in the buffer.c change (hard to tell since it looks to be retabbed but all within the same commit) that special ifdef code added so buffer offset is always zero - why?

Mellvik:
> > When XMS code is compiled in, part of the console output from the boot process is garbled
> 
> Never seen that before. I would guess problem has to do with ramdesc_t being 32 bit rather than 16, and/or %s taking a bad pointer and spitting out garbage. Unlikely anything to do with baud rates. It'll be the printk right after the "Partition" output.

Thanks, yes - that's most likely it.

> 
> I see in the buffer.c change (hard to tell since it looks to be retabbed but all within the same commit) that special ifdef code added so buffer offset is always zero - why?

I should have taken the time to fix the tabbing. Instead of importing a zillion small updates from the ELKS version of the file, I did it the other way around - took the ELKS version and worked in the TLVC changes. Can you be more specific about the question - location?

Thank you.

ghaerr:
> Instead of importing a zillion small updates from the ELKS version of the file, I did it the other way around - took the ELKS version and worked in the TLVC changes.

Oh, great - for internal kernel stuff, the method you used is actually more convenient for me to help with identifying issues. Having TLVC and ELKS core internals match somewhat closely is probably a good thing, since both of us have helped find and fix deep reliability issues for both systems, and allows both systems to remain very solid as far as not crashing unexpectedly.

> I should have taken the time to fix the tabbing. 

No problem, I now understand what you've done. I normally only post an initial retab commit when first retabbing a file, for both your and my sake, but when a file is updated directly from ELKS to TLVC or vice versa, there's no need.

> Can you be more specific about the question - location?

I just pulled down your version and find that the question I asked was - within my own code originally! LOL. So much for my memory improving!

Mellvik:
> > Instead of importing a zillion small updates from the ELKS version of the file, I did it the other way around - took the ELKS version and worked in the TLVC changes.
> 
> Oh, great - for internal kernel stuff, the method you used is actually more convenient for me to help with identifying issues. Having TLVC and ELKS core internals match somewhat closely is probably a good thing, since both of us have helped find and fix deep reliability issues for both systems, and allows both systems to remain very solid as far as not crashing unexpectedly.

My thinking exactly. The merge turned out quite easy actually - mostly a related to ASYNC and a few lines to accommodate raw IO. Plus we had quite different setups/formats for the debug listings that did the same things.

> > Can you be more specific about the question - location?
> 
> I just pulled down your version and find that the question I asked was - within my own code originally! LOL. So much for my memory improving!

Sounds vaguely familiar - cannot remember why ... :-)

ghaerr:
> What happens on the compaq 386 is that it boots fine, the stops in the display partition function:

> xms: using unreal mode

Looks like you have your Compaq XMS configured for unreal mode, not INT15. That's a problem, interestingly enough just discussed above. Could still be a %s issue, but if its not hanging rather than printing garbage, then likely just misconfiguration. Note that the partition code will be the first I/O the HD driver makes, for your rabbit chasing consideration.

Even though TLVC may be (is it?) completely free from the BIOS, that doesn't necessarily guarantee that unreal mode can be used for XMS instead of INT 15. And of course INT 15 is a BIOS function, so TLVC may still be dependent on the Compaq BIOS for operation!

Mellvik:
> > What happens on the compaq 386 is that it boots fine, the stops in the display partition function:
> 
> > xms: using unreal mode
> 
> Looks like you have your Compaq XMS configured for unreal mode, not INT15. 

That was intentional because you asked (and unrelated to the garbage print issue). 

>> I plan on enabling XMS as default shortly to all configurations, and wanted to make sure it ran well. Just now I'm realizing though that enabling XMS still won't work on Compaq systems, and could result in even worse - a boot failure. Have you tested what happens when XMS is enabled NOT using INT15 on Compaq? IIRC on my system, the Compaq failed to boot, as it passed the unreal mode CPU test, but then failed since the BIOS won't work with a kernel's unreal mode and its protected mode BIOS.

So you remembered correctly, although it doesn't seem to be the BIOS related since the quoted boot sequence doesn't use the BIOS. It may be worthwhile looking a little more into that. The 286 is a different story, requiring INT15 regardless of course.

As to the garbage print issue, I'll take a look at the object dump and find which variable is just in front of of the printf format string(s) that go astray. I subscribe to your ramdesc_t idea.

ghaerr:
> Partitions: dhda:(0,1000944) <---- hangs here 

> That was intentional because you asked (and unrelated to the garbage print issue).

Oops, thanks! I think I got confused because from a quick look, it crashes at the exact same place the garbage print issues starts... (?!)

> it doesn't seem to be the BIOS related since the quoted boot sequence doesn't use the BIOS. It may be worthwhile looking a little more into that. 

Not sure exactly what you mean by "quoted boot sequence" but the TLVC/ELKS boot does in fact still make BIOS calls during startup. Of course, not from the DF driver, and most IIRC happen in setup.S but there may be others snuck in there. As you mention, the trouble seems to start well past any early kernel startup, and partition check is the last part of kernel startup in device_init(), just before the below boot message, which I also don't see in your display.
```
printk("boot: BIOS drive %x, root device %04x\n", ROOT_DEV, rootdev);
```

> Partitions: dhda:(0,63504) no mbr (flat drive assumed)
?????1?1??߾?ُ?G?<Qˈ?????t???C??ó?U??PRQS????&??1?????????̈?1Ҋ?0???A??(ӋF?9?v??^?????u?8?v???PQR?F?^???s0???*?x??N?ZYX?u?X?O??F??.?a?ZYX?)F?tF???S????&??1?????????̈?1Ҋ?0???A??(ӋF?9?v??^?????u?8?v???PQR?F?^???s0???*?x??N?ZYX?u?X?O??F??.?a?ZYX?)F?tF??????̈?1Ҋ?0???A??(ӋF?9?v??^?????u?8?v???PQR?F?^???s0???*?x??N?ZYX?u?X?O??F??.?a?ZYX?)F?tF???
Delay calibration index: 115, skew: 0
> w1??]_^á
ű???뼋?????F?tF???F??g???]?Running /etc/rc.sys script

What's a bit strange about this is that the garbage starts, then stops for your delay calibrate printk, then starts again?

Mellvik:
> > Partitions: dhda:(0,1000944) <---- hangs here
> 
> > That was intentional because you asked (and unrelated to the garbage print issue).
> 
> Oops, thanks! I think I got confused because from a quick look, it crashes at the exact same place the garbage print issues starts... (?!)

Hmm, I didn't notice that but you're right of course - the garbled print and the XMS Compaq hang occurs at exactly the same place. May be a coincidence and it may not. The only new addition to the xms code in TLVC since it was working (that's probably 6 months ago though) is the `linear32_fmemset` - as mentioned in the PR notes.

> As you mention, the trouble seems to start well past any early kernel startup, and partition check is the last part of kernel startup in device_init(), just before the below boot message, which I also don't see in your display.
> 
> ```
> printk("boot: BIOS drive %x, root device %04x\n", ROOT_DEV, rootdev);
> ```

This message doesn't exist in TLVC, the 'normal' boot output goes like this:
```
athd: found 1 hard drive 
df: Direct floppy driver, FDC 82077 @ irq 6, DMA 2
df0: 1.44M (type 4) [CMOS]
Floppy cache 0k, available 6k
Partitions: dhda:(0,63504) no mbr (flat drive assumed)
PC/AT class machine (cpu 7), syscaps 0xff, 639K base ram, 17 tasks.
TLVC 0.6.0 (63136 text, 16592 ftext, 10592 data, 15904 bss, 39038 heap)
Kernel text 250:0, ftext 11ba:0, init 1371:0, data 15c7:0, top 9fc0:0, 497K free
MINIX: inodes 2048 imap 1 zmap 4, first data 71
MINIX-fs: mounting unchecked file system 0x500, running fsck is recommended.
VFS: Mounted root 0x0500 (minix filesystem).
[6]Running /etc/rc.sys script
```

> 
>  > Partitions: dhda:(0,63504) no mbr (flat drive assumed)
> > ?????1?1??߾?ُ?G?<Qˈ?????t???C??ó?U??PRQS????&??1?????????̈?1Ҋ?0???A??(ӋF?9?v??^?????u?8?v???PQR?F?^???s0???_?x??N?ZYX?u?X?O??F??.?a?ZYX?)F?tF???S????&??1?????????̈?1Ҋ?0???A??(ӋF?9?v??^?????u?8?v???PQR?F?^???s0???_?x??N?ZYX?u?X?O??F??.?a?ZYX?)F?tF??????̈?1Ҋ?0???A??(ӋF?9?v??^?????u?8?v???PQR?F?^???s0???*?x??N?ZYX?u?X?O??F??.?a?ZYX?)F?tF???
> > Delay calibration index: 115, skew: 0
> > w1??]_^á
> > ű???뼋?????F?tF???F??g???]?Running /etc/rc.sys script
> 
> What's a bit strange about this is that the garbage starts, then stops for your delay calibrate printk, then starts again?

Indeed weird, maybe a segment register is disturbed - and more or less incidentally being restored by a call, only to return back to destroyed stated upon returning. I'm looking into it.

Mellvik:
Thanks - yes, I've been working along something akin to that line, possible culprit is the directhd driver - which I'm narrowing down now. Buffer.c is definitely possible too...

Thank you

Mellvik:
> If you already know its the return from device_init() itself, then yes likely a header file issue, although your changes to buffer.c could also be a problem?

As far as I can tell, the problem appears upon return from device_init(), which must mean that the DS reg value pushed onto the stack when called has been overwritten. The fact that this appears only when XMS is active may not mean anything, there is nothing in the init HD routine that is affected by the presence of XMS. If the stack pollution idea holds, a local variable is likely the culprit and narrows down the search.

Mellvik:
> Not sure which DS push you're talking about, as %s in printk doesn't push DS and neither do the trampoline routines between for calling near procs from far procs.

Not understandable, I agree. Trying to get my head around the thunking code connecting the INITPROC to the NEAR world.

ghaerr:
The thunking code just copies the passed function arguments to the stack, but at a different offset, and then calls the near function. The reason a thunk is necessary is that, as you know, arguments are passed on the stack at a fixed offset from SP (and later BP). The near function code segment is different than the far segment, so a FAR CALL would be necessary to call the near function. That would push CS on the stack, and thus the argument offsets, e.g. 4(%bp) etc are all incorrect. The thunk instead is called which pushes the arguments (again) at the correct offset(s), calls the near function, removes the arguments, then returns to the caller. A different thunk is used based on how many fixed arguments the function has or is called with.

I highly doubt the issue is with thunks. Of course, an INITPROC mismatch causes all sorts of problems, because the stack is expecting CS (or not) and returns to the wrong address, usually crashing, but in some cases restarting the kernel. This is why I added "START" to kernel_start because the kernel was restarting in some cases and we didn't know it. The kernel restart is also not likely your problem.

Mellvik:
Thanks - that helps. I truly (still) hate the segmented architecture, yet here we are -:)

Since it seems the dataseg gets screwed, the plan is to display the DS value before and after. Of course it may be something else, but it is truly suspicious that the printks work, then stops working after a call to an INITSEG (directhd_init), but still works when calling near functions. All without affecting the flow (i.e.crashing).

Another rabbit...

PR #88, opened 2024-09-29
[kernel, directfd] Redesigned sector buffer for the directfd driver

Mellvik:
This PR introduces a new sector cache implementation replacing the old track buffer in the `directfd` driver. The 'TRACK_CACHE' configuration option is also back, for reasons detailed below. This is part of a floppy 'boot and use' performance enhancement project discussed on #81. Some specifics about the changes:
- The old track cache would always cache at least a full track, from sector 1 regardless of which sector was actually requested (not very useful and time consuming). Further the buffer size was always 9k (18 sectors) - if compiled in.
- The new sector cache has a very different approach:
    - Fills the cache from the requested sector to the end of the cylinder or end of cache, whichever comes first.
    - The cache can be any (even) number of blocks from 2 (1k) to 18 (9k) or larger, which doesn't make much sense.
    - This means
        - Cache size 1k behaves exactly like having no cache at all
        - Cache size 9k combined with 9 sectors per track floppies (360k, 720k) means full cylinder cache if the requested sector is the first on track (head) 0 of that cyl.
        - The cache size can be tuned and set to whatever works best for the actual environment.
    - Cache hits are being counted and reported on device close (release), which unfortunately will never show on a root file system.
- The cache size can be set using a third parameter to the `xtflpy=` setting in `bootopts`. This is a temporary solution until the `sysctl` facility from ELKS has been implemented. The metric is kilobytes, if missing the value is 0 which is read as 1, the minimum.
- For now, the actual cache size and the available cache size is being reported at boot time.
- The variable size and potential absence of the sector cache forced a recite of the raw driver's mechanism to handle 64k physical memory boundaries. A raw request, if its buffer spans such a boundary will blend split into 3 (occasionally just 2) physical requests, isolating the boundary block to 1k using the bounce buffer (which is the first k of the sector cache if present). This part of the driver has had limited testing.

Also on this PR:
- Simplified `menuconfig` page for block devices.

Performance
Surprisingly it turns out that fast machines (like a 40MHz 386) does not benefit from floppy caching at all. Thus turning off the cache using `menuconfig` and release most of the memory is the right thing to do. Also somewhat surprising that a smaller (than max) cache, like 6 or 7k, in most cases will give better performance than a smaller and larger cache. Check #81 for more details about this.

PR #103, opened 2024-11-07
[kernel] Reworked the use of lowmem DMASEG in xd and floppy driver + more

Mellvik:
This PR introduces the new scheme for allocating bounce buffers and floppy cache from low memory/DMASEG - discussed in https://github.com/ghaerr/elks/pull/2091 and https://github.com/Mellvik/TLVC/pull/88. The new DMASEG map has a bounce buffer for xd drives (XT class) or Lance ethernet (AT class) at the bottom, then a bounce buffer for the `directfd` driver, handling both DMA wraps and XMS bouncing. Then the floppy sector cache if configured. This means - for the `directfd` driver - that the cache and the bounce buffer are permanently separated and that the cache may be configured completely out of the code by setting the CONFIG_FLOPPY_CACHE value to 0 in menuconfig. For 386+ systems that don't need or benefit from the cache, this is meaningful memory saved.

Other changes to the `directfd` driver:
- A fix for the case that autconfig would fail on all (that is two) formats and loop endlessly. It will now try all (still 2) formats a few times, then fail.
- While it is possible to set the size of the FLOPPY cache to any number in menuconfig, the max is now 7k (14 sectors), corrected if necessary in `config.h`. As before, the actual size and the configured cache memory is reported at boot time - unless configured to 0.
- `fdcache=` in `/bootopts` may now be set to 0 and the cache will be disabled in the driver. Functionally this is similar to the way it worked before (when the bounce buffer and the cache were integrated), but then the 'turn-off' value was 1 to accommodate the bounce buffer, not 0.

Also in this PR is an update to the `xd.c` driver to work in the changes to the bounce buffer setup in `config.h`. These changes have not been tested yet for lack of working hardware. Further:
- Added DEBUG_RAW in `debug.h` and changed debug statements in `fs/block_dev.c` accordingly. `debug.h` has had a number of other additions too, part of a process to improve the debug tooling in TLVC.
- Some cleanups.

### DIFF
--- a/tlvc/arch/i86/drivers/block/blk.h
+++ b/tlvc/arch/i86/drivers/block/blk.h
@@ -213,11 +213,12 @@ static void end_request(int uptodate)
     bh->b_reqnext = NULL;
 #endif
 
-    mark_buffer_uptodate(bh, uptodate);
     mark_buffer_clean(bh);      /* EXPERIMENTAL: moved from ll_rw_blk  */
+    mark_buffer_uptodate(bh, uptodate);
 #if DEBUG_ASYNC
+    //printk("debug level: %d\n", debug_level);
     if (debug_level == 1) printk("ER:%04x;", req);
-    if (debug_level > 3 {
+    if (debug_level > 3 ) {
 	__far unsigned int *content;
 	content = _MK_FP(req->rq_seg, (unsigned int) (req->rq_buffer));
 	printk("ER%d|%04x|%04x|%04x;", uptodate, req, req->rq_dev, *content);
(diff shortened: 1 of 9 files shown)


ghaerr:
> This means - for the directfd driver - that the cache and the bounce buffer are permanently separated 

Earlier, you had mentioned that it was probably OK keep the DF bounce buffer and track cache overlapped and invalidate the cache when an XMS/DMA buffer was needed, rather than permanently allocate another 1K for the separate bounce buffer.
Did that assumption change, or is there another reason the DF bounce and cache are permanently separated?

Mellvik:
It's a result of our thorough and very useful discussions. Simplicity won. The ability to eliminate the caching code entirely when not needed wasn't in the plan, but was a welcome benefit. It has been an interesting journey!

PR #104, opened 2024-11-19
[kernel] directhd: Fix buffer allocation bug - and QEMU anomaly

Mellvik:
Fixes garbled boot messages when XMS buffers active

The garbled boot messages  problem described in #xx turned out to be a bug in the directhd driver. A while back the driver was changed to allocate a bounce buffer from the heap instead of statically allocating it. [The 512 byte IDE ID buffer was always (heap)allocated and released separately].  The heap-allocation  was (logically) done in the device open function and freed in _release, which ignored the fact that the partition detection code runs long before any regular device open, reading the first block of any drive. So the first block of the first drive was essentially overwriting the bottom 1k of memory.

An 'academically nice' fix would be to add an early call to device-open to fix the problem, but a more compact solution - the one chosen - is to simply heap-allocate a e bounce buffer at init-time and never release it. Less code - and it lends itself to make use of the DMASEG bounce buffer 'set aside' (conditionally) for XD and LANCE drivers. Given that the XD and IDE drivers are (currently) mutually exclusive and the LANCE driver not yet available, that would be tempting. 

The question that needs some pondering is whether the XD and IDE drivers should be permanently mutually exclusive: XD is XT class only, IDE runs on XT as XT-IDE and can possibly do so alongside an XD controller. So the conclusion for now is to leave the XD/LANCE bounce buffer alone.

One more thing - @ghaerr, you're going to like this one. 

As I was wrapping up the testing of the above fix (on QEMU), I enabled XT-IDE support in the directhd driver - and suddenly it stopped working. No hard drives found and hung when trying to mount root. WTF? The XT-IDE has been running well for months, and while the code is unused and thus not required on an AT (or later) system, the driver should work on any system with XT-IDE support compiled in. [The previous statement is not entirely true. There are plenty XT-IDE controllers in use on AT and later systems out there, for the simple reason that they are 'current' - in production, with new BIOS software. Even PCI versions IIRC.]

Anyway, some digging revealed that the `STATUS(port)` macro, which is different when XT-IDE is compiled in to accommodate the different register addressing schemes on XT-IDE variants, returned a different value than the 'normal' version, making the driver conclude that there is no controller - and the system startup fail (bot device not found).
```
#ifdef CONFIG_IDE_XT
static unsigned char STATUS_REG[3] = { 0x7, 0xE, 0xE};
static unsigned char ERROR_REG[3]  = { 0x1, 0x2, 0x8};
static unsigned char SECTOR_REG[3] = { 0x3, 0x6, 0xA};

#define STATUS(p) inb_p((p) + STATUS_REG[cf_shift])
#define ERROR(p)  inb_p((p) + ERROR_REG[cf_shift])
#define SECTOR(p) inb_p((p) + SECTOR_REG[cf_shift])
/* cf_shift indicates the type of XT_IDE controller, 0 is normal IDE */

#else           /* No IDE_XT */

#define STATUS(p) inb_p((p) + ATA_STATUS)
#define ERROR(p)  inb_p((p) + ATA_ERROR)
#define SECTOR(p) inb_p((p) + ATA_SECTOR)

#endif          /* CONFIG_IDE_XT */
```
Manually replacing this particular occurrence (in `reset_controller()`) of the macro with the 'normal' (non-XT-IDE) variant, and the problem went away. Again - WTF? The apparently non-functioning macro works well in other contexts, the probe-case (in `reset_controller()`) was the only occurrence in which it failed (IOW, using the 'normal' macro at this particular point in the code and the system would boot and work fine). Looked like memory or stack pollution to me.

Except it turned out that everything worked fine on real hardware. Wow, another QEMU bug? As I would (eventually) find out, yes - sort of.  What happens is this:
- the  probe code resets the controller ('externally' via the 'control' register)
- The reset has a well-defined effect on the controller, causing a sequence of events leading to a predictable starting point including initial values for all registers.
- The first thing to happen when the reset-line is activated is that the controller's BUSY status bit is set, which is what the probe routine is testing (after verifying that the status value isn't 0xff which indicates nothing there).
- This works fine on real hardware and on QEMU with one version of the macro but not the other. 

I accidentally put both the non-working and working variant of the macro in the code (in that order) and compiled. Voila - now the working variant stopped working. That solved the mystery.

What's happening is (again) that QEMU is too fast - compared to reality. QEMU sets the BUSY bit very briefly, then clears it and lets the status reg show other (presumably valid) status bits. The 'normal IDE' version of the macro is fast enough to catch the BUSY bit while on, the second is not - having to index into an array to get the offset before reading the register. [BTW, regardless of speed, QEMU's behaviour is incorrect per the ATA spec.]

IOW, testing for the BUSY bit in the status register is not a reliable probe - on QEMU. It's as simpe as that. Not a problem. We already know how real hardware behaves, now we know how QEMU behaves, we can handle both. Problem solved. Rabbit caught and caged.

### DIFF
--- a/tlvc/arch/i86/drivers/block/directhd.c
+++ b/tlvc/arch/i86/drivers/block/directhd.c
@@ -15,6 +15,7 @@
  * - test with 2 controllers, 4 drives
  * - add LBA support (will LBA work on a CHS initialized drive - or vise versa?)
  * - recognize the presence of solid state devices (to remove request sorting)
+ * - start_sect is sometimes used as signed long (as in '-1'), not good
  */
 /*
  * A note about IDE/ATA:
@@ -67,14 +68,22 @@ static unsigned char SECTOR_REG[3] = { 0x3, 0x6, 0xA};
 #define STATUS(p) inb_p((p) + STATUS_REG[cf_shift])
 #define ERROR(p)  inb_p((p) + ERROR_REG[cf_shift])
 #define SECTOR(p) inb_p((p) + SECTOR_REG[cf_shift])
+
 #else 		/* No IDE_XT */
+
 #define STATUS(p) inb_p((p) + ATA_STATUS)
 #define ERROR(p)  inb_p((p) + ATA_ERROR)
 #define SECTOR(p) inb_p((p) + ATA_SECTOR)
+
 #endif		/* CONFIG_IDE_XT */
 
 #define WAITING(p) ((STATUS(p) & BUSY_STAT) == BUSY_STAT)
 #define DRQ_WAIT(p) (STATUS(p) & DRQ_STAT) /* set when ready to transfer */
+
+#define DEBUG_DIRECTHD 0
+#if DEBUG_DIRECTHD
+static void dump_ide(word_t *, int);
+#endif
 
 /* #define USE_ASM */
 /* use asm insw/outsw instead of C version */
@@ -90,8 +99,6 @@ __asm__("cld;rep;outsw"::"d" (port),"S" (buf),"c" (nr))
 //#define USE_INTERRUPTS		/* EXPERIMENTAL - test interrupts */
 				/* cannot work with XT/IDE, XT/CF */
 #define OLD_IDE_DELAY 1000	/* delay required for old drives, system dependent */
-//#define DEBUG
-
 //#define USE_LOCALBUF		/* For debugging: use local bounce buffer instead of */
 				/* direct buffer io via far pointers.
 				 * When XMS buffers are active, the same 1k buffer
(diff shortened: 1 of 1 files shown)


ghaerr:
> the partition detection code runs long before any regular device open

Wow! I never realized that. Definitely bad behavior. I'll look into what it'd take to call the device open routine, as the partition read code is running at the very end of the kernel initialization. I would suppose the biggest issue would be potential conflicts with mount_root, if we try to not open/close/open the same driver on boot for heap and other efficiency reasons.

Mellvik:
Agreed. The again I decided to take the easy way out, thinking that we're accessing (actually reading) the device even before that, when reading the ID 'sector' off the drive, which I believe was a (hidden) physical sector on the earliest drives, later some EAROM variant. The difference is that in the sector 0 case, we actually have a choice. IOW it may not be worth the complications the make it 'right' ...

PR #107, opened 2024-11-23
[kernel] Experimental heap_free valitity checks

Mellvik:
Adds 'canary' based and 'list' based validity checks for `heap_free` for discussion, re. #105.

@ghaerr, to me these two variants do exactly the same, which means the first (list based) alternative is due for deletion. Do you agree?

### DIFF
--- a/tlvc/lib/heap.c
+++ b/tlvc/lib/heap.c
@@ -12,6 +12,9 @@
 #define HEAP_MIN_SIZE (sizeof (heap_s) + 16)
 #define HEAP_SEG_OPT	/* allocate small SEG descriptors from the upper */
 			/* end of the heap to reduce fragmentation */
+#define HEAP_CANARY	0xA5U	/* for header validation */
+//#define HEAPFREE_VALIDATE_1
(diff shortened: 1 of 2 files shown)


ghaerr:
> which means the first (list based) alternative is due for deletion. Do you agree?

Yes!

tlvc/lib/heap.c
```diff
@@ -150,8 +155,34 @@ void heap_free(void *data)
...
+	}
+	if (n == &_heap_all) {
+		printk("WARNING: bogus heap_free");
+		return;
+	}
+#endif
+
+#ifdef  HEAPFREE_VALIDATE_2
+	//printk("Heap canary %x\n", *((char *)data - 1)&0xff);
+	if ((*((char *)data - 1)&0xff) != HEAP_CANARY) {
```
ghaerr, line 180: Can't you just use `if (h->canary != HEAP_CANARY)` here rather than using complicated pointer arithmetic that also assumes canary is the last struct member?

Mellvik: Not my day! Thanks.

PR #108, opened 2024-11-25
[kernel] Fix serious bug in raw IO

Mellvik:
Discovered by @ghaerr in code review, this update fixes a serious flaw in raw block IO which would eventually cause system crash if XMS buffers were active. Thanks @ghaerr.

### DIFF
--- a/tlvc/fs/block_dev.c
+++ b/tlvc/fs/block_dev.c
@@ -213,7 +213,7 @@ static int raw_blk_rw(struct inode *inode, register struct file *filp,
 		}
 	} else {	/* moving full sectors */
 		char *o_data;
-		seg_t o_seg;
+		ramdesc_t o_seg;
 		unsigned char sec_cnt;
 
 		chars = (count & 0xffff); /* try to transfer the whole thing -
--- a/tlvc/fs/block_dev.c
+++ b/tlvc/fs/block_dev.c
@@ -222,7 +222,7 @@ static int raw_blk_rw(struct inode *inode, register struct file *filp,
 
 		ebh->b_blocknr = filp->f_pos >> SECT_SIZE_BITS;
 		o_data = bh->b_data;		/* save the 'original' values */
-		o_seg = (seg_t)ebh->b_L2seg;	/* may be long (xms active) or int */
+		o_seg = ebh->b_L2seg;	/* may be long (xms active) or int */
 		ebh->b_L2seg = (ramdesc_t)current->t_regs.ds;
 		bh->b_data = buf;
 		ebh->b_nr_sectors = (chars >> SECT_SIZE_BITS);
@@ -232,7 +232,7 @@ static int raw_blk_rw(struct inode *inode, register struct file *filp,
 
 	    	ll_rw_blk(wr, bh);
 	    	wait_on_buffer(bh);
-		ebh->b_L2seg = (ramdesc_t)o_seg;		/* restore */
+		ebh->b_L2seg = o_seg;		/* restore */
 		bh->b_data = o_data;
 		if (sec_cnt != ebh->b_nr_sectors) {
 			debug_raw("partial raw IO, requested %d, got %d\n",


tlvc/fs/block_dev.c
```diff
@@ -222,7 +222,7 @@ static int raw_blk_rw(struct inode *inode, register struct file *filp,
 
 		ebh->b_blocknr = filp->f_pos >> SECT_SIZE_BITS;
 		o_data = bh->b_data;		/* save the 'original' values */
-		o_seg = (seg_t)ebh->b_L2seg;	/* may be long (xms active) or int */
+		o_seg = ebh->b_L2seg;	/* may be long (xms active) or int */
 		ebh->b_L2seg = (ramdesc_t)current->t_regs.ds;
```
ghaerr, line 226: Since I'm doing a code review, I'll continue to comment...

Does removing the (ramdesc_t) cast cause a compiler warning? It should not. IMO, explicitly casting here is both not needed and dangerous. For instance, if ramdesc_t were 'signed', (which it happens not to be), then this forced cast would cause a sign-extension when expanding the 16-bit DS value to be stored in L2_seg, and no-one would know the better since this code is far from the ramdesc_t definition. Better to leave it alone and let the compiler figure it out, unless there's a warning being produced.

Mellvik: Thanks - very welcome input.

It's a very subtle point you're making. Removing the cast does not trigger a compiler warning. The cast was added not for the compiler but for the programmer (the would be me) as a reminder. It obviously didn't occur to me that this could possibly be bad, and like you point out, it isn't in this case since ramdesc_t is what it is. It's also easy to see, using your example, how it could turn out really bad.

If I'm reading you correctly, what you're saying is that if the signedness of the two sides is the same, leave it to the compiler to do the right thing. And I'm assuming this goes both ways, u_int -> u_long and vice versa? gcc doesn't complain either way.

Then, with a smile, I guess the lesson is "don't hide typing complacency behind casts".

ghaerr: > what you're saying is that if the signedness of the two sides is the same, leave it to the compiler to do the right thing. And I'm assuming this goes both ways, u_int -> u_long and vice versa? gcc doesn't complain either way.

Yes - but more importantly the C standards have a couple fairly straightforward rules on how expressions are treated when signedness differs. These need to be memorized, as they're the start of a huge mess of more complicated issues with C - see below.

> The cast was added not for the compiler but for the programmer (the would be me) as a reminder.

The system I use for reminding myself of various programming issues is not to add casts which have the possible undesirable effect of producing bad code and/or being permanently forgotten, but to add a keyword in a comment so the issue can be later easily found using a recursive grep. The three keywords are NOTE, TODO and FIXME, in increasing order of severity.

> the lesson is "don't hide typing complacency behind casts".

That's true, but the bigger issue with casts is that we're dealing with Optimizing Compilers now. We're no longer in the K&R world of writing a C statement with the expectation that it is duly translated into a single instruction or two, with a cast helping the compilers consideration for a better/proper instruction. Instead, the C language itself is defined by an abstract standard of (in this case) expression evaluation on an abstract machine, with the compiler having to comply to that standard. These are found in the C standards (C89, C99, C11 etc), now all hundreds of pages long.

So now, single statements are never "compiled" directly into code. Instead, the entire C source module (.c file) is converted to an intermediate representation and a host of data and control flow analysis is performed, ultimately allowing for entire swaths of code to be eliminated, combined or modified, far beyond what us old timers think a compiler is doing on a line-by-line basis. A cast is essentially a flag saying "don't follow the abstract machine standard here" and can have a number of far-reaching effects, as well as almost always being non-portable, causing future maintenance issues.

If you really want to get into it, start reading about UB (undefined behavior), or, for getting closer to how things work now, [Integer Overflow Handling is Undefined Behavior](http://www.pixelbeat.org/programming/gcc/integer_overflow.html) or [How undefined signed overflow enables optimizations in GCC](http://kristerw.blogspot.com/2016/02/how-undefined-signed-overflow-enables.html). This is just the tip of the iceberg in how modern day compilers operate, including ia16-elf-gcc.

The "signed vs unsigned" type conversions are pretty straightforward and should considered whenever any comparison or assignment expression is written in C. For hardcore C fanatics, read about [Implicit type promotion rules](https://stackoverflow.com/questions/46073295/implicit-type-promotion-rules) - extremely complicated stuff, especially for those of us trying to write portable code across machines where sizeof(int) differs.

Lesson - "Don't use casts unless you absolutely have to" - let the optimizing compiler do its job, unless you're checking the code produced with and without it!

Mellvik: Wow, @ghaerr - thanks for the explanation and not the least the links. I headed right over to the stackoverflow posts and realized I've been living under a rock for quite some time, at least when it comes to casts and type promotions. This is sort of scary (from the stackoverflow article):

>> "Unfortunately, the rules for implicit type promotion cause much more harm than good, to the point where they might be one of the biggest flaws in the C language. These rules are often not even known by the average C programmer and therefore cause all manner of very subtle bugs."

Holy macro. It occurs to me that even C has become a Committee language over the years. Lots of good intentions, lots of complexity. Instead of even trying to understand (and remember) it all, I'll take your advice: 

> Lesson - "Don't use casts unless you absolutely have to" - let the optimizing compiler do its job, unless you're checking the code produced with and without it!

... and my new practical rule is to not trust casts to do what I think they will be doing. The promotion of unsigneds to signeds for low rank integer types scares me. Agree, `objdump` is much better than 'ok, it works now, let's leave it alone'.

ghaerr: > I've been living under a rock for quite some time, at least when it comes to casts and type promotions. 

Yes, but I'm glad to hear that the net effect will be less explicit casting unless the code doesn't actually work. Which brings up the need for testing all code paths in any new code - but that's for another day.

> The promotion of unsigneds to signeds for low rank integer types scares me.

The good news is that, for ia16-elf-gcc and ELKS/TLVC, sizeof(unsigned) == sizeof(int) so unsigned int -> int promotions won't occur, which leaves only promotions within arithmetic on 'char' variables to worry about. Concerning that, I only use 'char/unsigned char' variables for bools and arrays, but never for function argument declarations and only when I must for passing to functions.

Don't forget to fix line 226! :)

PR #111, opened 2024-12-07
[kernel] Add support for 1k size bootopts

Mellvik:
Leftover from #106

### DIFF
--- a/tlvc/arch/i86/boot/setup.S
+++ b/tlvc/arch/i86/boot/setup.S
@@ -946,7 +946,10 @@ putc:	push %ax
 
 #ifdef CONFIG_BOOTOPTS
 //
-// load /bootopts for FAT filesystem boot
+// load /bootopts file for FAT filesystem boot
+//      If size > 512 bytes, both sectors must be contiguous.
+//      This is currently guaranteed by using a 1K /bootopts in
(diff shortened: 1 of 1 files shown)


tlvc/arch/i86/boot/setup.S
```diff
@@ -1013,7 +1016,7 @@ pc98_int1b:
 	mov	$DEF_OPTSEG,%ax	// ES:BX = DEF_OPTSEG:0
 	mov	%ax,%es
 	xor	%bx,%bx
-	mov	$0x0201,%ax	// BIOS read disk, 1 sector
+	mov	$0x0202,%ax	// BIOS read disk, 1 sector
```
ghaerr, line 1019: Not a big deal, but the comment here still reads 1 sector, when we're now reading 2.

I've purposely not retabbed this file (setup.S) so as to keep it easier in sync with TLVC via diff - it should be identical. Was your last diff showing other differences this PR didn't update, and/or are other TLVC changes required?

Mellvik: Thanks, I missed that one. And yes, it's a good idea to keep these in complete sync. I just checked the diff - there is no difference except in the comments. Plus there is the `hello_mess` - which could be fixed using `UTS_SYSNAME` from `config.h`.

PR #109, opened 2024-11-26
[cmd] New cut of meminfo - display memory layout 

Mellvik:
This update to `meminfo` adds a new option, `-M` for 'memory layout', and a number of small adjustments. The memory layout is the beginning of a more comprehensive display visualizing the usage of the entire RAM, from 0 to 640k or whatever size the main memory has. Possibly including UMBs which are not implemented in TLVC yet. For now the display looks like this:
```
tlvc15# meminfo -M
Heap:
 Block 1: 6BD2 - FFFF (37933 bytes)
 Block 2: 3E8C - 428C ( 1024 bytes)
 Total:		       38957 bytes
Main memory:
 Area 1: 24E6 - 9FC0 (503200 bytes)
 Area 2: 126F - 14E6 ( 10096 bytes)
 Area 3: 0050 - 00B0 (  1536 bytes)
 Total:		      514832 bytes
Memory:  502KB total,  214KB used,  288KB free

tlvc15#
```

There has also been some adjustments to the memory usage (`-m`) display:
```
tlvc15# meminfo -m
	 SEG   TYPE    SIZE  CNT  NAME
Area 1	24E6   BUF     4000    1  
	25E0   DSEG    5872    1  /bin/init 
	274F   CSEG    4096    1  telnetd 
	284F   CSEG   45936    1  -/bin/sh 
	3386   DSEG   25216    1  -/bin/sh 
	39AE   free   52208    0  
	466D   DSEG    5680    1  telnetd 
	47D0   CSEG   13936    1  ftpd 
	4B37   CSEG    4064    1  meminfo 
	4C35   free    1536    0  
	4C95   CSEG   18464    1  ktcp 
	5117   free   17616    0  
	5564   CSEG    5920    1  /bin/getty 
	56D6   free   16336    0  
	5AD3   DSEG   14656    1  ftpd 
	5E67   DSEG    3136    1  /bin/getty 
	5F2B   free    1168    0  
	5F74   DSEG   58832    1  ktcp 
	6DD1   free  204528    0  
Area 2	126F   CSEG    7056    1  /bin/init 
	1428   DSEG    2544    1  meminfo 
	14C7   free     496    0  
Area 3	0050   free    1536    0  
Memory:  502KB total,  214KB used,  288KB free
```

During this work it became apparent that the after-boot addition of released memory to the heap, block 2 in the display above changes heap usage in interesting and positive ways. Small allocations, such as seg descriptors, gravitate naturally to the added block because of its small size, as obvious from this display:
```
[...]

PR #114, opened 2024-12-09
[kernel] Rearranged low memory enabling release of fd cache space

Mellvik:
Discussed in https://github.com/Mellvik/TLVC/pull/109#issuecomment-2526204965

TLVC low memory has been rearranged to enable the release of an allocated floppy cache if unused (i.e. `fdcache=` in `bootopts` set to 0 or unset.  This reduces the 'penalty' of allocating a nominal floppy sector cache in the kernel configuration and then not taking advantage of that space. By placing the cache immediately after the `bootopts` buffer and the system setup segment, which are always released, the entire block can be release as one, which means 6.5k bytes if 5k has been allocated to the FD cache.

Also fixes a bug in `init/main.c` which would always disable the floppy cache if no `bootopts` was found or configured.

`meminfo` has been updated to work with these changes. Tested with and without all combinations of with/without ftext kernel, bootopts. 

This implementation does not enable the release of a partly unused floppy sector cache. If (say) 7k has been configured and only 5k used (by `fdcache=` in `bootopts`), the unused block will not be released. It is not clear that the code needed to take advantage of such an esoteric situation would be worth it.

### DIFF
--- a/tlvc/arch/i86/drivers/block/bioshd.c
+++ b/tlvc/arch/i86/drivers/block/bioshd.c
@@ -969,7 +969,7 @@ static int do_bios_readwrite(struct drive_infot *drivep, sector_t start, unsigne
 	if (error) return 0;
 	if (usedmaseg) {
 		if (cmd == READ)	/* copy DMASEG up to xms*/
-			xms_fmemcpyw(buf, seg, 0, FD_BOUNCE_SEG, this_pass*(drivep->sector_size >> 1));
+			xms_fmemcpyw(buf, seg, 0, FD_BOUNCESEG, this_pass*(drivep->sector_size >> 1));
 		set_cache_invalid();
 	}
 	return this_pass;
--- a/tlvc/arch/i86/drivers/block/directfd.c
+++ b/tlvc/arch/i86/drivers/block/directfd.c
@@ -360,7 +360,7 @@ static unsigned char use_cache;
 static int cache_drive = -1;
 static int cache_start;		/* first cached sector */
 static int cache_len;		/* valid length of cache in sectors */
-static int cache_size;		/* # of sectors - min=0 (off), max=FD_CACHE_SEGSZ>>9 */
+static int cache_size;		/* # of sectors - min=0 (off), max=FD_CACHESEGSZ>>9 */
 //static unsigned int cache_hits;	/* For stats */
 #endif
 
@@ -682,7 +682,7 @@ static void setup_DMA(void)
     count = nr_sectors<<9;
     if (use_bounce || (physaddr + (unsigned int)count) < physaddr) { /* 64k phys wrap ? */
 	use_bounce++;
-	dma_addr = _MK_LINADDR(FD_BOUNCE_SEG, 0);
+	dma_addr = _MK_LINADDR(FD_BOUNCESEG, 0);
 	if (raw) {	/* The application buffer spans a 64k boundary, split it into
 			 * 2 or three parts, using the first k of the sector cache
 			 * as a bounce buffer */
@@ -699,7 +699,7 @@ static void setup_DMA(void)
     } else {
 #if CONFIG_FLOPPY_CACHE
 	if (use_cache) 
-	    dma_addr = _MK_LINADDR(FD_CACHE_SEG, 0);
+	    dma_addr = _MK_LINADDR(FD_CACHESEG, 0);
 #endif
     }
     if (raw) {	/* ensure raw access doesn't span cylinders */
@@ -715,7 +715,7 @@ static void setup_DMA(void)
 
 #ifdef INCLUDE_FD_FORMATTING
     if (command == FD_FORMAT) {
-	dma_addr = _MK_LINADDR(FD_BOUNCE, 0);
+	dma_addr = _MK_LINADDR(FD_BOUNCESEG, 0);
 	count = floppy->sect << 2;	/* formatting data, 4 bytes per sector */
     }
 #endif
@@ -728,7 +728,7 @@ static void setup_DMA(void)
     } else
 #endif
     if (use_bounce && command == FD_WRITE)
-	xms_fmemcpyw(0, FD_BOUNCE_SEG, req->rq_buffer, req->rq_seg, BLOCK_SIZE/2);
+	xms_fmemcpyw(0, FD_BOUNCESEG, req->rq_buffer, req->rq_seg, BLOCK_SIZE/2);
 
     DEBUG("%d/%lx/%x;", count, dma_addr, physaddr);
     clr_irq();
@@ -999,16 +999,16 @@ static void rw_interrupt(void)
     }
     if (raw) CURRENT->rq_nr_sectors = nr_sectors;
     if (use_bounce && command == FD_READ) {
-	xms_fmemcpyw(req->rq_buffer, req->rq_seg, NULL, FD_BOUNCE_SEG, nr_sectors << (9-1));
+	xms_fmemcpyw(req->rq_buffer, req->rq_seg, NULL, FD_BOUNCESEG, nr_sectors << (9-1));
 									       /* words ^ */
     } else {
 #if CONFIG_FLOPPY_CACHE
 	if (use_cache) {
 	    cache_drive = current_drive;
 	    cache_start = CURRENT->rq_blocknr;
-	    DEBUG("rd:%04x:0->%04lx:%04x;", FD_CACHE_SEG,
+	    DEBUG("rd:%04x:0->%04lx:%04x;", FD_CACHESEG,
 		(unsigned long)req->rq_seg, req->rq_buffer);
-	    xms_fmemcpyw(req->rq_buffer, req->rq_seg, 0, FD_CACHE_SEG, BLOCK_SIZE/2);
+	    xms_fmemcpyw(req->rq_buffer, req->rq_seg, 0, FD_CACHESEG, BLOCK_SIZE/2);
 	}
 #endif
     }
@@ -1506,11 +1506,11 @@ static void redo_fd_request(void)
 
 	if (command == FD_READ) {	/* requested data is in cache */
 //	    cache_hits++;
-	    xms_fmemcpyw(req->rq_buffer, req->rq_seg, buf_ptr, FD_CACHE_SEG, BLOCK_SIZE/2);
+	    xms_fmemcpyw(req->rq_buffer, req->rq_seg, buf_ptr, FD_CACHESEG, BLOCK_SIZE/2);
 	    request_done(1);
 	    goto repeat;
     	} else if (command == FD_WRITE)	/* update track cache */
-	    xms_fmemcpyw(buf_ptr, FD_CACHE_SEG, req->rq_buffer, req->rq_seg, BLOCK_SIZE/2);
+	    xms_fmemcpyw(buf_ptr, FD_CACHESEG, req->rq_buffer, req->rq_seg, BLOCK_SIZE/2);
     } 
 #endif
 
@@ -1846,10 +1846,10 @@ void INITPROC floppy_init(void)
 	cache_size = fdcache<<1;	/* cache size is sectors, fdcache is k bytes */
     else if (arch_cpu == 7)
 	cache_size = 0; 		/* sector cache is slowing down fast systems */
-    else cache_size = FD_CACHE_SEGSZ>>9;	/* use menuconfig value */
+    else cache_size = FD_CACHESEGSZ>>9;	/* use menuconfig value */
 
-    if (cache_size > (FD_CACHE_SEGSZ>>9)) cache_size = FD_CACHE_SEGSZ>>9;
-    printk("Floppy cache %dk, available %dk\n", cache_size>>1, FD_CACHE_SEGSZ>>10);
+    if (cache_size > (FD_CACHESEGSZ>>9)) cache_size = FD_CACHESEGSZ>>9;
+    printk("Floppy cache %dk, available %dk\n", cache_size>>1, FD_CACHESEGSZ>>10);
 	
 #endif
 }
--- a/tlvc/arch/i86/drivers/block/genhd.c
+++ b/tlvc/arch/i86/drivers/block/genhd.c
@@ -362,7 +362,7 @@ void INITPROC setup_dev(register struct gendisk *dev)
 #endif
 	memset((void *)dev->part, 0, sizeof(struct hd_struct)*dev->max_nr*dev->max_p);
 	dev->init(dev);
-	printk("setup_dev %04x/%d(%d)\n", dev, dev->major,sizeof(struct hd_struct)*dev->max_nr*dev->max_p);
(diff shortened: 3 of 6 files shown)


Mellvik:
@ghaerr, 
while writing a wiki-piece explaining  TLVC memory, it occurred to me that low memory can be simplified even more. The bootopts buffer and the floppy bounce buffer are really doing the same thing -at different stages. So it's time to rearrange low memory again :-) What I'm thinking is to swap setup data and the  bootopts buffer and equate the bootopts buffer and the floppy bounce buffer. IOW when done with the options buffer, that 1k block becomes the FD_BOUNCE buffer.

This will shrink the entire low memory block by 1k while the setup data segment may be released along with the floppy cache if available and unused. Not only that: Since we're done with setup data at the time when the FD cache becomes active, there is no reason for them not to overlap. If there is no FD_CACHE, setup_data may be released as usable memory like now, unless it is deemed too small a segment to bother.

This will also simplify `meminfo` somewhat by eliminating a variable. What do you think?

The new low memory layout would be like this:
```
 | kernel text     |
 +-----------------+ XD_BOUNCESEG = (FD_CACHESEG + (CONFIG_FLOPPY_CACHE<<10))
 | Floppy cache    |  Optional, max 7k (6.5k+setup_data)
 | setup data      |  512B - mandatory, releasable after boot
 +-----------------+ 0x80 REL_INITSEG FD_CACHESEG
 | bootopts buffer | 
 | Floppy bounce   |  1k
 +-----------------+ 0x40  FD_BOUNCESEG DEF_OPTSEG
 | IRQ vectors/BDA |
 +-----------------+ 0x000
```

ghaerr:
@Mellvik, this seems like potentially a great idea. Since DEF_OPTSEG is copied into the kernel data segment very early in kernel init, I can't think of how any disk I/O overwriting it might be a concern. (genhd.c uses it for reading MBRs, right?)

The same idea for REL_INITSEG, as long as the disk I/O buffer(s) it might be shared with don't ever get used during startup.

It might be good to keep REL_INITSEG, or even have REL_INITSEG and DEF_OPTSEG share a portion of each other, however,  and ensure a portion always gets released, as that creates a small arena that is used to keep segment allocation descriptors out of the main memory area, right? Your previous research had shown that having this separate arena allowed for removing the special high-segment allocation handling which was added previously. I'm not sure how big this area needs to be - perhaps that size could determine how much of DEF_OPTSEG and REL_INITSEG is shared, and then released?

These are my quick thoughts, more coming after a little more time...

Mellvik:
Thanks @ghaerr - the memory rearrangement project - if we can all it that - was closed, now reopened!

> @Mellvik, this seems like potentially a great idea. Since DEF_OPTSEG is copied into the kernel data segment very early in kernel init, I can't think of how any disk I/O overwriting it might be a concern. (genhd.c uses it for reading MBRs, right?)

Yes, the sequence of things seems to allow for (even invite) this. The MBR is a different story - `genhd.c` is calling `bread` on the unopened device. The more I think about it, the more I agree with you: It's ugly, I may have to cave in and add open and close in `genhd.c`.  Did you do anything about this in ELKS yet?

> 
> The same idea for REL_INITSEG, as long as the disk I/O buffer(s) it might be shared with don't ever get used during startup.
> 
> It might be good to keep REL_INITSEG, or even have REL_INITSEG and DEF_OPTSEG share a portion of each other, however, and ensure a portion always gets released, as that creates a small arena that is used to keep segment allocation descriptors out of the main memory area, right? Your previous research had shown that having this separate arena allowed for removing the special high-segment allocation handling which was added previously. I'm not sure how big this area needs to be - perhaps that size could determine how much of DEF_OPTSEG and REL_INITSEG is shared, and then released?

That was in fact the heap, where releasing the bootopts buffer created a bucket into which small allocations automatically ended up. What we have here is - when there is no FDcache - a superfluous 512b segment which is pretty useless. I've studied 10s and 10s of outputs from `meminfo` from the scroll buffer, and there is never a memory allocation this small. Programmatically the simplest is to just let it be, make the 512b available and ignore the fact that they're never used. That increases the search marginally when allocating new memory segments, probably negligible.
Also, keeping it allows for the bold developer to experiment a bit with this area, vary its size and see what happens. Like - configure 1 or 2k FDcache in CONFIG, then set `fdcache=0`  and you have a 2.5k separate memory area and can use `meminfo` to see how the system is using it. Useful? I don't know - but possible.

I'm posting the changes later today when `meminfo` has been updated.
> 
> These are my quick thoughts, more coming after a little more time...

ghaerr:
> having this separate arena allowed for removing the special high-segment allocation handling which was added previously. 
> That was in fact the heap, where releasing the bootopts buffer created a bucket into which small allocations automatically ended up.

Oops - thanks, I got the two heaps confused there for a moment.

> Programmatically the simplest is to just let it be, make the 512b available and ignore the fact that they're never used.

Yes, probably, at least for now. Actually, though, REL_INITSEG doesn't have to be 512 bytes - it's origination was as a boot sector tacked on to the front of the kernel load image in early Linux. Neither ELKS nor TLVC use that "dummy boot" sector anymore, and REL_INITSEG isn't a copy of it. While for various reasons we'll probably want to keep the dummy boot sector, REL_INITSEG itself could be whittled down to probably around 32 bytes (~2 paragraphs). Doing so right now is a bit of a pain though, as the addresses used (see include/linuxmt/boot.h) are all in the higher portion of the 512 bytes; they'd all have to be moved down. Then some small changes in setup.S and possibly the boot code. 

So for now, lets leave REL_INITSEG as is, 512 bytes, and I like your idea of adding it to the floppy cache, possibly to be released along with the cache. I've made a note of thinking about reducing the size of REL_INITSEG, which if cut back to 32 bytes, could be forgotten about entirely with regards to discarding.

> The MBR is a different story - genhd.c is calling bread on the unopened device.
> Did you do anything about this in ELKS yet?

No, but I guess I didn't fully realize that bread itself is being used for the MBR reads. That opens (no pun intended) the larger question - if this works with no block device open at boot time, then why does the primary boot device need an open routine at all?? Lets talk about that, as if the block driver did need an open routine, it would seem that the MBR reading would fail for those block devices?

Mellvik:
> > having this separate arena allowed for removing the special high-segment allocation handling which was added previously.
> > That was in fact the heap, where releasing the bootopts buffer created a bucket into which small allocations automatically ended up.
> 
> Oops - thanks, I got the two heaps confused there for a moment.

Welcome to the club!!

> 
> > Programmatically the simplest is to just let it be, make the 512b available and ignore the fact that they're never used.
> 
> Yes, probably, at least for now. Actually, though, REL_INITSEG doesn't have to be 512 bytes - its origination was as a boot sector tacked on to the front of the kernel load image in early Linux. Neither ELKS nor TLVC use that "dummy boot" sector anymore, and REL_INITSEG isn't a copy of it. While for various reasons we'll probably want to keep the dummy boot sector, REL_INITSEG itself could be whittled down to probably around 32 bytes (~2 paragraphs). Doing so right now is a bit of a pain though, as the addresses used (see include/linuxmt/boot.h) are all in the higher portion of the 512 bytes; they'd all have to be moved down. Then some small changes in setup.S and possibly the boot code.
> 
> So for now, lets leave REL_INITSEG as is, 512 bytes, and I like your idea of adding it to the floppy cache, possibly to be released along with the cache. I've made a note of thinking about reducing the size of REL_INITSEG, which if cut back to 32 bytes, could be forgotten about entirely with regards to discarding.

OK, let's leave it then. But you've created an itch... like: I don't know much about the BDA and I suspect that most or all of it may be irrelevant to TLVC with its very limited BIOS usage. Maybe there's 32 bytes available that can allow REL_INITSEG to be ditched altogether?

BTW - this all works fine in the PR pushed earlier today. What I'm thinking right now is that the new low-mem structure makes it just too easy to release unused portions of the FDcache to not do it. Like, say 7k is configured for FDcache, 5k is used per `fdcache=`, the remaining 2k could easily be released at startup. Stay tuned.

> 
> > The MBR is a different story - genhd.c is calling bread on the unopened device.
> > Did you do anything about this in ELKS yet?
> 
> No, but I guess I didn't fully realize that bread itself is being used for the MBR reads. That opens (no pun intended) the larger question - if this works with no block device open at boot time, then why does the primary boot device need an open routine at all?? Let's talk about that, as if the block driver did need an open routine, it would seem that the MBR reading would fail for those block devices?

Well, that's how I found out. I forget which device, probably the XT_IDE part of `directhd`, but some code in open() was required to get bread to work (surprise...), and the rabbit was chased down to `genhd.c`. I guess the argument goes like this: Any block device may be a boot device, and open does set counters and other important state variables, so it's needed OK. A different question is whether supporting singular, 'unopened' `bread`s like this should be a requirement (which it is now), which in turn would affect what is put into open() and what goes into _init() instead.

I initially argued that adding open/release is just more code, I'm a little more nuanced now - acknowledging that `genhd` is INITPROC anyway.

Mellvik:
> > I don't know much about the BDA and I suspect that most or all of it may be irrelevant to TLVC with its very limited BIOS usage. Maybe there's 32 bytes available that can allow REL_INITSEG to be ditched altogether?
> 
> We're definitely asking for trouble using the BDA area for our own purposes - there's just too many BIOSes out there to make an assumption about free BDA area, and IIRC last I looked it was (almost) full anyways.
> 
> I have been trying to think of where REL_INITSEG could go besides low memory: high memory would be great for it, in fact it's already moved up there just after boot, but then so are some other things during the very complicated boot process, so I concluded that high memory isn't a great idea. Then perhaps at 128k or 256k would work, but this involves making more assumptions about the size of memory TLVC/ELKS runs on, so probably not a great idea. In front of or behind the kernel data segment came to mind, but that's a non-fixed location. So I finally thought the best solution is to make it ~32 bytes, allocate it in low memory, and leave it there. (I'll look more at squeezing REL_INITSEG smaller, but I'm knee deep in some other stuff at the moment, and there are potential boot issues that I would want to make sure are OK with it not being 512 bytes anymore).

Agreed, let's put this one on the back burner. For now.

> 
> > probably the XT_IDE part of directhd, but some code in open() was required to get bread to work
> 
> In that case, it's a bug that genhd does I/O without an open, period. I'm pretty sure the whole reason it worked in the first place was because ELKS used BIOS I/O which doesn't need opening. Trying to finagle various HD drivers to work without an open or moving init code around is not a good answer here, really. Genhd should open the device.

Good - we're on the same page. I'll put it in shortly.

> 
> The complications here are which device(s) genhd should open - does it open all HD devices that are returned from the BIOS as being available, or just the boot HD device, or which devices exactly? TLVC operates without BIOS info, but ELKS uses bioshd_gethdinfo() for that.

You're probably right about the history behind it. And I was either lucky or smart when I started to do non-BIOS HD, and it just worked. Until it didn't. `genhd` reads the partition table, that's it - applies to any HD - SSD, XD (MFM), Directhd, ...

> 
> We should probably open another discussion for genhd issues, since it might be complicated, and won't be immediately changing. This may also affect early kernel heap fragmentation as well.

Good point. Let's get back to this after I fix the actual problem, then we can see how it works out in practice.

> 
> > What I'm thinking right now is that the new low-mem structure makes it just too easy to release unused portions of the FDcache to not do it. Like, say 7k is configured for FDcache, 5k is used per fdcache=, the remaining 2k could easily be released at startup.
> 
> Agreed - the FD cache should be released if not used, especially since this might be often on 386+ systems, right? Better get this figured out correctly now while your head is into it!

That's right, and it's working as we speak. Simply one more statement in init/main.c. `meminfo` 'adjustments' - well, that's different story :-( .

ghaerr:
>  meminfo 'adjustments' - well, that's different story :-( .

Yes - meminfo is already becoming a maintenance burden. I don't think there's anything wrong at all with it's using config.h - it just shouldn't make assumptions about the values of the defines found in config.h! If that could be made to work, (using a for(;;) loop testing for (if seg == DMASEG) etc or something like that) then meminfo maintenance should go away for the most part, right?

I still plan on pulling meminfo.c over, and it's highly likely that ELKS and TLVC config.h will always differ, but our overall system organization is very similar, which is good, for both our efforts as well as making meminfo maintainable, it seems to me.

Mellvik:
Yes, I need to take a close look at that. Again, this is not a place where we would see frequent changes but it would be good to have reasonable (??) portability.

Anyway, this is where we are at - some more testing needed for the more odd cases:

```
tlvc15# meminfo -M
Kernel heap (DS-base 159A):
	   SEG   OFFS   SIZE
 Block 1: 1C2E   6942  38589 bytes
 Block 2: 195E   3C4E   1024 bytes
 Kernel heap total:    39613 bytes

Main memory:
 Arena 1: 259A - 9FC0 (500320 bytes, 488K)
 Arena 2: 132F - 159A (  9904 bytes, 9.6K)
 Arena 3: 01D0 - 0250 (  2048 bytes, 2.0K)

9FC0:0 +--------------------+ Top of conv. memory
       |                    |
       |                    |
       |                    |
       | Main memory        | 488K Arena 1
       |                    |
       |                    |
       |                    |
259A:0 +--------------------+ Kernel DS end
       |                    |
       | Main kernel heap   | 37K 
       |                    |
 +6942 +- - - - - - - - - - + 
       |                    |
       | Kernel data (bss)  | 10K 
 +404E +- - - - - - - - - - + BSS continues
       | [bootopts buffer]  | 1.0K Heap block 2
 +3C4E +- - - - - - - - - - + 
       |                    | 4.7K 
 +2922 +- - - - - - - - - - + BSS start
       | Kernel data        | 10K 
159A:0 +--------------------+ Kernel DS start
       | [INITPROC code]    | 9.6K Main memory arena 2
132F:0 +- - - - - - - - - - + 
       |                    |
       | Kernel fartext     | 6.7K 
117E:0 +--------------------+ 
       |                    |
       |                    |
       | Kernel text        | 60K 
       |                    |
       |                    |
0250:0 +--------------------+ 
       | [ Unused FDcache ] | 2.0K Main memory arena 3
01D0:0 +--------------------+ 
       | Floppy cache       | 5.0K 
0090:0 +--------------------+ 
       | OPTSEG/FD_BOUNCE   | 1.0K 
0050:0 +--------------------+ 
       | IRQ vectors/BDA    | 1.2K 
0000:0 +--------------------+ 

Memory:  501KB total,  255KB used,  246KB free

tlvc15#
```

Mellvik:
> Yes - meminfo is already becoming a maintenance burden. I don't think there's anything wrong at all with it's using config.h - it just shouldn't make assumptions about the values of the defines found in config.h! If that could be made to work, (using a for(;;) loop testing for (if seg == DMASEG) etc or something like that) then meminfo maintenance should go away for the most part, right?

The challenge with meminfo is sequencing. When the sequence of blocks or their properties change, manual intervention is required. The rest is reasonably automatic. 

Today's changes simplify things a bit but there are still quite a few contingencies involved, bootopts present? Fdcache present, size? All or part of what's available? Released? 

I haven't found a way to turn these contingencies into a loop with reasonable complexity. And I am seriously looking forward to a fresh pair of eyes in. The tool has already contributed a lot to my understanding of not only TLVC memory structure but how the system works overall.

PR #115, opened 2024-12-16
[kernel,cmd] Another simplification of low memory usage

Mellvik:
Another simplification/optimization of low memory usage as discussed in https://github.com/Mellvik/TLVC/pull/114#issuecomment-2543878716

OPTSEG is no longer released after boot, but instead reused as the floppy bounce buffer. Also, if there is a floppy cache in use, the setup_data segment (512b) is reused as part of that. Thus, there will in many cases not be anything to release from low memory after boot and overall memory usage is reduced.

Part of the update is to move processing of the `fdcache=` bootopts parameter from the `directfd` driver to `init/main.c` to ensure that the fdcache may be released as main memory if possible.  

The `meminfo` -M display has been adjusted accordingly, a simplification really because there are fewer variants to consider. Some variants have not been tested yet.

### DIFF
--- a/tlvc/arch/i86/drivers/block/directfd.c
+++ b/tlvc/arch/i86/drivers/block/directfd.c
@@ -1841,14 +1841,7 @@ void INITPROC floppy_init(void)
     config_types();
 
 #if CONFIG_FLOPPY_CACHE
-    /* sector cache setup - /bootopts fdcache= has preference, otherwise autoconfig */
-    if (fdcache != -1)			/* allow fdcache=0 in bootopts */
-	cache_size = fdcache<<1;	/* cache size is sectors, fdcache is k bytes */
-    else if (arch_cpu == 7)
-	cache_size = 0; 		/* sector cache is slowing down fast systems */
-    else cache_size = FD_CACHESEGSZ>>9;	/* use menuconfig value */
-
-    if (cache_size > (FD_CACHESEGSZ>>9)) cache_size = FD_CACHESEGSZ>>9;
+    cache_size = fdcache<<1;	/* cache_size is sectors, fdcache is Kbytes */
     printk("Floppy cache %dk, available %dk\n", cache_size>>1, FD_CACHESEGSZ>>10);
 	
 #endif
--- a/tlvc/include/linuxmt/config.h
+++ b/tlvc/include/linuxmt/config.h
@@ -121,35 +121,31 @@
 
 /* Define segment locations of low memory, must not overlap.
 
-   Reorganized dec 2024 (HS): setup data + bootopts buffer +
-   floppy cache (if present) are now contiguous. If the
-   floppy cache is allocated but not used (bootopts fdcache=0),
-   the entire block may be released and made available as general memory.
-   The structure is: Fixed segs first, then variable/configurable, 
-   then the kernel REL_SYSSEG.
+   Reorganized dec 2024 (HS): The bootopts buffer and the FD bounce buffer
+   are now the same (they're used at different stages), so there is no
+   OPTSEG to release after boot. Also, the setup-data seg (REL_INITSEG) now
+   overlaps with the FD_CACHE if present. If there is a FD_CACHE allocated
+   and not used, it will be released - including the detup-data seg.
+   If there is no FD_CACHE, the setup-data-seg may be released but its small
+   size may make less than useful.
+
    NOTE: The meminfo program makes assumptions about the order of these blocks!
 
  | kernel text     |
  +-----------------+ Kernel CS = REL_SYSSEG = (XD_BOUNCESEG + (XD_BOUNCESEGSZ>>4))
  | Other bounce    |  Optional, 1k
- +-----------------+ XD_BOUNCESEG = (FD_BOUNCESEG + (0x400>>4))
+ | (MFM, Lance)    |
+ +-----------------+ XD_BOUNCESEG = REL_INITSEG + (HAS_FDCACHE ? (FD_CACHESEGSZ>>4):0x10)
+ | Floppy cache    |  Optional, variable, max 6.5 + 0.5k
+ | setup data      |  Mandatory, 512B
+ +-----------------+ 0x90 REL_INITSEG = FD_CACHESEG = DMASEG
  | Floppy bounce   |  1k
- +-----------------+ FD_BOUNCESEG = (FD_CACHESEG + (CONFIG_FLOPPY_CACHE<<10))
- | Floppy cache    |  Optional, max 7k
- +-----------------+ FD_CACHESEG = (DEF_OPTSEG + (OPTSEGSZ>>4)) = DMASEG
- | bootopts buffer |  Optional 1k
- +-----------------+ 0x70 DEF_OPTSEG
- | setup data      |  512B
- +-----------------+ 0x50 REL_INITSEG 
+ | bootopts buffer | 
+ +-----------------+ 0x50 DEF_OPTSEG FD_BOUNCESEG
  | IRQ vectors/BDA |
  +-----------------+ 000
 
 */
-#ifdef CONFIG_BOOTOPTS
-#define OPTSEGSZ        0x400	/* 0x400 bytes boot options*/
-#else
-#define OPTSEGSZ        0
-#endif
 
 #if defined(CONFIG_BLK_DEV_XD) || defined(CONFIG_ETH_LANCE)
 #define XD_BOUNCESEGSZ		0x400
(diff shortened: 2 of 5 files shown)


Mellvik:
the low memory map now looks like this:
```
 | kernel text     |
 | (MFM, Lance)    |
 +-----------------+ XD_BOUNCESEG = REL_INITSEG + (HAS_FDCACHE ? (FD_CACHESEGSZ>>4):0x10)
 | Floppy cache    |  Optional, variable, max 6.5 + 0.5k
 | setup data      |  Mandatory, 512B
 +-----------------+ 0x90 REL_INITSEG = FD_CACHESEG = DMASEG
 | Floppy bounce   |  1k
 | bootopts buffer |
 +-----------------+ 0x50 DEF_OPTSEG FD_BOUNCESEG
 | IRQ vectors/BDA |
 +-----------------+ 000

```

ghaerr:
> the low memory map now looks like this:

Nice visual. In my last comment https://github.com/Mellvik/TLVC/pull/114#issuecomment-2546051445 I didn't fully realize that REL_INITSEG and the floppy cache were overlapped. In most cases, the floppy cache will be configured, right?

If so, then perhaps its not a big deal that REL_INITSEG be squashed to smaller size - just release the portion of floppy cache that is not needed, which could be as small as 512 bytes (REL_INITSEG) when the floppy cache isn't configured at all, which is the rare case, right? In the other cases, the cache is overlapped with REL_INITSEG so REL_INITSEG's size doesn't matter much.

Is this along the lines of what you're now thinking?

Mellvik:
> > the low memory map now looks like this:
> 
> Nice visual. In my last comment [#114 (comment)](https://github.com/Mellvik/TLVC/pull/114#issuecomment-2546051445) I didn't fully realize that REL_INITSEG and the floppy cache were overlapped. In most cases, the floppy cache will be configured, right?

I guess this depends - in particular on my ability to communicate how this works (thus the Wiki I referred to that started this n'th round of reorganization). A kernel with fdcache configured is a more generic kernel and now that the memory assigned is more or less automatically recovered, the waste becomes really small. Automatically as in if the machine is cpu=7, the fdcache will be forced to 0 and the memory freed. This is brute force, I know, but the case where this is not desirable is a case where the developer knows enough to remove the forcing hand'. Also, CONFIG_FLOPPY_CACHE is always set but may be 0. In that case the floppy cache code is not compiled in.

So the thing is, with this change, keeping a FDcache configured becomes cheap enough to let it stay in there, 5k or 7k. Which in practice means that if unused, /bin/init's CSEG or DSEG is going to end up in that arena and very little is wasted.

Personally I tend to set CONFIG_FLOPPY_CACHE to 0 because it makes sense, but it hits me every time I move to one of the XT class systems (or the AT/286) and run off of a floppy drive.  So I'm arguing in circles, aren't I? 

> 
> If so, then perhaps its not a big deal that REL_INITSEG be squashed to smaller size - just release the portion of floppy cache that is not needed, which could be as small as 512 bytes (REL_INITSEG) when the floppy cache isn't configured at all, which is the rare case, right? In the other cases, the cache is overlapped with REL_INITSEG so REL_INITSEG's size doesn't matter much.
> 
> Is this along the lines of what you're now thinking?

Yes, this is the way it has worked since this morning and it looks really good. If there is no FDcache configured, REL_INITSEG is released and the 512b arena just sits there. In all other cases it's part of FDcache and invisible. If it can be reduced to a few paragraphs, fine, but the way this looks now it's low priority.

ghaerr:
> keeping a FDcache configured becomes cheap enough to let it stay in there, 5k or 7k. Which in practice means that if unused, /bin/init's CSEG or DSEG is going to end up in that arena and very little is wasted.

Agreed.

> but it hits me every time I move to one of the XT class systems (or the AT/286) and run off of a floppy drive. So I'm arguing in circles, aren't I?

Yes, you are. IMO there is a lot to be said about not having to configure a system and having it "just work". Programmers don't realize how important this is to users until they run into it themselves, which is what you're doing here. Also, the "don't fix until its broken" rule can apply here: how often is the system actually running out of memory, and what can be done (UMB, high memory buffers or kernel data segment, etc) can be done about it, instead of requiring the system to be reconfigured to run (on older systems in this case)?

We have a discussion going on in ELKS right now where a Nano-X BIOS 8x14 font isn't available on newer systems. Long story short, a configuration option was discussed that would allow for Nano-X to be recompiled with RAM fonts for those that need it. Then I mentioned how I've been removing, not adding, configuration options for ELKS for the last four years - and producing a system where, if it can, it just works. That's the (new) ELKS way. And guess what? Someone found a way to get an alternative 8x16 font, and after that, a way to automate determining whether the system had 8x14 vs 8x16 fonts, and voila - a contribution is on the way that adds no extra RAM fonts and "just works". Just took a little more programmer time, and the end users won't ever have to worry about it. I'm kinda proud of that.

>  what can be done (UMB, high memory buffers or kernel data segment, etc)

On this note, I have working patches that can allow for ELKS or TLVC to run with a full extra 64K of memory (using HMA), that can host either system buffers or in fact the entire kernel data segment. A huge ram upgrade for those who need it, and also have working HMA (typically HMA will work if the system can run protected mode OSes, as it has memory >= 1M and a working BIOS-control on the A20 gate). We can talk about this at another time.

Mellvik:
BTW, @ghaerr - you may be interested in this one: The routing error we discussed (`make_kb()`) turned out to need more (or less) than the divide adjustment. Dividing the reminder by 103 instead of 102 correctly changed the '.10' to '.9' but also turned 0.5 into 0.4. The best I could come up with was this:
```
	else {
		if (r > 1019) r = 1019;		/* avoid confusing rounding errors */
		sprintf(n, "%-u.%1uK", t, r/102);
	}
```

Mellvik:
>  IMO there is a lot to be said about not having to configure a system and having it "just work". Programmers don't realize how important this is to users until they run into it themselves, which is what you're doing here. Also, the "don't fix until its broken" rule can apply here: how often is the system actually running out of memory, and what can be done (UMB, high memory buffers or kernel data segment, etc) can be done about it, instead of requiring the system to be reconfigured to run (on older systems in this case)?

I agree, that is a key point - making it simple to get started. (Which BTW, sent me down a new path: Will the system boot if both directHD and XD/MFM disks are configured in? That's a different story). Then again, not knowing what a random user might be doing on/with the system, preserving/recovering memory also remains a key objective. There will never be enough available RAM. I keep running out of memory all the time because that's how I use the system - pushing resources to provoke failures, being my own Simian Army. A regular user may run out of memory for a number of different reasons, like not lack or RAM but RAM fragmentation. Or heap fragmentation, which has become less of a problem recently, first with the 'small allocations to high heap' mechanism, then the release of the `bootopts` buffer which does the same thing.

The 'simple to get started focus' has made me  reticent about UMB support: It's for experts only and not something that may be included in a default catch-all kernel. Also rarely useful for XT class systems which has received a lot of attention in TLVC since the beginning. With HMA on the horizon that may change, simply because it becomes too big a temptation.  (And simple to handle in `meminfo`:-) since they - UMBs and HMA - fall outside the complicated regions.)

> 
> We have a discussion going on in ELKS right now where a Nano-X BIOS 8x14 font isn't available on newer systems. Long story short, a configuration option was discussed that would allow for Nano-X to be recompiled with RAM fonts for those that need it. Then I mentioned how I've been removing, not adding, configuration options for ELKS for the last four years - and producing a system where, if it can, it just works. That's the (new) ELKS way. And guess what? Someone found a way to get an alternative 8x16 font, and after that, a way to automate determining whether the system had 8x14 vs 8x16 fonts, and voila - a contribution is on the way that adds no extra RAM fonts and "just works". Just took a little more programmer time, and the end users won't ever have to worry about it. I'm kinda proud of that.

Every reason to, I would say. This is open source collaboration at its best. And removing options an important focus. I have a handful on my list that I believe have been removed from ELKS already.

> 
> > what can be done (UMB, high memory buffers or kernel data segment, etc)
> 
> On this note, I have working patches that can allow for ELKS or TLVC to run with a full extra 64K of memory (using HMA), that can host either system buffers or in fact the entire kernel data segment. A huge ram upgrade for those who need it, and also have working HMA (typically HMA will work if the system can run protected mode OSes, as it has memory >= 1M and a working BIOS-control on the A20 gate). We can talk about this at another time.

Let's do that. As noted above, I'm interested.

Thanks.

Mellvik:
The final commit before merge includes the `init/main.c` missing from the previous commit and adds displaying the ext-buffer cache (if present) with the `-M` option. The plan is to eventually 1) include XMS memory in this display line, and 2) include the ext_buffer in the 'graphic'.

The dump below shows that the ext buffer is allocated at the top of conventional memory, not at the bottom. This is part of an upcoming experimental change.

```
tlvc15# meminfo -M
Kernel heap (DS-base 163C):
	   SEG   OFFS   SIZE
 Block 1: 1CF6   6BA2  37981 bytes
 Block 2: 1A21   3E5A   1024 bytes
 Kernel heap total:    39005 bytes

Main memory:
 Arena 1: 263C - 9FC0 (497728 bytes, 486K)
 Arena 2: 13C0 - 163C ( 10176 bytes, 9.9K)
 Arena 3: 0090 - 0250 (  7168 bytes, 7.0K)
 Ext buf: 95C0 - 9FC0 ( 40960 bytes, 40K)            <---------------

9FC0:0 +--------------------+ Top of conv. memory
       |                    |
       |                    |
       |                    |
       | Main memory        | 486K Arena 1
       |                    |
...
```

PR #122, opened 2024-12-30
[cmds] Another meminfo update

Mellvik:
`meminfo` will now show external buffers - if present - as a visible block in main memory with the `-P` and `-M` options.

Likewise, the assumption that the kernel DS segment always totals to 64k is gone, variable kernel heaps display correctly when the graphic options are used.

Plus some adjustments.

### DIFF
--- a/tlvc/arch/i86/kernel/system.c
+++ b/tlvc/arch/i86/kernel/system.c
@@ -41,13 +41,14 @@ unsigned int INITPROC setup_arch(void)
 	/* Heap allocations at even addresses */
 	endbss = (unsigned int)(_endbss + 1) & ~1;
 
-	/* Calculate size of heap, which extends end of kernel data segment */
+	/* Calculate or set size of heap which - unless set specifically,
+	 * extends to the end of end of the kernel DS */
 
-#ifdef SETUP_HEAPSIZE
-	if (!heapsize)          /* may also be set via heap= in /bootopts */
+#ifdef SETUP_HEAPSIZE		/* MK88 only */
+	if (!heapsize)
 	    heapsize = SETUP_HEAPSIZE;
 #endif
-	if (heapsize) {
+	if (heapsize) {			/* may be set via heap= in /bootopts */
             heapsegs = (1 + ~endbss) >> 4;  /* max possible heap in segments */
             if ((heapsize >> 4) < heapsegs) /* allow if less than max */
         	heapsegs = heapsize >> 4;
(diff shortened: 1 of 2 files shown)


Mellvik:
@ghaerr, you may be interested in this little weirdness in `tiny_printf`: 

With this printf format:
`                printf("       | %-19s| %3s %s\n", txt, make_kb(n, size), suffix);`
When the txt parameter is
`      txt = "[setup data]";`
the output is correct (I'm adding some context to visualise the alignment):
```
00B0:0 +--------------------+
       | [setup data]       | 0.5K Main memory arena 3
0090:0 +--------------------+
```
but when the txt parameter is
`      txt = "[setup-data]";`
the output looks like this:
```
00B0:0 +--------------------+
       | [setup–data]     | 0.5K Main memory arena 3
0090:0 +--------------------+
```

ghaerr:
> you may be interested in this little weirdness in tiny_printf:

That is very strange indeed! Are you sure the only difference is/was the '-' in the middle of the txt string? If so, I'll go track it down. There's nothing in `__fmt` that even looks at the output data, just uses the pad character to fill out the width, and pad is either space or '0'.

Mellvik:
> > you may be interested in this little weirdness in tiny_printf:
> 
> That is very strange indeed! Are you sure the only difference is/was the '-' in the middle of the txt string? If so, I'll go track it down. There's nothing in `__fmt` that even looks at the output data, just uses the pad character to fill out the width, and pad is either space or '0'.

Yes, I spent quite some time on this not immediately understanding why the alignment was off. This is obviously from `meminfo` - the actual context is this:
```
void p_block(int count, long_t size, char *txt, char *suffix)
{
        int i, mid=1;
        char n[8];

        mid = count/2;
        for (i = 0; i < count; i++) {
            if (i == mid)
                printf("       | %-19s| %3s %s\n", txt, make_kb(n, size), suffix);
            else
                p_empty();
        }
}
```

I figure it must be related to the presence of the square brackets?

Mellvik:
Sure enough - I just tested that: Removing the square brackets and putting back the hyphen and all is good.

ghaerr:
> Removing the square brackets and putting back the hyphen and all is good.

That's even stranger. Now that I look at `__fmt` it does look at the first character of the string being output, to see if its a negative number, and adjusts the width. This is incorrect for strings that are not numbers. But it only checks for '-', not '['.

The relevant code is:
```
static int
__fmt(FILE *op, unsigned char *buf, int ljustf, int width, int preci, char pad, char sign)
{
   int cnt = 0, len;
   unsigned char ch;

   len = strlen((char *)buf);

   if (*buf == '-') // <--- the first char of "string" is checked for negative here
      sign = *buf++;
   else if (sign)
      len++;
```
Can you check your version of tiny_printf.c to see if that code is the same, or was perhaps changed?

ghaerr:
> Removing the square brackets and putting back the hyphen and all is good.

If that's the case, then the output should be incorrect regardless of whether there's a hyphen or not, it would only depend on having square brackets as the first character. So something is amiss.

Mellvik:
Ahhhh, it must be the ansi interpretation in the term window! I'll check a different term wintype.

Mellvik:
Sorry - false alarm: There is actually invisible garbage in the string! Shows when using the console.
Skjermbilde 2024-12-30 kl  19 00 01

Thanks

PR #124, opened 2025-01-04
[kernel] Add missing device open before MBR access in genhd.c

Mellvik:
Possibly a leftover from when BIOS block IO was the rule, `genhd.c` has been reading (`bread`) the disk MBR without opening the device first. Not a problem until a disk device driver did some device initialization in _open - which is a natural thing to do. See also the discussion in https://github.com/Mellvik/TLVC/pull/114#issuecomment-2546051445 and onwards.

This PR fixes that by faking an inode and calling `blkdev_open()` before reading the MBR. The fix is using a trick to avoid having to call device close after the access: By setting the `S_IFCHR` bit in the inode's `i_mode` variable, the HD drivers will skip incrementing the access counter, which would be the only useful thing a call to _release could do in this case.

### DIFF
--- a/tlvc/arch/i86/drivers/block/genhd.c
+++ b/tlvc/arch/i86/drivers/block/genhd.c
@@ -24,12 +24,14 @@
 #include <linuxmt/major.h>
 #include <linuxmt/string.h>
 #include <linuxmt/memory.h>
+#include <linuxmt/stat.h>
 #include <arch/directhd.h>
 
 #include <arch/system.h>
 
 #include "blk.h"
 
+#define MUST_OPEN_DEVICE		/* open the device before bread of blk 0 */
 #define NR_SECTS(p)		p->nr_sects
 #define START_SECT(p)		p->start_sect
 
@@ -224,7 +226,16 @@ static int INITPROC msdos_partition(struct gendisk *hd,
     register struct hd_struct *hdp;
 #endif
     unsigned short int i, minor = current_minor;
+#ifdef MUST_OPEN_DEVICE
+    struct inode inode;
 
+    inode.i_rdev = dev;
+    inode.i_mode = S_IFCHR;	/* Pretend it's a raw device: No _release needed */
+    if (blkdev_open(&inode, NULL)) {
+	printk(" device open failed\n");
+	return -1;
+    }
+#endif
     if (!(bh = bread(dev, (block_t) 0))) {
 	printk(" no MBR");
 	return 0;


ghaerr:
Well, I suppose something is needed, since more sophisticated block drivers need to be opened before calling their entry points. But this is pretty much a dirty hack, and IMO is the sort of thing that will come back and bite years later when all is forgotten. Or perhaps much sooner.

The issues I raise are mostly concerned with the assumptions taken with regards to an unknown block driver, but assumed valid for specific current device drivers:
- Code assumes block driver doesn't use contents of inode meaningfully, and passes an uninitialized inode filled with garbage to driver. There's a routine get_emtpy_inode() to do that properly. But that's still a poor approach because many drivers use inode contents (e.g. inode->i_size, etc) for internal calculations.
- Code assumes block driver doesn't need release called. This is the mother of memory leaks and all sorts of future issues.
- Code sets fake char device planning on certain block device behaviors being followed - namely that char devices don't need release. That assumption alone is bad kernel design - all devices should mandate open and release, not special cased because of "requirements" of problems of some upper level code.

Don't take this the wrong way, I don't mean my comments harshly. Its just that I've been cleaning up lots of this kind of stuff or the last four years at ELKS, when bad or incomplete design was added slowly over time, resulting in a buggy mess. Kernel programming must be done precisely and correctly, (as should all code, really), as code committed tends to be forgotten, buried, and then buggy (to the next maintainer, of course!) :)

Mellvik:
Thanks @ghaerr - comments appreciated (and not taken the wrong way at all, more like as expected). My first version (below) did not use the 'skip release' trick, quite a bit more code but this is INITPROC code so it's not a big deal. Then I got the idea about the S_IFCHR trick and had to check it out. Less code and in itself, very simple. 

While agreeing with the concerns and principles you convey (last two points), the case for skipping the _release call in this particular case is as strong as for the char driver to do the same: We're performing an atomic operation - reading a single disk block into a fixed memory location, outside any buffer system or other dependencies. Not textbook pretty, but efficient, pragmatic.

The 'first cut' code looked like this:
```
#ifdef MUST_OPEN_DEVICE
    struct file_operations *fop;
    struct inode inode;

    inode.i_rdev = dev;
    inode.i_mode = 0;
    fop = get_blkfops(MAJOR(dev));
    if (fop->open(&inode, NULL)) {
        printk(" device open failed\n");
        return -1;
    }
    bh = bread(dev, (block_t) 0);
    fop->release(&inode, NULL);

    if (!bh) {
#else
    if (!(bh = bread(dev, (block_t) 0))) {
#endif
        printk(" no MBR");
        return 0;
    }
```

As to your first point - about the fake inode, I don't agree (I did take a look at `get_empty_inode()`, which obviously would work but seemed to me as complete overkill for the case). IMO a storage driver open() routine is there to fill in inode content, using only the two fields initialized manually here. If I'm mistaken I'd love to see a better way of doing this.

Thanks for the code review - I figure this may eventually be useful in ELKS too.

ghaerr:
> which obviously would work but seemed to me as complete overkill 

There's lots of subtle issues when it comes to the bugs down the line I'm referring to... even to the supposedly smallest things, like using a proper routine or kluging it with an uninitialized data structure, passed to a driver that you're already making assumptions about fields it might or might not use. But that's not even the subtly - in this case, just "declare an inode on the stack" - as though that's nothing. Have you checked how big an inode is? One can't just declare variables on the stack in the kernel, or you get stack overflow (another bug I've worked hard to overcome, and there's still way too many variables declared and buried deep in code, but few in ELKS. Linus had good reason to demand that all function variables get declared at top of function - so one can calculate stack usage). In this case, while seemingly not an issue, needlessly declaring a large structure on the stack allows for possible stack overflow should the routine be used elsewhere, or async operations operating on the limited interrupt stack, etc, etc. I only mention this as one of many subtle things to think about when kernel programming, not as a yes/no answer to a particular hack. Having said all that, you won't find me passing uninitialized data to another function in the kernel!

I won't go on. My suggestions for kernel programming is: always do it the right way, not the "fast" programmer way, or the "convenient to the current design" way, when more consideration finds ways to do it, almost always using more programmer time, but also almost always saving future maintenance time, and usually ending with better design. On this topic, success is measured in the kernel never crashing, ever, and having elegance in the design and architecture of the code matching the problem at hand. My $.02!

Mellvik:
@ghaerr,
I agree with many, probably most of the points you're making. Good programming is an art. Part of that artistry is balancing creativity and rules, which inevitably means breaking the rules some times. There would be no ELKS, TLVC, Unix or Linux if the rules of good programming were to be followed at all times. The existence of a goto in C is a great example. Hardly ever good programming, but remarkably efficient at times, sometimes even required.

Back to the case at hand: In any OS, initialization and startup is a more or less continuous special case with dependencies and assumptions left and right, implicit and explicit. Unless the developer understands exactly what's going on, there will be unpleasant surprises - even if all the rules are followed. If anywhere this is the place for creativity, not only to ensure functional correctness but to avoid bloat.

Specifically, let's keep in mind that the introduction of a device open in `genhd.c` is not required for functionality or stability. It's a structural improvement to make the system more robust against future surprises. As such, the addition fits perfectly into our agreed-upon principles for good programming. The addition itself does not, because it makes assumptions that stretches beyond the routine and even the file in which it exists. And it has constructs that don't fit into general best practices. This may be avoided by adding more code (bloat) and more dependencies/assumptions. So where is the balance?

It's a subjective evaluation which we may never completely agree on, but that doesn't make the discussion less valuable. For example, the sole purpose of the inode struct in this case is to pass parameters to the driver `open` routine. A `memset` to initialize it would be good programming but it's not useful. Allocating it on the stack eliminates a dependency (using `get_free_inode()` assumes that `init_inodes()` has been called). It may be argued that the following `bread` assumes that `init_buffers()` has been called, which in turn must be preceded by `init_inodes()`, but it's still a dependency. And while allocating a big structure like `inode` on the stack of a running kernel is bad, this is not a running kernel. If this allocation creates havoc, there is a serious problem somewhere else and detecting it early may be a good thing.

I think it's useful to take a step back from the details and see what we're actually doing here: Device initialization, including reading one block off of a device. It doesn't get much simpler than that. The code is used once per device detected, then discarded. This calls for simplicity while taking reasonable steps to avoid side effects that may pop up later. 

Storm in a teacup? Maybe - but interesting nevertheless!

PR #128, opened 2025-01-16
[cmds] Update and enhance the fdisk command

Mellvik:
A significant update/enhancement to `fdisk`:
Planned as just an update to force `fdisk` to use raw devices, but it turned into a lot more:

- Regular (image) files are now acceptable targets for `fdisk` - see below.
- The target device/file is now `stat`ed early on to determine the optimal course of action: 
    - Reject floppy devices (using major device number) and block devices - unless CONFIG_BLK_DEV_BIOS is set - and to determine device type (or plain file).
    - Hard disks must be raw devices - again unless CONFIG_BLK_DEV_BIOS is set. If `fdisk` is using the ROOTDEV (no device on the command line), the name is changed into the corresponding raw device.
- Warnings are issued if no MBR is found and when the 'device' is a plain file. Warnings may be suppressed via the `-q` option. 
- `fdisk` can now operate on plain files, and can initialize a file if it has no MBR if the new `-g` (geometry) flag is used: `-g c/h/s` to set the geometry. This is a simplification of the 'big' Linux variant using `-C`, `-H`, `-S` for the same purpose. It could easily have created the file too, but `dd` manages that task more than adequately.
- If an image file has an MBR already, the geometry is determined (guessed) from the partition table unless the `-g` option is used, which is more reliable. If there is no MBR, a missing ´-g` option is an error and will cause termination.
- The _cylinder_ value of the geometry specification is really superfluous since the size of the file is easily available. It does, however, provide some extra flexibility at minimal  cost. The cylinder value is sainty checked against the size and adjusted if too large. The heads and sectors values are not sanity checked.
- Using the geometry option when the target is a device is not allowed since the geometry is obtained from the device itself. If there is a need to override the device data, the `hdparm=` setting in `bootopts` should be used.

The practical value of being able to work with image files is limited at this point, and was inspired by the fact that the `usage()` message in the original program mentioned it (`disk_device_or_image`) - and subsequently that its implementation turned out to be rather trivial. Whenever loopback volumes become supported, the ability to treat files may come in handy. 

An updated man page is part of this PR.

### DIFF
--- a/tlvccmd/disk_utils/fdisk.c
+++ b/tlvccmd/disk_utils/fdisk.c
@@ -15,14 +15,18 @@
 #include <ctype.h>
 #include <string.h>
 #include <sys/ioctl.h>
-#include <sys/types.h>
 #include <sys/stat.h>
-#include <linuxmt/kdev_t.h>
+//#include <linuxmt/kdev_t.h>
 #include <linuxmt/major.h>
 
 #ifdef __ia16__
 #include <arch/hdio.h>
 #endif
+
+#define SECT_SIZE	512
+/* from linuxmt/kdev_t.h */
+#define MINORBITS	8
+#define MAJOR(dev)	((unsigned short)((dev) >> MINORBITS))
 
 struct partition
 {
@@ -36,7 +40,7 @@ struct partition
     unsigned char end_cyl;	/* end cylinder */
     unsigned short start_sect;	/* starting sector counting from 0 */
     unsigned short start_sect_hi;
-    unsigned short nr_sects;		/* nr of sectors in partition */
+    unsigned short nr_sects;	/* nr of sectors in partition */
     unsigned short nr_sects_hi;
 };
 
@@ -64,7 +68,7 @@ struct geometry {
 
 int pFd, initialize, is_file, quiet, chs[3];
 char dev[80];
-unsigned char MBR[512];
+unsigned char MBR[SECT_SIZE];
 
 typedef struct {
   int cmd;
@@ -335,8 +339,8 @@ void write_out()
     else {
         MBR[510] = 0x55;
         MBR[511] = 0xAA;
-	if ((i=write(pFd,MBR,512))!=512) {
-	    printf("error: wrote %d of 512 bytes to the partition table.\n", i);
+	if ((i=write(pFd,MBR,SECT_SIZE))!=SECT_SIZE) {
+	    printf("error: wrote %d of %d bytes to the partition table.\n", i, SECT_SIZE);
 	} else
 	    printf("Partition table written to %s\n",dev);
     }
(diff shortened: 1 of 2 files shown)


ghaerr:
> The practical value of being able to work with image files is limited at this point,

It would seem a nice addition to have fdisk.c compilable as both a host and TLVC application. In that way, the tool could be used on any system (e.g. macOS) to produce MBR images, rather than the current way of cobbling together an MBR block, `dd` image and then running `setboot` with special options to rewrite the partition table.

On the matter of host applications, I am moving towards adding a number of "ELKS" applications to be automatically built as tools for use with ELKS/TLVC images, files, etc: `eman`, `compress`, `decomp`, `hd`, `basic`, `disasm` are examples, which come in very handy to be used outside of ELKS itself. Some of these are already being built with the same name as the ELKS executable but with a "host" prefix to the name. Others are not yet, but when compilable would be copied to tools/bin with the same name and then the host prefix discarded.

Mellvik:
A quick test later shows `mkfat` to compile out of the box on macOS, while `fdisk` runs into a shipload of typedef collisions. I suspect you may have a handle on how to attack that...

ghaerr:
>  It would however be trivial to have it create a file with a given size and a default MBR and a single partition spanning the drive - using a corresponding non-interactive option. Is that what you're looking for?

I'm not really looking for anything, but was just commenting that using fdisk rather than setboot to write an MBR partition entry would seem more appropriate. I agree with you thought that fdisk doesn't need to get into the business of becoming 'dd' and installing MBR code though - we already have a decent mechanism for that.

> I suspect you may have a handle on how to attack that...

Yes, you'll have to think through getting rid of the \#include <linuxmt/...> that ends up pulling in the internal kernel types.h typedefs, in order to not conflict with possible conflicting host definitions for the same thing. The ELKS kernel uses its own typedefs for a majority of types which are now not exposed through the libc .h interface. However,  some non portable programs need to use kernel headers, and these drag in linuxmt/ or arch/types.h. Some might be able to be cleaned up, otherwise more ifdefs in the source. It's one of those long term time consuming projects. Ultimately, using as a few as possible <linuxmt/> includes and possibly reorganizing certain exposed headers is the desired solution for moving in the direction of portable tools.

Mellvik:
> Yes, you'll have to think through getting rid of the #include <linuxmt/...> that ends up pulling in the internal kernel types.h typedefs, in order to not conflict with possible conflicting host definitions for the same thing. The ELKS kernel uses its own typedefs for a majority of types which are now not exposed through the libc .h interface. However, some non portable programs need to use kernel headers, and these drag in linuxmt/ or arch/types.h. Some might be able to be cleaned up, otherwise more ifdefs in the source. It's one of those long term time consuming projects. Ultimately, using as a few as possible includes and possibly reorganizing certain exposed headers is the desired solution for moving in the direction of portable tools.

Yeah, turned out (in `fdisk`) that `kdev_t.h` was pulled in just to get the `MAJOR()` macro, so that was an easy fix. `mkfs` is a different story but unimportant for now. BTW - macOS has a very powerful `fdisk` that easily does whatever the TLVC version does and much more. Same with Linux and - I suspect - Windows. That said, the utilities you mention - `basic` and `disasm` in particular - would be interesting to have available.

ghaerr:
> BTW - macOS has a very powerful fdisk

Didn't know that, thanks.

> mkfs is a different story but unimportant for now.

I've made a note about looking into mkfs, mkfat, etc as host applications.

> basic and disasm in particular - would be interesting to have available.

Both are currently built using "make host" in the elkscmd/basic and elkscmd/debug directories. I plan on adding a mechanism that will allow all utilities like this to be built and installed into tools/bin.

Issue #66, opened 2024-06-29
IBM 5155 console keyboard 


Mellvik:
Not seen on other hardware, but on the IBM 5155, the "XT Portable", the keyboard doesn't work with TLVC. 

Works fine with DOS, serial io is not affected. Also works fine with COMPAQ Portable Plus, also an XT class machine. Since the 5155  has a standard XT mainboard, this problem presumably hits all IBM PC and XT models.

vkoskiv:
Confirmed on my 5150 (+5160 expansion unit), 83-key keyboard doesn't seem to work on the login prompt.

Mellvik:
It turns out this problem appears only when the root volume is FD. Booting from HD, or even from FD with root=<some HD volume> in bootopts, and we're good. So it seems the sequence of initialization is key. 
When starting from HD, using the FD doesn't affect the keyboard, it still works. In short it seems like if the FD is accessed (driver opened) before the console (keyboard really) is completely initialized, the kbd locks up. Maybe an interrupt mask.

@ghaerr, would you possibly have an idea about where to look for this one?

ghaerr:
> In short it seems like if the FD is accessed (driver opened) before the console (keyboard really) is completely initialized, the kbd locks up.

If setting the root volume in /bootopts causes the problem to go away, then I would disagree with your conclusion above, as changing the root volume in /bootopts doesn't reorder any console initialization.

What does differ between actual BIOS boot to a drive number versus /bootopts etc options is that on BIOS boot, register DL contains the BIOS drive number to boot from. The boot loader then uses this information to boot from the specified device. But I don't think that sounds like the issue.

Other possible issues are that the A20 line is turned on/off/tested at boot when XMS is enabled - and A20 fiddling uses the keyboard controller. You can look at this code in elks/arch/i86/lib/unreal.S, functions enable_a20_gate, etc. That file has a couple of options to enable/disable A20, and by default uses an INT 15h function, which possibly isn't implemented on your system, or something strange? The top of the file documents the various options, if this is even the issue.

There may be something funky with the actual communications to the keyboard controller chip, but I don't remember offhand which bits are masked and rerouted, etc. It might be worth looking at the 8042  in the IBM XT documentation, so we can have a more technical discussion.

[...]

PR #137, opened 2025-01-31
[cmd] clock: Fix XT keyboard lockup when accessing non existent CMOS

Mellvik:
Reported in issue #66 - the problem turns out to be the CMOS probe which `clock` calls to determine whether the system has an AT style CMOS RTC or not. It turns out that a single read to the CMOS 'command' port is all it takes to lock up the keyboard. 

Note that this problem affects IBM XT systems only, not any compatibles that we are aware of. IOW the spruce of the problem is a hardware 'anomaly' in the IBM design.

While the obvious solution (or workaround) would be to test for system type and skip the probe if running on a pre-AT system (this is what the `directfd` driver does and the reason we haven't seen the problem before). 

This fix (workaround) changes `clock` to not probe the CMOS if the -A (for AST) option is present. If running an XT system with no RTC, the real fix is not to run `clock` at all (comment it out from `/etc/rc.sys`).

Based on the comments from @ghaerr in https://github.com/Mellvik/TLVC/issues/66#issuecomment-2626070550 it is likely that there will be amendments to this PR before merge turning it into a real fix instead of a workaround.

### DIFF
--- a/tlvccmd/rootfs_template/etc/rc.sys
+++ b/tlvccmd/rootfs_template/etc/rc.sys
@@ -9,8 +9,6 @@ export PATH=/bin
 # get TZ timezone variable from /bootopts or /etc/profile
 source /etc/profile
 
-# comment out clock if you don't have one (XT class) or add -A
-# (AST Sixpakplus) to avoid keyboard lockup on IBM 5150, 5155, 5160
 clock -s -u
 
 # mount and check filesystems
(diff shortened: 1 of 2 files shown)


Mellvik:
Interestingly, the compaq Portable Plus, an XT clone, has the same missing A4 line as the XT, but never displayed those symptoms (IOW, never a keyboard problem). FWIW (and for the record) here's the readout from ports 70 and 71 from various systems:
```
systype   inb(0x70)  inb(0x71)
--------------------------------------
86box (xt)  0xff      0xff
IBM5155        0      0x48
AMI386SX    0xff      0xff
XTclone     0x8c      0x8d    (Eliah Miller's 88board  w/Sergey's RTC)
QEMU        0xff      0xff
Compaq         0      0x48    (compaq portable Plus XT clone)
```

Issue #138, opened 2025-02-05
mv directory hangs the system


Mellvik:
Moving directories emit error messages (that may be purely informational) and works initially. Eventually - in the 7th `mv` actually - the system will hang, responding to pings and keystrokes but nothing else (presumably a tight kernel loop). The following commands will trigger the problem:
```
[tlvc15] TLVC 0.6.0

login: root
tlvc15# cd /tmp
tlvc15# mkdir xx 1 2 3 4 5 6 7 8 9
tlvc15# mv ? xx
iput: trying to free free inode dev 0000 inode 361 mode 0040755
iput: trying to free free inode dev 0000 inode 362 mode 0040755
iput: trying to free free inode dev 0000 inode 363 mode 0040755
iput: trying to free free inode dev 0000 inode 364 mode 0040755
iput: trying to free free inode dev 0000 inode 365 mode 0040755
iput: trying to free free inode dev 0000 inode 366 mode 0040755
iput: trying to free free inode dev 0000 inode 367 mode 0040755
   <---- hang

```

The behaviour is repeatable on any platform including QEMU. And on ELKS. The 7th move hangs predictably - also when other activity has taken place in between the moves.

ghaerr:
> inode dev 0000

Yes, there's a few kernel messages in the filesystem code to show when things "can't happen". The problem is likely multiple `iput`s somewhere in the chain of function calls, but that won't be all of it since the system crashes.

I seem to remember a kernel mv problem similar to this from long on ELKS. I just happened to be looking though some of the elks/fs top-level filesystem code the other day - this is one of the the last area of wilderness left, and some of the code is not straightforward. Just yesterday I found a new system call bug where if `access` is called prior to a file open, the file open doesn't work. Sometimes.

Will the kernel execute a single mv properly, or does it always display this message and the directory isn't renamed? I'm also assuming this only occurs on directory inodes.

Mellvik:
As the text points out, the behaviour is similar in ELKS so you might check it out. The 'advisory' message is always there and the command works fine - that is, 6 of them to, not the 7th...

ghaerr:
> that is, 6 of them to, not the 7th.

If that's the case, perhaps its just a missing iput - and the system busy looping trying to find a free buffer after running out. Perhaps try this with a different L1 buffer setting to remove/prove that theory?

Mellvik:
> Just yesterday I found a new system call bug where if access is called prior to a file open, the file open doesn't work. Sometimes.

[...]

PR #139, opened 2025-02-10
[kernel] Fix for mv directory hang in #138

Mellvik:
Fixes both problems reported on #138:

1) The `trying to free free inode` error message turned out to be harmless and caused by an unnecessary `iput` after a call to `do_mknod` in `fs/namei.c`. 

2) The system hang after `mv`ing a certain number of directories turned out to be a problem in `rmdir()`, specifically in `empty_dir()` in `minix/namei.c`, where a leftover from our mega-L1-L2 buffer cleanup last year left a buffer mapped when it should have been unmapped. Any action that involved rmdir would leave a L1 buffer mapped forever, and the system would eventually run out and go into a spin waiting for a release.

### DIFF
--- a/tlvc/fs/buffer.c
+++ b/tlvc/fs/buffer.c
@@ -744,7 +744,7 @@ void unmap_buffer(struct buffer_head *bh)
         } else
 #endif
         if (--ebh->b_mapcount == 0) {
-            debug_map("unmap: %d\n", buf_num(bh));
+            debug_map("unmap: %d/%d\n", buf_num(bh), ebh->b_blocknr);
             wake_up(&L1wait);
         } else
             debug_map("unmap_buffer: %d mapcount %d\n", buf_num(bh), ebh->b_mapcount+1);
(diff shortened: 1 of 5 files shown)


ghaerr:
Nice work finding and fixing this one @Mellvik!

Did you end up solving this by code inspection, or did you have to enter a few rabbit holes and testing to nail it? The FS code is complicated and I have yet to fully read and understand all of it. Was the incorrect location of unmap_brelse something that's been in the FS code forever, or was that changed during the L1/L2 buffer cleanup process last year?

Kind of amazing this late in the game that rmdir doesn't actually work for very long and we didn't know it. Glad to have another hanging kernel bug fixed, thank you!

Mellvik:
Thank you @ghaerr,
this turned into another rabbit chase for the exact reason you mention: This code is indeed complicated and at times really hard to follow - much fun with the extensive use of structures like `inode_operations` and `file_operations` indexed by `offsetof()` and varying number of arguments. 

I haven't checked but I'm assuming - given the severity - that the bug, a `brelse()` that should be an `unmap_brelse()`,  must be a 'leftover' from the L1/L2 cleanup. 

Anyway, yes, it's great to have another one bite the dust.

Issue #59, opened 2024-04-18
/bin/sh lingering mount problem


Mellvik:
If a shell is started from a (sub)mounted filesystem, then exited, that fs can no longer be unmounted (boot required).
- Applies to `ash` only, `sash` is fine
- Applies to both FAT and minix fs
I haven't investigated yet, but it seems `sh` may be leaving open file descriptors behind somehow.

```
TLVC 0.6.0

login: root
tlvc17# mount /dev/dhda2 /mnt
MINIX: inodes 21845 imap 3 zmap 8, first data 696
MINIX-fs: mounting unchecked file system 0x502, running fsck is recommended.
tlvc17# umount /mnt
tlvc17# fsck -a /dev/rdhda2
Filesystem on /dev/rdhda2 is dirty, needs checking.
tlvc17# fsck -a /dev/rdhda1
Filesystem on /dev/rdhda1 is dirty, needs checking.
tlvc17# !mo
mount /dev/dhda2 /mnt
MINIX: inodes 21845 imap 3 zmap 8, first data 696
tlvc17# /mnt/bin/sh
# df
Filesystem    1K-blocks     Free     Used    %  FUsed%  Mounted on
/dev/dhda3        65535    62984     2551   4%     1%   /
/dev/dhda2        65535    63707     1828   3%     1%   /mnt
# ^D
tlvc17# !um
umount /mnt
/mnt: Device or resource busy                         <----- cannot unmount even if process is gone
tlvc17# mount /dev/dhda1 /mnt2                   <--- mounting a different fs this time
MINIX: inodes 21845 imap 3 zmap 8, first data 696
tlvc17# /mnt2/bin/sash                                  <----- try sash
# ls /mnt2
.
..
386.linux
bin
bootopts
bootopts.386
dev
etc
good.linux
linux
mnt
root
tmp
# ^D
tlvc17# umount /mnt2                               <----- no problem unmounting
tlvc17# mount /dev/dhda4 /mnt2            <----- mounting FAT fs
FAT: me=f8,csz=16,#f=2,floc=1,fsz=147,rloc=295,#d=512,dloc=327,#s=601776,ts=601776
FAT: total 300M, fat16 format
tlvc17# /mnt2/bin/sh
/mnt2/bin/sh: not found
tlvc17# cp /bin/sh /mnt2/bin/
tlvc17# sync
tlvc17# /mnt2/bin/sh
# pwd
/root
# ^D
tlvc17# umount /mnt2
[...]

PR #141, opened 2025-02-13
[kernel] Bugfix for occasional problem with umounting minix fs

Mellvik:
As reported in #59, unmounting a minix file system was impossible after having executed `ash` locally on that FS. The problem turned out to much broader than that and was finally located in `execve()` where a missing `iput` on the old/parent process inode caused the inode counter to continually increase and thus remain busy.

The problem had no other effect than preventing amounts until the system ran out of available inodes, which rarely happened. The fix is one simple line of code in `fs/exec.c`. 

@ghaerr, I found out that you fixed this in ELKS some time between v 0.7 and 0.8 - and it's my turn to ask: Was it code inspection or a regular bug hunt?

BTW thanks for the updated inode ^N listing code, making the path visual like that was great!

### DIFF
--- a/tlvc/fs/exec.c
+++ b/tlvc/fs/exec.c
@@ -131,7 +131,7 @@ int sys_execve(const char *filename, char *sptr, size_t slen)
     register struct task_struct *currentp;
     struct inode *inode;
     struct file *filp;
-    int retval;
+    int retval, i;
     __u16 ds;
     seg_t base_data = 0;
     segment_s *seg_code;
@@ -469,64 +469,60 @@ int sys_execve(const char *filename, char *sptr, size_t slen)
 
     /* Wipe the BSS */
     fmemsetb((char *)(size_t)mh.dseg + base_data, seg_data->base, 0, (size_t)mh.bseg);
-    {
-	register int i = 0;
+    i = 0;
 
-	/* argv and envp are two NULL-terminated arrays of pointers, located
-	 * right after argc.  This fixes them up so that the loaded program
-	 * gets the right strings. */
+    /* argv and envp are two NULL-terminated arrays of pointers, located
+     * right after argc.  This fixes them up so that the loaded program
+     * gets the right strings. */
+ 
+    slen = 0;	/* Start skiping argc */
+    do {
+	i += sizeof(__u16);
+	if ((retval = get_ustack(currentp, i)) != 0)
+	    put_ustack(currentp, i, (currentp->t_begstack + retval));
+	else slen++;	/* increments for each array traversed */
+    } while (slen < 2);
+    retval = 0;
 
-	slen = 0;	/* Start skiping argc */
-	do {
-	    i += sizeof(__u16);
-	    if ((retval = get_ustack(currentp, i)) != 0)
-		put_ustack(currentp, i, (currentp->t_begstack + retval));
-	    else slen++;	/* increments for each array traversed */
-	} while (slen < 2);
-	retval = 0;
+    /* Clear signal handlers */
+    i = 0;
+    do {
+	currentp->sig.action[i].sa_dispose = SIGDISP_DFL;
+    } while (++i < NSIG);
+    currentp->sig.handler = (__kern_sighandler_t)NULL;
 
-	/* Clear signal handlers */
-	i = 0;
-	do {
-	    currentp->sig.action[i].sa_dispose = SIGDISP_DFL;
-	} while (++i < NSIG);
-	currentp->sig.handler = (__kern_sighandler_t)NULL;
+    /* Close required files */
+    i = 0;
+    do {
+	if (FD_ISSET(i, &currentp->files.close_on_exec))
+	    sys_close(i);
+    } while (++i < NR_OPEN);
 
-	/* Close required files */
-	i = 0;
-	do {
-	    if (FD_ISSET(i, &currentp->files.close_on_exec))
-		sys_close(i);
-	} while (++i < NR_OPEN);
-    }
+    iput(currentp->t_inode);	/* 'close' the current process' inode */
+    currentp->t_inode = inode;
 
-    {
     /* this could be a good place to set the effective user identifier
      * in case the suid bit of the executable had been set */
 
-	iput(currentp->t_inode);
-	currentp->t_inode = inode;
-
     /* can I trust the following fields?  */
-	if (inode->i_mode & S_ISUID)
-	    currentp->euid = inode->i_uid;
-	if (inode->i_mode & S_ISGID)
-	    currentp->egid = inode->i_gid;
-    }
+    if (inode->i_mode & S_ISUID)
+        currentp->euid = inode->i_uid;
+    if (inode->i_mode & S_ISGID)
+        currentp->egid = inode->i_gid;
 
     currentp->t_enddata = (size_t)mh.dseg + (size_t)mh.bseg + base_data;
     currentp->t_endbrk =  currentp->t_enddata;
 
-	/* ease libc memory allocations by setting even break address*/
-	if ((int)currentp->t_endbrk & 1)
-		currentp->t_endbrk++;
+    /* ease libc memory allocations by setting even break address */
+    if ((int)currentp->t_endbrk & 1)
+    	currentp->t_endbrk++;
 
     /*
      *      Arrange our return to be to CS:entry
      */
     arch_setup_user_stack(currentp, (word_t) mh.entry);
 
-#if 0	/* used only for vfork()*/
+#if UNUSED	/* used only for vfork() */
     wake_up(&currentp->p_parent->child_wait);
 #endif
 
--- a/tlvc/fs/exec.c
+++ b/tlvc/fs/exec.c
@@ -504,6 +504,7 @@ int sys_execve(const char *filename, char *sptr, size_t slen)
     /* this could be a good place to set the effective user identifier
      * in case the suid bit of the executable had been set */
 
+	iput(currentp->t_inode);
 	currentp->t_inode = inode;
 
     /* can I trust the following fields?  */
--- a/tlvc/fs/inode.c
+++ b/tlvc/fs/inode.c
@@ -89,14 +89,15 @@ static void list_inode_status(void)
 
     do {
         if (inode->i_count || inode->i_dev || inode->i_dirt) {
-            printk("\n#%2d: dev %D inode %5lu dirty %d count %u", i, inode->i_dev,
-                (unsigned long)inode->i_ino, inode->i_dirt, inode->i_count);
+            inode->i_path[sizeof(inode->i_path)-1] = '\0';
+            printk("\n#%2d: dev %p inode %5lu cnt %2d %c %06o %s", i, inode->i_dev,
+                (unsigned long)inode->i_ino, inode->i_count, inode->i_dirt? 'D':' ',
+                inode->i_mode, S_ISSOCK(inode->i_mode)? " [socket]": inode->i_path);
         }
         i++;
         if (inode->i_count) inuse++;
     } while ((inode = inode->i_prev) != NULL);
     printk("\nTotal inodes inuse %d/%d (%d free)\n", inuse, NR_INODE, nr_free_inodes);
-
 }
 #endif
 
--- a/tlvc/fs/namei.c
+++ b/tlvc/fs/namei.c
@@ -195,6 +195,16 @@ static int dir_namei(const char *pathname, size_t * namelen,
     return error;
 }
 
+/* saves path for list_inode_status */
+static void save_path(struct inode *inode, const char *pathname)
+{
+#ifdef CHECK_FREECNTS
+    if (!inode->i_path[0])
+        fmemcpyb(inode->i_path, kernel_ds, (void *)pathname, current->t_regs.ds,
+            sizeof(inode->i_path));
+#endif
+}
+
 /*
  *	_namei()
  *
@@ -227,6 +237,7 @@ int _namei(const char *pathname, register struct inode *base, int follow_links,
 	    } else
 		iput(base);
 	    *res_inode = inode;
+	    save_path(inode, pathname);
 	} else
 	    iput(base);
     }
@@ -362,6 +373,7 @@ int open_namei(const char *pathname, int flag, int mode,
 	    inode->i_dirt = 1;
 	}
 	*res_inode = inode;
+        save_path(inode, pathname);
     } else iput(inode);
 #undef inode
 
--- a/tlvc/include/linuxmt/fs.h
+++ b/tlvc/include/linuxmt/fs.h
@@ -8,6 +8,7 @@
 #ifdef __KERNEL__
 
 #include <linuxmt/config.h>
+#include <linuxmt/trace.h>
 #include <linuxmt/types.h>
 #include <linuxmt/wait.h>
 #include <linuxmt/kdev_t.h>
@@ -239,6 +240,9 @@ struct inode {
 #endif
 		void * generic_i;
     } u;
+#ifdef CHECK_FREECNTS
+    char	i_path[16];
+#endif
 };
 
 struct file {
--- a/tlvc/kernel/exit.c
+++ b/tlvc/kernel/exit.c
@@ -37,55 +37,55 @@ static void reparent_children(void)
 /* note: 'usage' parameter ignored */
 int sys_wait4(pid_t pid, int *status, int options, void *usage)
 {
-	register struct task_struct *p;
-	int waitagain;
+    register struct task_struct *p;
+    int waitagain;
 
-	debug_wait("WAIT(%d) for %d %s\n", current->pid, pid, (options & WNOHANG)? "nohang": "");
+    debug_wait("WAIT(%d) for %d %s\n", current->pid, pid, (options & WNOHANG)? "nohang": "");
 
- for (;;) {
+    for (;;) {
 	waitagain = 0;
 
 	for_each_task(p) {
-		if (p->p_parent == current && p->state != TASK_UNUSED) {
-		  if (p->state == TASK_ZOMBIE || p->state == TASK_STOPPED) {
-			if (pid == (pid_t)-1 || p->pid == pid || (!pid && p->pgrp == current->pgrp)) {
-				if (status) {
-					if (verified_memcpy_tofs(status, &p->exit_status, sizeof(int)))
-						return -EFAULT;
-				}
-
-				/* just return status on stopped state, don't release task*/
-				if (p->state == TASK_STOPPED)
-					return p->pid;
-
-				p->state = TASK_UNUSED;		/* unassign task entry*/
-				next_task_slot = p;
-				task_slots_unused++;
-
-				debug_wait("WAIT(%d) got %d\n", current->pid, p->pid);
-				return p->pid;
+	    if (p->p_parent == current && p->state != TASK_UNUSED) {
+		if (p->state == TASK_ZOMBIE || p->state == TASK_STOPPED) {
+		    if (pid == (pid_t)-1 || p->pid == pid || (!pid && p->pgrp == current->pgrp)) {
+			if (status) {
+			    if (verified_memcpy_tofs(status, &p->exit_status, sizeof(int)))
+				return -EFAULT;
 			}
+
+			/* just return status on stopped state, don't release task*/
+			if (p->state == TASK_STOPPED)
+			    return p->pid;
+
+			p->state = TASK_UNUSED;		/* unassign task entry*/
+			next_task_slot = p;
+			task_slots_unused++;
+
+			debug_wait("WAIT(%d) got %d\n", current->pid, p->pid);
+			return p->pid;
+		    }
 		} else {
-			/* keep waiting while process has non-zombie/stopped children*/
-			debug_wait("WAIT(%d) again for pid %d state %d\n", current->pid, p->pid, p->state);
-			waitagain = 1;
+		    /* keep waiting while process has non-zombie/stopped children*/
+		    debug_wait("WAIT(%d) again for pid %d state %d\n", current->pid, p->pid, p->state);
+		    waitagain = 1;
 		}
-	  }
+	    }
 	}
 
 	if (options & WNOHANG)
-		return 0;
+	    return 0;
 	if (!waitagain)
-		break;
+	    break;
 
 	debug_wait("WAIT(%d) sleep\n", current->pid);
 	interruptible_sleep_on(&current->child_wait);
 	if (current->signal) {
-		debug_wait("WAIT(%d) return -EINTR\n", current->pid);
-		return -EINTR;
+	    debug_wait("WAIT(%d) return -EINTR\n", current->pid);
+	    return -EINTR;
 	}
 	debug_wait("WAIT(%d) wakeup\n", current->pid);
-  }
+    }
 
     debug_wait("WAIT(%d) return -ECHILD\n", current->pid);
 	return -ECHILD;
--- a/tlvccmd/sys_utils/ps.c
+++ b/tlvccmd/sys_utils/ps.c
@@ -226,7 +226,7 @@ int main(int argc, char **argv)
 		return 1;
 	}
 
-	printf("  PID   GRP  TTY USER STAT ");
+	printf("  PID  PPID  PGRP  TTY USER STAT ");
 #ifdef CONFIG_CPU_USAGE
 	printf("CPU");
 #endif
@@ -259,9 +259,10 @@ int main(int argc, char **argv)
 		}
 		pwent = getpwuid(task_table.uid);
 
-		/* pid grp tty user stat*/
-		printf("%5d %5d %4s %-8s%c ",
+		/* pid ppid grp tty user stat*/
+		printf("%5d %5d %5d %4s %-8s%c ",
 				task_table.pid,
+				task_table.ppid,
 				task_table.pgrp,
 				tty_name(fd, (unsigned int)task_table.tty, ds),
 				(pwent ? pwent->pw_name : "unknown"),


ghaerr:
> I found out that you fixed this in ELKS some time between v 0.7 and 0.8 
> thanks for the updated inode ^N listing code, making the path visual like that was great!

Thanks. Actually, this bug was identified as a result of adding the visual inode ^N listing for another hard-to-see issue, which  then showed an incorrect inode count value on an unrelated inode - the shell. Seeing that, I was able to track that down and fix it as a result of code inspection and some hunting.

You have probably seen that over the last couple of years, lots of instrumentation code has been added to ELKS. I had been leaning in this direction for years after spending tons of time fixing ELKS issues, but became completely convinced of  how important such tooling is after deeply studying the Cosmopolitan C library codebase, written by jart, an actual genius. In Cosmo, instrumentation was initially 25% of the project, and includes automatic function tracing/display, timing, crash stack dumps and register display, etc etc. 

What I have learned is that the ability to quickly see what is going on is just as important as doing it in the first place. Ideas and implementations like this for ELKS now include enhanced debug.h, sysctl-based malloc debug output, multiple memory allocators, CONFIG_APP_FTRACE option (command --ftrace) , adding of symbol tables to all executables, symbolic stack backtracking, ia16-elf-gcc -finstrument-functions-simple, and my [Blink16](https://github.com/ghaerr/blink16) project. Lots of work and many are WIP!

> It turned out to be a missing i_count decrement in do_exit(), probably been around forever.

Was this found to be incorrect, or is something still needed in do_exit?

Mellvik:
>> It turned out to be a missing i_count decrement in do_exit(), probably been around forever.

> Was this found to be incorrect, or is something still needed in do_exit?

it was incorrect as it didn't cover pure exec's in which there is no exit(), which sent me to `exec` and the 'right' solution.

Instrumentation: indeed - it was ultimately the inode list that solved this one at this end too. Very powerful indeed, topped by the addition of the inode 'name'.

A seriously ambitious list of projects on your plate - I'll be following that closely! Thanks.

Mellvik:
@ghaerr, it turns out there is a similar problem in the FAT fs code - inodes not being `iput`ted properly, is that something that has been seen/fixed in ELKS?
```
tlvc17# mv *.??_ ../cmprssed.622/
VFS: No free inodes
VFS: No free inodes

# 1: dev 0504 inode  4750 cnt  2   040755 
# 2: dev 0504 inode     1 cnt  1   040755 /mnt
# 3: dev 0504 inode 142549 cnt  2   100755 smartmon.ex_
# 4: dev 0504 inode 142548 cnt  1 D 100755 smartdrv.ex_
# 5: dev 0504 inode 142547 cnt  1 D 100755 sizer.ex_
# 6: dev 0504 inode 142546 cnt  1 D 100755 share.ex_
# 7: dev 0504 inode 142545 cnt  1 D 100755 setver.ex_
# 8: dev 0504 inode 142510 cnt  1 D 100755 restore.ex_
# 9: dev 0504 inode 142509 cnt  1 D 100755 replace.ex_
#10: dev 0504 inode 142544 cnt  1 D 100755 ramdrive.sy_
#11: dev 0504 inode 142543 cnt  1 D 100755 qbasic.hl_
#12: dev 0504 inode 142589 cnt  1 D 100755 print.ex_
#13: dev 0504 inode 142542 cnt  1 D 100755 power.ex_
#14: dev 0504 inode 142588 cnt  1 D 100755 mwundel.hl_
#15: dev 0504 inode 142587 cnt  1 D 100755 mwundel.ex_
#16: dev 0504 inode 142586 cnt  1 D 100755 mwgrafic.dl_
#17: dev 0504 inode 142585 cnt  1 D 100755 mwbackup.hl_
#18: dev 0504 inode 142584 cnt  1 D 100755 mwbackup.ex_
#19: dev 0504 inode 142541 cnt  1 D 100755 mwbackr.dl_
#20: dev 0504 inode 142583 cnt  1 D 100755 mwbackf.dl_
#21: dev 0504 inode 142582 cnt  1 D 100755 mwavtsr.ex_
#22: dev 0504 inode 142581 cnt  1 D 100755 mwavsos.dl_
#23: dev 0504 inode 142580 cnt  1 D 100755 mwavscan.dl_
#24: dev 0504 inode 142579 cnt  1 D 100755 mwavmgr.dl_
#25: dev 0504 inode 142578 cnt  1 D 100755 mwavdrvl.dl_
#26: dev 0504 inode 142577 cnt  1 D 100755 mwavdosl.dl_
#27: dev 0504 inode 142576 cnt  1 D 100755 mwavdlg.dl_
#28: dev 0504 inode 142575 cnt  1 D 100755 mwavabsi.dl_
#29: dev 0504 inode 142574 cnt  1 D 100755 mwav.hl_
#30: dev 0504 inode 142573 cnt  1 D 100755 mwav.ex_
#31: dev 0504 inode 142540 cnt  1 D 100755 mstools.dl_
#32: dev 0504 inode 142538 cnt  1 D 100755 msbconfg.hl_
#33: dev 0504 inode 142536 cnt  1 D 100755 msbackup.hl_
#34: dev 0504 inode 142535 cnt  1 D 100755 msbackup.ex_
#35: dev 0504 inode 142568 cnt  1 D 100755 msavirus.ls_
#36: dev 0504 inode 142567 cnt  1 D 100755 msavhelp.ov_
#37: dev 0504 inode 142566 cnt  1 D 100755 msav.hl_
#38: dev 0504 inode 142534 cnt  1 D 100755 move.ex_
#39: dev 0504 inode 142533 cnt  1 D 100755 mouse.co_
#40: dev 0504 inode 142531 cnt  1 D 100755 monoumb.38_
#41: dev 0504 inode 142530 cnt  1 D 100755 mode.co_
#42: dev 0504 inode 142529 cnt  1 D 100755 memmaker.in_
#43: dev 0504 inode 142528 cnt  1 D 100755 memmaker.hl_
#44: dev 0504 inode 142469 cnt  1 D 100755 mem.ex_
#45: dev 0504 inode 142526 cnt  1 D 100755 loadfix.co_
#46: dev 0504 inode 142525 cnt  1 D 100755 label.ex_
#47: dev 0504 inode 142506 cnt  1 D 100755 keybrd2.sy_
#48: dev 0504 inode 142524 cnt  1 D 100755 intersvr.ex_
#49: dev 0504 inode 142523 cnt  1 D 100755 interlnk.ex_
#50: dev 0504 inode 142522 cnt  1 D 100755 himem.sy_
#51: dev 0504 inode 142521 cnt  1 D 100755 help.hl_
#52: dev 0504 inode 142519 cnt  1 D 100755 graphics.pr_
#53: dev 0504 inode 142518 cnt  1 D 100755 graphics.co_
#54: dev 0504 inode 142517 cnt  1 D 100755 find.ex_
#55: dev 0504 inode 142516 cnt  1 D 100755 fc.ex_
#56: dev 0504 inode 142515 cnt  1 D 100755 fastopen.ex_
#57: dev 0504 inode 142514 cnt  1 D 100755 fasthelp.ex_
#58: dev 0504 inode 142505 cnt  1 D 100755 emm386.ex_
#59: dev 0504 inode 142504 cnt  1 D 100755 ega3.cp_
#60: dev 0504 inode 142503 cnt  1 D 100755 ega2.cp_
#61: dev 0504 inode 142502 cnt  1 D 100755 ega.cp_
#62: dev 0504 inode 142513 cnt  1 D 100755 edit.hl_
#63: dev 0504 inode 142501 cnt  1 D 100755 drvspace.sy_
#64: dev 0504 inode 142500 cnt  1 D 100755 drvspace.in_
#65: dev 0504 inode 142499 cnt  1 D 100755 drvspace.hl_
#66: dev 0504 inode 142512 cnt  1 D 100755 driver.sy_
#67: dev 0504 inode 142497 cnt  1 D 100755 doskey.co_
#68: dev 0504 inode 142496 cnt  1 D 100755 doshelp.hl_
#69: dev 0504 inode 142495 cnt  1 D 100755 dmdrvr.bi_
#70: dev 0504 inode 142494 cnt  1 D 100755 display.sy_
#71: dev 0504 inode 142493 cnt  1 D 100755 diskcopy.co_
#72: dev 0504 inode 142492 cnt  1 D 100755 diskcomp.co_
#73: dev 0504 inode 142491 cnt  1 D 100755 deltree.ex_
#74: dev 0504 inode 142490 cnt  1 D 100755 deloldos.ex_
#75: dev 0504 inode 142489 cnt  1 D 100755 defrag.hl_
#76: dev 0504 inode 142487 cnt  1 D 100755 dblwin.hl_
#77: dev 0504 inode 142457 cnt  1 D 100755 country.tx_
#78: dev 0504 inode 142485 cnt  1 D 100755 chkstate.sy_
#79: dev 0504 inode 142564 cnt  1 D 100755 bkudav.gr_
#80: dev 0504 inode 142563 cnt  1 D 100755 bkud.gr_
#81: dev 0504 inode 142562 cnt  1 D 100755 bkav.gr_
#82: dev 0504 inode 142561 cnt  1 D 100755 bk.gr_
#83: dev 0504 inode 142560 cnt  1 D 100755 av.gr_
#84: dev 0504 inode 142484 cnt  1 D 100755 append.ex_
#85: dev 0504 inode 142483 cnt  1 D 100755 ansi.sy_
#86: dev 0503 inode   151 cnt  1   100755 /bin/mv
#87: dev 0504 inode  4749 cnt  2 D 040755 dos622.all/
#88: dev 0503 inode    65 cnt  1   040755 /mnt
#89: dev 0503 inode   128 cnt  1   100755 /bin/sh
#90: dev 0503 inode   108 cnt  2   100755 /bin/getty
#91: dev 0503 inode    55 cnt  1   020644 /dev/ttyS2
#92: dev 0503 inode    29 cnt  1 D 020644 /dev/ttyS0
#93: dev 0503 inode    28 cnt  1   020644 /dev/tty1
#94: dev 0503 inode   184 cnt  1   100644 /etc/inittab
#95: dev 0503 inode   247 cnt  1   100755 /bin/init
#96: dev 0503 inode     1 cnt 12   040755 /
Total inodes inuse 96/96 (0 free)
```

ghaerr:
> it turns out there is a similar problem in the FAT fs code - inodes not being iputted properly, is that something that has been seen/fixed in ELKS?

Geez, no. But I seem to remember perhaps some comments and/or issues by you filed long ago that sound similar

It seems that perhaps this might be easily duplicated by mv'ing a single file to another directory, can you duplicate that on your test case? I also note a "dos622.all/" very early on in the inode listing, not sure whether that's the culprit here or if its just any FAT file move.

Mellvik:
> > it turns out there is a similar problem in the FAT fs code - inodes not being iputted properly, is that something that has been seen/fixed in ELKS?
> 
> Geez, no. But I seem to remember perhaps some comments and/or issues by you filed long ago that sound similar

Might very well be the case. Again we're seeing the incredible benefit of your new level of instrumentation! Getting the names into the list is a real killer.

> 
> It seems that perhaps this might be easily duplicated by mv'ing a single file to another directory, can you duplicate that on your test case? I also note a "dos622.all/" very early on in the inode listing, not sure whether that's the culprit here or if its just any FAT file move.

yes, any `mv` leaves the 'old' file's inode 'open', I'll take a look at the code tomorrow. The 'dos622.all' the source directory in the example .

Mellvik:
Ooopps, turns out this one was self inflicted. The two lines I recently commented out in `sys_link()` because I couldn't see their usefulness definitely had their purpose!

```
int sys_link(char *oldname, char *pathname)
{
#ifdef CONFIG_FS_RO
    return -EROFS;
#else
    struct inode *oldinode;
    int error;

    debug_file("LINK '%t' '%t'\n", oldname, pathname);
    error = namei(oldname, &oldinode, 0, 0);
    if (!error) {
        error = do_mknod(pathname, offsetof(struct inode_operations,link), (int)oldinode, 0);
        //if (error)     <----------------------
            //iput(oldinode);  <------------------------------
    }
    return error;
#endif
}
```
`do_mknod` sets an error code if the files are on different devices - or if there is no `link()` entry for the filesystem type.
It became obvious when `mv` between minix filesystems displayed the same problem. :-)

I should have seen that. Learning is sometimes a slow process.

ghaerr:
Oh, geez. Yes, you said in https://github.com/Mellvik/TLVC/pull/139 that these two lines eliminated the "trying to free free inode" message when mv'ing MINIX files (along with a system hang fix due to an L1 buffer not being released) - Does this mean then that the message sometimes occurs when mv'ing MINIX inodes, and we don't know the reason for that?

Do we know under what circumstances the "trying to free free inodes" message occurs?

I had committed this change in ELKS, I'll unfix it, thank you.

ghaerr:
> fixing the root cause only to find that it would be more expensive 

I would be interested in hearing what the actual root cause is, regardless of "expense". There are no other cases I know of where an inode->i_count is checked before calling iput(), ultimately that's just a cover-up.

Mellvik:
Let's keep some perspective on this: First, there is no bug, not even a problem. The message about freeing free inode is information, really a debug-message. Freeing a free inode is no more a problem than freeing a free buffer, which we're doing all the time. So the message may be deleted or better, encapsulated in a `debug_inode` macro so as to be available when needed.

IOW, what the 'check i_count' 'fix' does - instead of wholesale removal of the message - is to eliminate this particular case which is fine. We know the reason, we know it works, the message is not useful except possibly when changes are made. Whether the fix is smart or not is in the eyes of the beholder.

Changing code so that this 'double put' doesn't happen at all is probably not a big deal but requires a lot of testing – and fixes a non-existent problem.  (The root cause is that when `minix_link()` finds that the src is a directory, it sets EPERM and PUTs the inode and the parent inode before returning, then the former is PUTed again in `sys_link` where the 'fix' is). 

Further - this code - this particular functionality, renaming directories, is due for a significant update soon anyway, paving way for real (Linux/Unix style) intra-filesystem directory moves. A long time itch that is due now that I've spent time becoming familiar with the code.

ghaerr:
> Let's keep some perspective on this

Yes, at first I thought this whole issue left unhandled the -EPERM error from sys_link downstream to `mv`. Now that I have spent more time looking at this, I remembered that sys_rename uses sys_link to attempt a rename (the real root of this problem), and your request (and my fix) to make `mv` work back four years ago in https://github.com/ghaerr/elks/issues/583. Then came another "fix" for another FAT problem https://github.com/ghaerr/elks/pull/1824 where I added the `iput` and caused this incorrect "error message" in the first place. Frankly, I'd forgotten about all of this.

So this fix is essentially a continuation of making sys_rename work on FAT from ELKS 1824 with a bad side-effect for MINIX. Geez, sorry!

Speaking of kluges/hacks, looking further at do_mknod and how that function is being used to implement sys_link is pretty much a hack in itself, very messy and I'm sure that code's been in there for 30 years. do_mknod could definitely use a better name.

Thank you for the latest in figuring this all out. I'll add #142 into ELKS.

> this particular functionality, renaming directories, is due for a significant update soon anyway, paving way for real (Linux/Unix style) intra-filesystem directory moves.

Sounds great, but very tricky given the mv.c hacks and ongoing FAT and MINIX filesystem issues/failures associated around this design. We will definitely want a test suite to prove it works well, given how many problems we've added code to solve already!

Mellvik:
> > this particular functionality, renaming directories, is due for a significant update soon anyway, paving way for real (Linux/Unix style) intra-filesystem directory moves.
> 
> Sounds great, but very tricky given the mv.c hacks and ongoing FAT and MINIX filesystem issues/failures associated around this design. We will definitely want a test suite to prove it works well, given how many problems we've added code to solve already!

Certainly not a walk in the park, but the lack if it keeps haunting me almost daily when updating physical systems - like 1) tar up the `target/bin` directory, 2) `ftpput` it to tlvc, 3) (on tlvc) tar extract, 4) mv old bin to whatever, mv newbin into place, 5) delete olden and reboot. Next up would be adding devices to tar, that would really make life easy. 

BSD 2.11 ufs_rename() has really great comments btw.

ghaerr:
Wow, are you saying that your workflow requires copying each file in /bin to another directory, then copying back from the new directory, rather than just using a quick link-rename of the directories themselves? For a large directory like /bin, that sounds quite time-consuming.

There has been a need for updating an installed system in a better way for a long time. Have you found that you can't use ftpput to directly write /bin, because of the problem of running executables and their shared text segment? It would be nice to design a way around that entirely, to allow a running system to be updated quickly.

> Next up would be adding devices to tar, that would really make life easy.

You mean to allow block/char devices to be created? I might be missing what you're asking for, but agreed that tar could use some updates. Have you found it to be compatible with most host tar formats?

ghaerr:
> are you saying that your workflow requires copying each file in /bin to another directory

No need to answer that, I see that `mv` will work without copying if /bin doesn't contain symlinks or subdirectories. Did you remove ftpget, httpget, etc from /bin so this works?

I realize also that shared text is no issue since the inodes stay the same, only their name in the path hierarchy changes when /bin is renamed.

> BSD 2.11 ufs_rename() has really great comments btw.

Holy heck is that complicated! It generates some concern over how to prove correctness when the kernel rename function is implemented.

Mellvik:
> > are you saying that your workflow requires copying each file in /bin to another directory
> 
> No need to answer that, I see that `mv` will work without copying if /bin doesn't contain symlinks or subdirectories. Did you remove ftpget, httpget, etc from /bin so this works?

I did - wanting to look into replacing the symlinks with hard links which I never got to - I haven't checked whether that's even possible with the host tools. Also, while the ftpget etc. links are rarely used and thus less important, I'm using `uptime` quite a bit, so some intervention, manual or scripted is required.

The 'busy' (shared text) binaries continue to be a problem though, and I ended up doing the wholesale /bin replacement only when absolutely necessary. Scripted ftp doesn't provide for a delete command, so even with single file transfers (like this:
`f17='ftp -u ftp://xx:yy@10.0.2.17/linux target/linux'` it becomes a three-step process: transfer, rm old, mv new.

I don't see an easy (or complicated for that matter) way to fix this except getting `rename()` to work right, enabling the bnuilding of shadow directory trees that can be easily moved into place.

The interim solution, which is still incomplete in that 1) tar doesn't support devices and 2) devices aren't created on the development host anyway, is to jiggle between two partitions: Ftp the tar tree into the 'new' one, create /dev, copy the locally edited files - bootopts, mount.cfg, net.cfg, inittab etc. - and change the partition boot flag.

Slightly off topic: In order to simplify all this in an environment with many physical machines and even more Ethernet interfaces, I renamed the bootopts `net=` to `NET=` to make it propagate into the regular shell environment. that way `net start` doesn't require me to remember which NIC is currently the default (yah, I know, memory ...). This change has not been merged yet.

> 
> I realize also that shared text is no issue since the inodes stay the same, only their name in the path hierarchy changes when /bin is renamed.

It is indeed an issue as noted above, a real pain - not when moving directories but when replacing the binaries.

> 
> > BSD 2.11 ufs_rename() has really great comments btw.
> 
> Holy heck is that complicated! It generates some concern over how to prove correctness when the kernel rename function is implemented.

Agreed, but I think it looks worse than it is because our setting is much simpler. Many of the tests and considerations are not relevant here, and I suspect that we have some of the functionality in `namei` and `do_mknod` (a misnomer, I agree)  already. Anyway, development needs conditionals so before/after tests are easy.

Mellvik:
> 
> > Next up would be adding devices to tar, that would really make life easy.
> 
> You mean to allow block/char devices to be created? I might be missing what you're asking for, but agreed that tar could use some updates. Have you found it to be compatible with most host tar formats?

Yes, big linux and most recent unixen have device support in tar, extremely convenient, and IIRC a relatively easy fix. As to tar compatibility across platforms,  I cannot recall having problems. Then again,  the only systems I'm moving tar files from/to regularly are macOS and (big) Linux.

As noted before, since devices for TLVC and ELKS  cannot be created on the host, it's sort of falling on the side of this discussion. What would have been a really useful enhancement is a MAKEDEV script, which I believe I mentioned many years ago. BSD used to have that, a script that takes as a parameter which devices to create, such as 'std' for the basics, then devices for each controller (like hd*, fd, xd*, tty*, pty* etc where the asterisk is a number of such controllers to generate devices for). 

Having that available, a 'custom' dev directory could easily be created locally - the `sys` script could use it, and essentially shrink to nothing in size. (I do have a script that creates devices already, but it takes no parameters,  essentially just a cut from `sys`). I like the idea, and might take a stab at a TLVC MAKEDEV variant (which in BSD lives in the /dev directory btw).

```
#
#       @(#)MAKEDEV     4.27.6 (2.11BSD) 1999/2/19
#
# Device "make" file.  Valid arguments:
#       std     standard devices
#       local   configuration specific devices
#       fd      file descriptor driver
# Tapes:
#       ht*     unibus tu77 & te16
#       tm*     unibus tm11 & te10 emulations (e.g. Emulex tc-11)
#       ts*     unibus ts11
#       tu*     tmscp (tk50/tu81)
# Disks:
#       br*     unibus Eaton br1538 or br1711
#       hk*     unibus rk06 and rk07
#       ra*     unibus uda50 w/ ra??
#       rk*     unibus rk05
#       rl*     unibus rl01/rl02
#       rx*     unibus rx02
#       si*     unibus cdc 9766 w/ si 9500
#       xp*     unibus rm02/03/05, rp04/05/06, diva, eagle, cdc9766, fuji 160.
# Terminal multiplexors:
#       dl*     dl-11 units other than the console
#       dz*     unibus dz11 and dz32
#       dh*     unibus dh11 and emulations (e.g. Able dmax, Emulex cs-11)
#       dmf*    unibus dmf32
#       dhu*    unibus dhu11
#       dmz*    unibus dmz32
#       dhv*    qbus dhv11
# Pseudo terminals:
#       pty*    set of 16 master and slave pseudo terminals
# Printers:
#       lp*     unibus lp11 parallel interface
#       va*     unibus varian parallel interface
#       vp*     unibus versatec parallel interface
# Call units:
# Special purpose devices:
#       dr*             unibus dr11
#       ram*            memory disk
#       dn*             dn11 dialer
#       ingreslock      Ingres lock driver.

PATH=/etc:/sbin:/usr/sbin:/bin:/usr/bin
umask 77
for i
do
case $i in

std)
        mknod console   c 0 0
#       mknod kUmem     c 1 3   ; chmod 600 kUmem ;: untested
        mknod kmem              c 1 1   ; chmod 640 kmem ; chgrp kmem kmem
        mknod mem               c 1 0   ; chmod 640 mem ; chgrp kmem mem
        mknod null              c 1 2   ; chmod 666 null
        mknod zero              c 1 3   ; chmod 444 zero
        mknod tty               c 9 0   ; chmod 666 tty
        mknod klog              c 22 0  ; chmod 600 klog
        mknod errlog            c 22 1  ; chmod 600 errlog
        mknod acctlog           c 22 2  ; chmod 600 acctlog
        ;;

fd)
        umask 0
        rm -rf fd
        rm -f stdin stdout stderr
        mkdir fd
        chmod 755 fd
        mknod stdin c 26 0
        mknod stdout c 26 1
...
```

ghaerr:
> I did - wanting to look into replacing the symlinks with hard links which I never got to - I haven't checked whether that's even possible with the host tools. Also, while the ftpget etc. links are rarely used and thus less important, I'm using uptime quite a bit, so some intervention, manual or scripted is required.

I'm thinking of getting rid of ftpget, ftpput and httpget symlinks altogether, and recode urlget to allow, say, urlget -f for ftpget, etc. Given the kernel problem of disallowing a directory rename if symlinks are present, I think this is worth it.

Unfortunately, it is not possible to create hard links at this time, only symlinks. Hard links could be made possible with a mod to `mfs`, but IIRC there may be complications with this approach. Also of course, FAT doesn't support any links, so the files are duplicated, using even more valuable floppy space. For uptime, perhaps a `ps -u` option, or just copy another binary to uptime and forget the symlink?

> The 'busy' (shared text) binaries continue to be a problem though

> I don't see an easy (or complicated for that matter) way to fix this

One possibility, which is complicated but could work, would be a C program that would be copied (by the install script) from /bin/copysys to /tmp/copysys, and then executed by the shell script, replacing the shell binary using `exec /tmp/copysys`. The program would then go and kill all other running programs (except itself), and then proceed to copy /bin without any other shared text processes running from it. The complication here would be /bin/init, which cannot be killed since when restarting it wouldn't have PID 1. To fix that, /bin/init could be installed as /etc/init, which removes it from /bin. A bit complicated, but not really :)

ghaerr:
> Yes, big linux and most recent unixen have device support in tar, extremely convenient, and IIRC a relatively easy fix.

Adding mknod creat from tar extraction should be very easy - but getting the proper device numbers might be a little trickier: are you thinking of just getting then from target/dev, just before `mfs` etc creates the MINIX filesystem? That could work.

If going that route, a tar file could be created from target/, possibly with an --exclude=target/etc/init so that the above copysys would work later. That tar file would then be transferred over to the update system, and then the copysys or sysinst script run with that tar file as an argument, which would untar /bin etc directly after terminating other programs.

A complication here would be: does the host tar accept creating device files from target/dev?

Another idea would be to write a host shell script that creates a MAKEDEV at image build time that would create the /dev devices directly. This script could be run again after /bin is updated, instead of having /dev included in the tar file.

ghaerr:
Yet another idea would be to have a /bootopts option, or just have the kernel check at boot time, to check for a specially-named tar file to exist in /, say /update.tar. When that file is seen, instead of execing /bin/init, a separate program /etc/update is executed, which untars update.tar directly into /bin before the system even starts. After performing all the updates, the system reboots itself.

Sound familiar?!?! LOL!

Mellvik:
> > Yes, big linux and most recent unixen have device support in tar, extremely convenient, and IIRC a relatively easy fix.
> 
> Adding mknod creat from tar extraction should be very easy - but getting the proper device numbers might be a little trickier: are you thinking of just getting then from target/dev, just before `mfs` etc creates the MINIX filesystem? That could work.

I haven't thought that far about the ELKS/TLC special requirements. The first step is to implement general device tar/untar support in the TLVC/ELKS version of tar. Then to take a look at the tricks required to make it useful in this special case... probably making the are TLVC tar executable on the host and do whatever tricks we figure out are required/smart/whatever.

> 
> If going that route, a tar file could be created from target/, possibly with an --exclude=target/etc/init so that the above copysys would work later. That tar file would then be transferred over to the update system, and then the copysys or sysinst script run with that tar file as an argument, which would untar /bin etc directly after terminating other programs.
> 
> A complication here would be: does the host tar accept creating device files from target/dev?

That does sound like a way to go, let me think some more about this.

> 
> Another idea would be to write a host shell script that creates a MAKEDEV at image build time that would create the /dev devices directly. This script could be run again after /bin is updated, instead of having /dev included in the tar file.

Now we're talking! More about this later.
Thanks.

Mellvik:
@ghaerr,

> Another idea would be to write a host shell script that creates a MAKEDEV at image build time that would create the /dev devices directly. This script could be run again after /bin is updated, instead of having /dev included in the tar file.

This is actually a great idea. Thinking about it overnight, this is the way to go. Have the host image creation tool generate a script that goes into `target/dev` or `target/etc` (I prefer the first). When executed on the target it populates the current directory (presumably `/dev` but not necessarily) to become an exact copy of that in the corresponding image. 

My guess it's low threshold and solves the problem. Suggested name - keeping the BSD heritage - `MAKEDEV.local`, leaving `MAKEDEV` to the 'big brother' that I still intend to make (in BSD, MAKEDEV.local creates devices not covered by the general script, not exactly the same but `.local` indicates some specificity to the local system).

Mellvik:
> Yet another idea would be to have a /bootopts option, or just have the kernel check at boot time, to check for a specially-named tar file to exist in /, say /update.tar. When that file is seen, instead of execing /bin/init, a separate program /etc/update is executed, which untars update.tar directly into /bin before the system even starts. After performing all the updates, the system reboots itself.
> 
> Sound familiar?!?! LOL!

I've been playing with this idea on and off for quite some time, and ended up scrapping it because it always involved a shell - which would require a copy of the shell in say /etc. `sash` could possibly live there but it doesn't do scripting so it's useless for the purpose. BTW - just as frequently I've contemplated breaking the policy of having all executables in /bin, putting the system tools back into /etc where they came from (before someone invented /sbin). There are distinct advantages to that ... but I'm straying off topic.

I never though of installing a copy of `tar` (possibly limited, with some special features - like deleting the tar file and rebooting after etc) into /etc or whatever, and more importantly, set a flag at boot time to indicate 'do_update`. This idea is worth more discussion.

ghaerr:
> I never though of installing a copy of tar 

>  more importantly, set a flag at boot time to indicate 'do_update`

An important point to consider, which got me on this idea, is that we can't replace any running program due to shared text issues. This means no /bin/tar or /etc/tar, or /bin/init, etc. For maximum flexibility, the system will have to be updated before other programs are running - that solves the shared text dilemma. Of course - now I realize that's why even modern OSes do exactly the same thing - they boot up, specially apply an update, and reboot, very probably for the same reasons.

Here's more precisely how it could work:
- Two files are ftp'd over to the to-be-updated system: a /update.tar file, and a /update updater program
- The system is then rebooted using "shutdown -r"
- On boot, just before otherwise exec'ing /bin/init, the kernel checks for existence of "/update". If found, it runs it instead.
- /update runs, which has an internal tar capability, and untars /update.tar onto the filesystem. All and any programs from / on down will be immediately replaced with their updated version, with never an issue of shared text disallowing overwrites.
- /update then executes sys_sync,  deletes itself and calls sys_reboot system call.
- The system reboots normally, fully updated

A nice trick would be that the updating process on the host end would specially create the /update.tar file such that customized files in /etc, like /etc/passwd, /etc/hostname, etc would not be present, so they would not be updated. The included tar file would also contain /dev, so that those new devices would be created, without the need for a shell and MAKEDEV, etc.

The problem areas are that the tar/untar capability will need the ability to delete a directory (rm -rf style) in some cases. This would be for things like /dev, where old devices might need to be removed, or for when removing /bin programs rather than replacing them. This feature could use a special value for the mode bits and come later. The special file creating ability can come later.

To move forward, I would suggest taking a hard look at tar.c, and fixing it up a little so that it can be called as a function from a new update.c program. It will need the new capability of updating special files. Update.c can then do some stuff, then call the untar() function, then unlink itself, etc. The kernel mod to exec /update is trivial. Note that update has to reboot and can't return, since the kernel isn't waiting for it, it's in the idle loop, and /bin/init can't be exec'd anyways, since it needs PID 1.

The system can be protected by /update refusing to run if not PID 1, but during development, this would be turned off.

This will be a really great feature that is needed for some time!!!

ghaerr:
> While running programs cannot be overwritten, they can be deleted. So while the running program obviously keeps running on the inode in memory, traditional unix style, the new version can be written.

That's nice, but has the disadvantage of not reusing the original likely contiguous blocks of each of the replaced system file(s) initially created by `mfs` etc. When a file is just overwritten it will keep the same disk blocks (and inode numbers) it originally had, while after deletion and new create, the file disk blocks can be in any order, along with the inode position in the filesystem index, and will likely be from far across the disk surface because the free list likely starts there. The extensive startup testing I did around some of the disk cache experiments you were doing last October showed that the disk startup block allocations were extremely important in system startup speed. So I'm not seeing this approach as having any benefit over just rewriting from an updater when no other processes are running, which solves the shared text issue without this big disadvantage.

Mellvik:
It's a good point, I didn't think about that. That said, the discussion you refer to was about floppies, not hard disks. While file and inode placement will make a difference even on hard disks, in particular pre-IDE disks, I'm not convinced it is significant, on newer disks barely measurable. I believe the update tools/mechanisms we're discussing are very unlikely to be used on floppy drives.

IOW I think the tools we design should take advantage of this if it does indeed make things simpler.

PR #143, opened 2025-02-27
[kernel, bootopts] Change net= to NET= and more

Mellvik:
Changes the `bootopts` parameter `net=` to `NET=` in order to make it propagate in the user shell default environment. This simplifies the use of the `net start` command in setting where this command is used a lot.

Also changes the automatic turning off `fdcache` on 386 and later machines: Keep the setting in `bootopts` if `debug` is set.

Finally, this PR updates the boot device table in `init/main.c` (used to set `ROOTDEV`) to recognize MFM (`xd`) disks.

### DIFF
--- a/tlvc/init/main.c
+++ b/tlvc/init/main.c
@@ -108,6 +108,7 @@ static char * INITPROC root_dev_name(int dev);
 static int INITPROC parse_options(void);
 static void INITPROC finalize_options(void);
 static char * INITPROC option(char *s);
+static void add_env(char *);
 
 #endif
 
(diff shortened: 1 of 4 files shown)


ghaerr:
> Changes the bootopts parameter net= to NET= in order to make it propagate in the user shell default environment. 

Sounds like a good enhancement. As you know, this was originally designed to keep the network start information known only around /etc/rc.sys execution time, but I agree it's better to allow `net start` without any parameters to be used after boot as well. I'll make this change in ELKS also.

> Also changes the automatic turning off fdcache on 386 and later machines

I guess this is the culmination of all the testing from last October, but the automatic cache disable wasn't ever fully automated, until now? I'll have to check what the state of this is on ELKS.

Mellvik:
> 
> > Also changes the automatic turning off fdcache on 386 and later machines
> 
> I guess this is the culmination of all the testing from last October, but the automatic cache disable wasn't ever fully automated, until now? I'll have to check what the state of this is on ELKS.

Well, it has been around -like you say - since last fall. Then came the situation in which I needed to see the cache on/off for testing even on the 386, so fdcache=100 became a secret code for debugging. Unsmart and unneeded - so now debug= >0 does the trick.

The `Floppy cache ...` boot message will always appear when space has been allocated (menuconfig) to the cache - as a reminder. Otherwise a login and `meminfo` would be required to check.

PR #144, opened 2025-02-27
[cmds] Enhancements to /bin/sys, /bin/mount

Mellvik:
- `/bin/sys`: Updated to fully support `xd` devices
- `mount`: 
    - When a flat drive is specified (like `/dev/hda`), mount will now attempt to open the first partition on that device in order to avoid the chaos that some times ensues when erroneously mounting a partitioned device as flat.
    - When attempting to mount a filesystem with an unsupported file system type, like FAT when FAT support is not compiled in, a reasonable error message is displayed instead of the rather confusing (if technically correct) 'no such device'.

### DIFF
--- a/tlvccmd/sys_utils/mount.c
+++ b/tlvccmd/sys_utils/mount.c
@@ -14,6 +14,7 @@
 #include <stdlib.h>
 #include <errno.h>
 #include <ctype.h>
+#include <limits.h>
 #include <sys/mount.h>
 #include <sys/stat.h>
 #include <sys/types.h>
@@ -64,7 +65,7 @@ int main(int argc, char **argv)
 	int	flags = 0;
 	int	query = 0;
 	char	*option;
-	char	fsname[18];
+	char	fsname[PATH_MAX];
 
 	argc--;
 	argv++;
@@ -132,7 +133,7 @@ int main(int argc, char **argv)
 		return 1;
 	}
 
-	strncpy(fsname, argv[0], 18);
+	strncpy(fsname, argv[0], PATH_MAX-2);
 	if (isalpha(fsname[strlen(fsname)-1])) {	/* we're mounting a flat drive,
 							 * check that it really is flat */
 		strcat(fsname, "1");
@@ -144,7 +145,7 @@ int main(int argc, char **argv)
 	if (flags == 0 && type == 0)
 		flags = MS_AUTOMOUNT;
 	if (mount(argv[0], argv[1], type, flags) < 0) {
-		char failed[] = "mount failed";
+		char *failed = "mount failed";
 		if (flags & MS_AUTOMOUNT) {
 			type = (!type || type == FST_MINIX)? FST_MSDOS: FST_MINIX;
 			if (mount(argv[0], argv[1], type, flags) >= 0)
--- a/tlvccmd/sys_utils/mount.c
+++ b/tlvccmd/sys_utils/mount.c
@@ -12,8 +12,11 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <stdlib.h>
+#include <errno.h>
+#include <ctype.h>
 #include <sys/mount.h>
 #include <sys/stat.h>
+#include <sys/types.h>
 #include <linuxmt/fs.h>
 #include <linuxmt/limits.h>
 
@@ -51,16 +54,17 @@ static void show(void)
 static void usage(void)
 {
 	errmsg("usage: mount [-a][-q][-t minix|fat] [-o ro|remount,{rw|ro}] <device> <directory>\n");
-	return 1;
+	return;
 }
 
 int main(int argc, char **argv)
 {
 	char	*str;
-	int	type = 0;		/* default fs*/
+	int	type = 0;		/* default fs */
 	int	flags = 0;
 	int	query = 0;
 	char	*option;
+	char	fsname[18];
 
 	argc--;
 	argv++;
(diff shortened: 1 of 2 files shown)


tlvccmd/sys_utils/mount.c
```diff
@@ -128,16 +132,28 @@ int main(int argc, char **argv)
...
+		strcat(fsname, "1");
+		if (open(fsname, O_RDWR) != -1) {
+			printf("Cannot mount %s: Device is partitioned\n", argv[0]);
+			return 1;
+		}
+	}
 	if (flags == 0 && type == 0)
 		flags = MS_AUTOMOUNT;
 	if (mount(argv[0], argv[1], type, flags) < 0) {
+		char failed[] = "mount failed";
```
ghaerr, line 147: Looks like you're not aware how much code is generated, and data duplicated, when a string array is declared and initialized as auto on the stack. Essentially what happens is that the array string initializer is actually allocated in the .data segment, then the stack is grown down and a `memcpy` executed (similar to a full structure copy), to initialize the same string as an auto variable. Complete waste of stack and data space, along with additional code.
Here's an example:
```
int main(int ac, char **av)
{       
    char string[] = "this is a test";

    printf("hello %s!\n", string);
}       
```
ia16-elf-gcc code:
```
.LC1:
        .string "hello %s!\n"
.LC0:
        .string "this is a test"
        .section        .text.startup,"ax",@progbits
        .global main
        .type   main, @function
main:
        pushw   %si
        pushw   %bp
        movw    %sp,    %bp
        subw    $16,    %sp
        leaw    -15(%bp),       %si
        movw    $15,    %ax
        pushw   %ax
        movw    $.LC0,  %ax
        pushw   %ax
        pushw   %si
        call    memcpy
        pushw   %si
        movw    $.LC1,  %ax
        pushw   %ax
        call    printf
        xorw    %ax,    %ax
        movw    %bp,    %sp
        popw    %bp
        popw    %si
        ret
```
Changing `char string[] = ` to `char *string = ` eliminates the memcpy, but duplicate strings are merged anyways by the linker to a single .data location - thereby rendering any special handling at all as completely unneeded!

tlvccmd/sys_utils/mount.c
```diff
@@ -128,16 +132,28 @@ int main(int argc, char **argv)
 		return 1;
 	}
 
+	strncpy(fsname, argv[0], 18);
```
ghaerr, line 135: The max size passed to strncpy must be one less than the buffer in order to ensure the buffer gets NUL-terminated.

In addition, the strcpy "1" below would require a strncpy max size two less than the buffer size in order to prohibit buffer overflow even after a successful strncpy, with a max-1 length argument.

Chopping off argv[0] at an arbitrary length could result in undefined behaviour should a path somehow exceed 18. In this case, IMO best to just use `char fsname[PATH_MAX]` as the entire system is built around that limit, as far as unwanted undefined behavior is concerned.

tlvccmd/sys_utils/mount.c
Mellvik: Thanks, and no, I wasn't aware! Very interesting, learning every day. Appreciated.

tlvccmd/sys_utils/mount.c
Mellvik: Thanks - that's 3 in a row. Turns out even my original thinking, maxname+'/dev/', was one short. Appreciated.

tlvccmd/sys_utils/mount.c
Mellvik: @ghaerr - thanks for pointing out that `char string[]` is different from `char *string`. I just googled that - quite important! I may have known at some point - totally gone now!

tlvccmd/sys_utils/mount.c
```diff
@@ -128,16 +133,28 @@ int main(int argc, char **argv)
...
+		strcat(fsname, "1");
+		if (open(fsname, O_RDWR) != -1) {
+			printf("Cannot mount %s: Device is partitioned\n", argv[0]);
+			return 1;
+		}
+	}
 	if (flags == 0 && type == 0)
 		flags = MS_AUTOMOUNT;
 	if (mount(argv[0], argv[1], type, flags) < 0) {
+		char *failed = "mount failed";
```
ghaerr, line 148: This is definitely better, but did you get my point about the fact that the ld.gold linker "merges" all strings that are exactly the same into a single address/allocation? This means that there is no benefit at all to what I am guessing is an attempt at saving string space: it doesn't, while also make the program harder to read (and also generates more code, unless the address happens to be kept in a register, which is unlikely given the number of auto variables in this function).

If this were rewritten to use a string literal in each place, the same string literal would also want to be passed as the 2nd argument (as is done now) in line 155, in order to keep all 3 of 3 strings merged.

tlvccmd/sys_utils/mount.c
```diff
@@ -12,8 +12,12 @@
...
 #include <fcntl.h>
 #include <stdlib.h>
+#include <errno.h>
+#include <ctype.h>
+#include <limits.h>
 #include <sys/mount.h>
 #include <sys/stat.h>
+#include <sys/types.h>
 #include <linuxmt/fs.h>
 #include <linuxmt/limits.h>
```
ghaerr, line 22: FYI: <limits.h> above includes <linuxmt/limits.h>, so one less explicit kernel include needed here.

Also, <sys/types.h> is included by <sys/stat.h>, <sys/mount.h> and almost every other include, so it's not needed here.

By all rights, FST_MINIX etc should also be available through the C library includes somehow, but we haven't got around to that yet, thus necessitating <linuxmt/fs.h>. I'm not sure yet how to write portable system programs like mount.

tlvccmd/sys_utils/mount.c
Mellvik: I vividly remember the discussion we had years ago re. network drivers and printk messages, which ended up in a header file with macros for common messages. gcc's smartness about strings was unknown to me before that. 

So, yes, I'm aware - and in this case it's a conscious choice.
`printf("%s: Filesystem type not available\n", failed);` explicitly enables the string to be shared,
`printf("Mount failed: Filesystem type not available\n", failed);` does not - I don't think the linker splits strings, does it? And
`printf("%s: Filesystem type not available\n", "Mount failed");` looks weird but does what you're looking for - if I understand you correctly. Readability is a valid point, but then only the 2. alternative is really readable - and botches the desired optimization.

tlvccmd/sys_utils/mount.c
Mellvik: Thanks! Every little bit ...

PR #145, opened 2025-03-02
[cmd] Added device store/extract capability to tar (#141 comment)

Mellvik:
One of many items discussed in https://github.com/Mellvik/TLVC/pull/141#issuecomment-2687399735 - `tar` can now save and restore devices. 

More testing needed, but it's a relatively simple addition.

### DIFF
--- a/tlvccmd/misc_utils/tar.c
+++ b/tlvccmd/misc_utils/tar.c
@@ -61,6 +61,7 @@ daddr_t	lookup();
 #define TBLOCK	512
 #define NBLOCK	20
 #define NAMSIZ	100
+
 union hblock {
 	char dummy[TBLOCK];
 	struct header {
@@ -71,10 +72,29 @@ union hblock {
 		char size[12];
 		char mtime[12];
 		char chksum[8];
-		char linkflag;
+		char typeflag;
 		char linkname[NAMSIZ];
+		char magic[6];                /* 257 */
+		char version[2];              /* 263 */
+		char uname[32];               /* 265 */
+		char gname[32];               /* 297 */
+		char devmajor[8];             /* 329 */
+		char devminor[8];             /* 337 */
+		char prefix[155];             /* 345 */
+		                              /* 500 */
 	} dbuf;
 } dblock, tbuf[NBLOCK];
+
+/* Values used in typeflag field.  */
+#define REGTYPE  '0'            /* regular file */
+#define AREGTYPE '\0'           /* regular file */
+#define LNKTYPE  '1'            /* link */
+#define SYMTYPE  '2'            /* reserved */
+#define CHRTYPE  '3'            /* character special */
+#define BLKTYPE  '4'            /* block special */
+#define DIRTYPE  '5'            /* directory */
+#define FIFOTYPE '6'            /* FIFO special */
+#define CONTTYPE '7'            /* reserved */
 
 struct linkbuf {
 	ino_t	inum;
@@ -90,6 +110,7 @@ int	rflag, xflag, vflag, tflag, mt, cflag, mflag;
 int	term, chksum, wflag, recno, first, linkerrok;
 int	hflag, oflag, pflag;
 int	nblock = 1;
+int	curtype;
 
 daddr_t	low;
 daddr_t	high;
@@ -141,13 +162,14 @@ int	m8[] = { 1, WOTH, 'w', '-' };
 int	m9[] = { 2, STXT, 't', XOTH, 'x', '-' };
 
 int	*m[] = { m1, m2, m3, m4, m5, m6, m7, m8, m9};
+char	ftype[] = { '-', '-', 'l', 'c', 'b', 'd', 'f', '-' };
 
-void
-pmode(st)
+void pmode(st)
 register struct stat *st;
 {
 	register int **mp;
 
+	printf("%c", curtype ? ftype[curtype] : '-');
 	for (mp = &m[0]; mp < &m[9];)
 		doselect(*mp++, st);
 }
@@ -165,15 +187,19 @@ usage()
 	done(1);
 }
 
-void
-longt(st)
+void longt(st)
 register struct stat *st;
 {
 	register char *cp;
+	char dev[8];
 
 	pmode(st);
 	printf("%3d/%1d", st->st_uid, st->st_gid);
-	printf("%7ld", st->st_size);
+	if ((st->st_mode & S_IFCHR) + (st->st_mode & S_IFBLK)) {
+		sprintf(dev, "%d,%d", st->st_rdev>>8, (st->st_rdev&0xff));
+		printf("%7s", dev);
+	} else
+		printf("%7ld", st->st_size);
 	cp = ctime(&st->st_mtime);
 	printf(" %-12.12s %-4.4s ", cp+4, cp+20);
 }
@@ -205,8 +231,7 @@ char *name;
 	return(1);
 }
 
-checkupdate(arg)
-char	*arg;
+checkupdate(char *arg)
 {
 	long	mtime;
 	daddr_t seekp;
@@ -214,7 +239,7 @@ char	*arg;
 
 	fseek(tfile, 0L, SEEK_SET);
 	for (;;) {
-		if ((seekp = lookup(arg)) < 0)
+		if ((long)(seekp = lookup(arg)) < 0)
 			return(1);
 		fseek(tfile, seekp, 0);
 		fscanf(tfile, "%s %lo", name, &mtime);
@@ -360,13 +385,16 @@ checksum()
 	return(i);
 }
 
-void
-getdir()
+/* Get a directory entry (1 block) from the tape, fill the global stbuf 
+ * (struct stat buffer) from the contents
+ */
+void getdir()
 {
 	register struct stat *sp;
 	int i;
+	char typeflag;
 
-	readtape( (char *) &dblock);
+	readtape((char *) &dblock);
 	if (dblock.dbuf.name[0] == '\0')
 		return;
 	sp = &stbuf;
@@ -379,6 +407,20 @@ getdir()
 	sscanf(dblock.dbuf.size, "%lo", &sp->st_size);
 	sscanf(dblock.dbuf.mtime, "%lo", &sp->st_mtime);
 	sscanf(dblock.dbuf.chksum, "%o", &chksum);
+
+	/* typeflag may be 0 or from '0' and up ... */
+	typeflag = dblock.dbuf.typeflag;
+	curtype = typeflag ? typeflag  - '0' : 0;
+	if (typeflag == CHRTYPE || typeflag == BLKTYPE) {
+		sscanf(dblock.dbuf.devmajor, "%8d", &i);
+		sp->st_rdev = (i << 8);
+		sscanf(dblock.dbuf.devminor, "%8d", &i);
+		sp->st_rdev |= i ;
+		if (typeflag == CHRTYPE)
+			sp->st_mode |= S_IFCHR;
+		else
+			sp->st_mode |= S_IFBLK;
+	}
 	if (chksum != checksum()) {
 		fprintf(stderr, "directory checksum error\n");
 		done(2);
@@ -393,7 +435,7 @@ passtape()
 	long blocks;
 	char buf[TBLOCK];
 
-	if (dblock.dbuf.linkflag == '1')
+	if (dblock.dbuf.typeflag == LNKTYPE)
 		return;
 	blocks = stbuf.st_size;
 	blocks += TBLOCK-1;
@@ -500,6 +542,7 @@ putfile(longname, shortname, parent)
 			}
 			stbuf.st_size = 0;
 			tomodes(&stbuf);
+			dblock.dbuf.typeflag = DIRTYPE;
 			strcpy(dblock.dbuf.name,buf);
 			(void)sprintf(dblock.dbuf.chksum, "%6o", checksum());
 			(void) writetape((char *)&dblock);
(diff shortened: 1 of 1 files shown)


ghaerr:
Nice. Will this create char/block special files created from using macOS tar?

Mellvik:
It will. I haven't tested that until you asked, assuming creating devices on macOS is impossible given the systems' extreme security measures. I was wrong. With root privileges it seems you can actually create the TLVC /dev directory on the mac, ready for tar'ing and transfer. Need to test that tomorrow.

ghaerr:
I was actually asking about the reverse - the ability for macOS to use its tar to create a .tar file from an image created using mfs or other tools, and then have that tar file be untar'd on TLVC. I'm not sure what usefulness there is in creating TLVC-numbered special files on macOS from a TLVC system, unless I guess you're archiving TLVC systems for some reason. I had thought the purpose of this enhancement was for untar'ing device files created from the host to be created on TLVC, and am concerned about the tar format and whether it works with host-based tar source files. I assume the "linkflag/typeflag" values are the same for .tar files from macOS tar also?

Mellvik:
It seems we're coming from different places. I'm describing exactly what you describe as the purpose of this enhancement... ?
Tar is always cross platform compatible for files and directories, for *ix systems other types work fine too - if supported by tar itself. Whatever you create with macos tar and no options works fine back and forth to/from tlvc. Unsupported types will usually be reported and ignored.

Maybe we're still talking about different things? 

One of the purposes of this enhancement was to be able to create a full TLVC filesystem on the macos host - to be ftp'd over and dumped into a fresh filesystem. Now that seems possible.

Mellvik:
> TLVC. I'm not sure what usefulness there is in creating TLVC-numbered special files on macOS  *from* a TLVC system, 

Not from, but 'on' is the keyword. For transfer TO TLVC.

Mellvik:
The final commit does what newer versions of `GNU tar` does, disallows absolute pathnames by eliminating the first '/' if it's there. Saves many a catastrophe in un-taring...

ghaerr:
> Tar is always cross platform compatible for files and directories, for *ix systems other types work fine too

My questions were leading in this direction, I wasn't sure that the tar header format is in fact compatible across all versions of UNIX, Linux, macOS etc. It would seem someone might have changed the header format over all these years such that some tar's were incompatible with others, making cross platform usage incompatible. Nice to know tar is fully compatible across all systems.

Mellvik:
`tar` has been 'enhanced' any number of times over the years and there are many versions - identified by the `magic[]` field in the header. However, the original tar header (512bytes) had ample space for accommodating extensions, so it's still backwards compatible for the basic content types, two of which are CHRDEV and BLKDEV. Of course there are issues that require checking the version, like how sparse files are saved and restored, but we don't deal with that.

To this day, `tar` remains one of my most used commands, I keep forgetting what an oddball I am. Still, there is no safer and more effective way to copy directory hierarchies between filesystems or even on the same file system - possibly with the exception of `cpio -p`.

Now that devices are supported, duplicating a partition on TLVC or ELKS can take advantage of  this example which I believe come for the V7 tar man page:
```
# cd /
# mount /dev/hda2 /mnt
# tar cf - . |(cd /mnt; tar xvfp -)
# umount /mnt
```
`tar` will not traverse mount points (which reminds me I haven't checked that yet).

ghaerr:
Thanks for the explanation @Mellvik. I see now - of course, with a magic field various versions can be supported, depending on compatibility needs, but in general tar works across systems using a basic header. Nice! Agreed tar will be very useful as part of the faster upgrade capability we've been talking about. (And IMO much easier and smaller than cpio).

> tar cf - . |(cd /mnt; tar xvfp -)

That will be interesting to test. IIRC at the moment the pipe buffer is ~80 bytes, but that may or may not make much different in extraction speed.

ghaerr:
> tar will not traverse mount points

I'm trying to think how to programmatically determine that (without messy interpretation of `mount` command output). Does ELKS/TVLC tar do that already?

ghaerr:
One more thought on the eventual problem of using tar as the mechanism for updating the system: we need the ability to delete a file or directory sometimes, when a system file is renamed or deleted. Can you see a way where tar could do this?

I suppose its not a big deal, but if we eventually choose to just "rm -rf" various directories, that could remove files which have been customized and we don't want deleted. But we might want to have a mechanism where tar could specify that a directory should be cleared before further untaring (e.g. using path/to/dir/with/trailing/ and a special mode bit perhaps, etc).

Another way to solve this might be to try to move files that will be customized to special directories; but this is a pain even for /etc, meaning portions of /etc would stay for system updating, and /etc/passwd would have to be moved. Other thoughts would be to use special mode bits on the destination, or even comparing text (non-binary) files before overwriting, with a special option, or of course having the update.tar creator have "full" and "partial" update file lists. 

More food for thought on how this mechanism might be used as the core of the update process.

Mellvik:
> One more thought on the eventual problem of using tar as the mechanism for updating the system: we need the ability to delete a file or directory sometimes, when a system file is renamed or deleted. Can you see a way where tar could do this?

No I don't. This would be 'synchronization' as opposed to save and restore. I suspect Minix `syncdir`, possibly `remsync` may be worthwhile looking onto for this.

> 
> I suppose its not a big deal, but if we eventually choose to just "rm -rf" various directories, that could remove files which have been customized and we don't want deleted. But we might want to have a mechanism where tar could specify that a directory should be cleared before further untaring (e.g. using path/to/dir/with/trailing/ and a special mode bit perhaps, etc).

There may be options to facilitate this in newer tar version, what would be the application of this?

Many, even older tars have the option to provide an exclude list, a list for files either not to save or not to exctract. That's a relatively easy thing to add. Re below.

>
> 
> Another way to solve this might be to try to move files that will be customized to special directories; but this is a pain even for /etc, meaning portions of /etc would stay for system updating, and /etc/passwd would have to be moved. Other thoughts would be to use special mode bits on the destination, or even comparing text (non-binary) files before overwriting, with a special option, or of course having the update.tar creator have "full" and "partial" update file lists.

Also, many tars can check dates before restoring. We don't have that - it may be implemented, but I'm not sure it's useful. Many also have a diff option, bnut even that is hard to apply to the present case. It seems include/exclude lists may be best.

> 
> More food for thought on how this mechanism might be used as the core of the update process.

Mellvik:
> > tar will not traverse mount points
> 
> I'm trying to think how to programmatically determine that (without messy interpretation of `mount` command output). Does ELKS/TVLC tar do that already?

I looked briefly at this, it does not. I suspect the programs doing this (cpio, pax ...) consult `/etc/mtab` to find mount points. In our case using an exclude list is the easiest for now. I dont think there is a `statfs` that provides such info and it would have to be applied to every single directory ...

ghaerr:
> There may be options to facilitate this in newer tar version, what would be the application of this?

> This would be 'synchronization' as opposed to save and restore.

The idea was more for keeping synchronization between the installed files and files that were renamed and/or deleted in a subsequent installation. I'm not sure that comprehensive "updating" of an OS is only concerned with save/restore, when updating revised files, it might want to know about changed/renamed files. But as I think of it, this entails even more computation and history on the creating side, so I suppose this is something that can be worried about later, after a basic update capability has been established.

> I suspect the programs doing this (cpio, pax ...) consult /etc/mtab to find mount points. In our case using an exclude list is the easiest for now.

Agreed. And since most host tar's have --exclude-files or the like already for create, this may not be an option that's needed on the extract side (except for possibly "mnt/...", but that could be excluded on the create side as well). When creating an "update.tar", there could be options available for which kinds of (or specific) files should be included (excluded).

Mellvik:
> Agreed. And since most host tar's have --exclude-files or the like already for create, this may not be an option that's needed on the extract side (except for possibly "mnt/...", but that could be excluded on the create side as well). When creating an "update.tar", there could be options available for which kinds of (or specific) files should be included (excluded).

Yes. I think the heavy lifting needs to be done on the host side where the tools are much more powerful . I'm adding the exclude option to tar and leave it for now. If the exclude option is applied to file extraction, it would be useful for protecting locally modified files: Keeping a file with a list of exceptions.

I also plan to experiment more with `syncdir` which looked really interesting when I tested it last week, in order to understand its potential completely. A possible scenario would be to untar a complete system into a shadow root on the target, then sync that to the real root and delete the shadow.

Mellvik:
@ghaerr, can the host tools create hard links? If yes, we should change the symlinks in /bin to hard links to make life easier on `mv`.

Mellvik:
Honestly, since I go the ftpd server stable on elks - that's a long time - I have barely used `ftpget` in any variant. It's simply much more convenient to go the other way, not the least because `ftp`-servers are endangered species these Fays for security reasons. So I would probably not miss them.  `uptime` is a different story.

What I've had on my list recently is simply to remove the symlink restriction from `mv`.  I don't know there it came from, but ss I see it, it's not useful and no other system I know of have this restriction. Dead links are all over the place and the user's responsibility. So as far as minix fs goes, that solves the immediate problem.

ghaerr:
>  simply to remove the symlink restriction from mv. I don't know there it came from

Well, I wrote all of the mv enhancements. `mv` needs to empty the directory in order to delete it after the "move" (rename). The way it does this is by creating a new directory, then link each file to the new directory, then unlink the file from the old directory, then delete the old directory. The `link` system call fails on directories and symlinks. That's why `mv` fails to rename a directory when there are symlinks or subdirectories.

To make this work w/symlinks, you'll need to either delete the symlinks (not recommended as they might be valuable to the user), or recreate the symlinks on the destination directory. I didn't do that in my initial implementation years ago, just like I didn't write a way to recursively handle subdirectories. The latter sounds like it could get complicated, so much so that it finally matches the complexity of just getting a proper kernel directory rename to work in the first place.

Thus my current recommendation to, for the time being, just don't produce the symlinks in the /bin directory in the first place. However, an enhancement to create symlinks on a directory move isn't a bad idea, and shouldn't be a big deal to implement (and the links would remain working).

Mellvik:
OK, I get that - and my thinking, possibly erroneous - was that a 'simple' directory mv, no subdirs, same fs, should be only slightly more complicated than a file mv (times 3), more or less a first step towards the real thing. You've been there, I have not - except for the recent rabbit hunts.

ghaerr:
> was that a 'simple' directory mv, no subdirs, same fs, should be only slightly more complicated than a file mv

Sorry, I'm not really following you here: are you talking about a user-space "mv"` using mv.c, or a kernel "mv". 

In the first case, a user space mv will fail if it can't delete the original directory.

In the second case, a kernel mv, here is no kernel mv, only link/unlink, and link fails on directories. To implement a kernel rename (which isn't quite mv since all kernel renames and/or mv's would alway have to be on the same fs), I would think the following would need to happen, which is IMO much more complicated than a link/unlink "file mv":
- namei/diri parent of original directory
- enter kernel lock. Needed because although kernel is not reentrant, any ensuing disk I/O could cause requesting process to sleep and same routines reentered, possibly using same inode as above
- create new directory inode, release original on fail
- write new code to copy original directory entry from original directory entry to new entry in new directory 
- remove original directory entry in parent inode
- delete original directory inode
- release lock

Sounds complicated, I'll need to think through this more to be sure.

Mellvik:
Yes, I know all this, we discussed it a few days ago - including looking at the BSD code for `rename()`.

My point is something else: `mv dira dirb` works already and does the right thing, essentially  - if there are no subdirectories or symlinks. There is no reason to block this operation because of symlinks - they could be treated like plain files.

ghaerr:
> There is no reason to block this operation because of symlinks - they could be treated like plain files.

I see. So the kernel should treat symlinks like plain files, with regards to hard links. That is, apparently now symlinks will always have a link count of 1 and can't be linked to. Is this consistent with the way Linux and other UN*X handles symlinks?

We will have to look at the kernel to see if what other, if any, restrictions are imposed on symlinks with regards to link/unlink system calls.

Mellvik:
Yes, in this particular situation a symlink would be just another plain file. Like all the other files in the directory it would not be interpreted anyway, technically just an inode update.

Mellvik:
The last commit fixes the 'disallow absolute pathnames', which didn't initially work if the root directory only was specified on the command line. It also add a `-X name` option telling `tar`to skip the given file or directory while building the tar file.

The `-X` implementation is preliminary, needs to be a reference to a file containing the names of files and directories to exclude, one per line. It allows us to create a complete TLVC file system tar file from a running system. An `exclude` option for extraction is also needed, to avoid overwriting locally modified files.

ghaerr:
> in this particular situation a symlink would be just another plain file. Like all the other files in the directory it would not be interpreted anyway

I looked at the kernel code for sys_link, which as you know is used by mv.c to link each file to the new directory before deleting the old directory (with regards to the scenario we've been discussing). Ultimately the actual problem is different than I thought and has to do with paths containing symlinks get followed in almost every system call except lstat.

> So the kernel should treat symlinks like plain files, with regards to hard links. That is, apparently now symlinks will always have a link count of 1 and can't be linked to.

There doesn't seem to be any restriction on the business of symlinks being able to be hard-linked to. Instead, what seems to be happening is that sys_link calls namei() to traverse the passed path to return a pointer to an inode - but when namei runs into a symlink, it follows the symlink, instead of return the symlink inode itself. And this is seemingly why ultimately sys_link is failing in mv.c. I haven't got into the full details of exactly when it fails, but bottom line is that, in the sys_link case, symlinks passed to link are followed, and that ultimately will result in the directory being emptied to continue to contain entries (either symlinks remaining or new links from following previous symlinks calling sys_link).

This situation has required various system calls to have "non-follow" versions, e.g. stat, lstat and fstatat, and many more on modern kernels with hundreds of system calls. There's non-follow versions of link and unlink called linkat which takes a flag argument (AT_SYMLINK_FOLLOW), and unlinkat, which take a flags argument (AT_REMOVEDIR), both of which might be needed in order to get mv.c to work as discussed.

Is it worth adding more system calls to handle this case for link/unlink in mv.c? IMO I think not. Far simpler to just remove symlinks from the distribution /bin directory, or perhaps, wait for a real kernel rename function.

Mellvik:
> Is it worth adding more system calls to handle this case for link/unlink in mv.c? IMO I think not. Far simpler to just remove symlinks from the distribution /bin directory, or perhaps, wait for a real kernel rename function.

It seems to me such functionality (and thus system calls) are required for any application that traverses the fs hierarchy an has (or needs) an option to not follow links - cp, tar, find, ...

ghaerr:
> such functionality (and thus system calls) are required for any application that traverses the fs hierarchy an has (or needs) an option to not follow links - cp, tar, find, ...

No, no extra system calls are required to traverse a filesystem hierarchy: on ELKS, a directory tree is traversed using `opendir`, then `readdir` is used to read directory entries. `readdir` always returns the actual directory entry name, never the followed symlink. In order to determine whether to follow/not follow any particular "directory", `lstat` is then used on the name returned from readdir, where it can be tested for S_IFDIR or S_IFLNK (symlink). S_IFDIR will be set only for real directories, S_IFLINK will be set for symlinks. To follow the symlink anyways, call `opendir` on the name (which isn't an S_IFDIR) and the symlink will be followed and work correctly if it points to a directory, just like it is on `open`. This is essentially the method `cp -R` uses now, except that it never follows symlinks.

Thus, the default directory recurse method in ELKS is to not follow symlinks, and no new system calls are required to follow symlinks, if desired.

> What I've had on my list recently is simply to remove the symlink restriction from mv.

> There's non-follow versions of link and unlink called linkat which takes a flag argument (AT_SYMLINK_FOLLOW), and unlinkat, which take a flags argument (AT_REMOVEDIR)

Thanks for your comments, as I now realize after thinking through the way opendir/readdir works, that I was incorrect saying that `unlink` follows symlinks and we need a non-follow version `unlinkat`. That's incorrect: `unlink` never follows a symlink, it removes the symlink itself. The problem in mv.c isn't the link/unlink used for the rename, but just that `link` fails on symlinks. And we don't need any new system calls to create another symlink, we just can't use `link` to do it.

So (for the 3rd time, trying to get this correct): mv.c can in fact be fairly easily enhanced to allowing renaming directories, with no new system calls, even with symlinks in them. In the function linkfiles(), if link() fails, the source should be `lstat`ed, and if its a symlink, call `readlink` to get the symlink value, then `symlink` to create a new identical symlink in the destination directory. Then call `unlink` on the source, as it already does.

Sorry for my confusion. Getting system calls figured out correctly on filesystems is tricky, yet another reason I'm hesitant to add new functionality without really thinking things through.

Mellvik:
Thanks @ghaerr, that's what I call a fruitful discussion.

Who fixes this - you or I?

ghaerr:
It's all fixed for you. I'm still planning on removing the symlinks from /bin, especially since on FAT filesystems, the links get converted to file copies, and urlget is 11k, which gets copied three additional times, for a program that is barely used!!

Mellvik:
Thinking about `urlget`, it may be an idea to cut history, rename it to `curl`, get rid of the links and change the options to match `curl`. Very limited functionality compared to 'real' `curl`, but it may be considered a start, a work in progress project.

Re. FAT file system and copies instead of links, I'm probably missing something since I don't see the problem. A FAT based boot floppy, which is the only case in which space is an issue, seems nonsensical. Why have a 'bad' boot floppy (image format) when you can have a good one? What I'm missing is the case where a FAT boot floppy is actually useful or beneficial in some way.

Issue #156, opened 2025-03-22
Build error with gcc 14


rafael2k:
When trying to build TLVC by running "build.sh", I get:
```
gcc -DHAVE_CONFIG_H -I. -I../../binutils-src/gas  -I. -I../../binutils-src/gas -I../bfd -I../../binutils-src/gas/config -I../../binutils-src/gas/../include -I../../binutils-src/gas/.. -I../../binutils-src/gas/../bfd -DLOCALEDIR="\"/home/rafael2k/programs/devel/TLVC/cross/share/locale\""  -W -Wall -Wstrict-prototypes -Wmissing-prototypes -Wshadow -Wstack-usage=262144 -Werror -Wwrite-strings -I../../binutils-src/gas/../zlib -g -O2 -MT write.o -MD -MP -MF $depbase.Tpo -c -o write.o ../../binutils-src/gas/write.c &&\
mv -f $depbase.Tpo $depbase.Po
../../binutils-src/gas/write.c: In function ‘chain_frchains_together’:
../../binutils-src/gas/write.c:426:7: error: dangling pointer ‘prev_frag’ to ‘dummy’ may be used [-Werror=dangling-pointer=]
  425 |     info->frchainP->frch_last
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~
  426 |       = chain_frchains_together_1 (section, info->frchainP);
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../binutils-src/gas/write.c:391:9: note: ‘dummy’ declared here
  391 |   fragS dummy, *prev_frag = &dummy;
      |         ^~~~~
cc1: all warnings being treated as errors
make[5]: *** [Makefile:1224: write.o] Error 1
make[5]: Leaving directory '/home/rafael2k/programs/devel/TLVC/cross/build/binutils-build/gas'
make[4]: *** [Makefile:1269: all-recursive] Error 1
make[4]: Leaving directory '/home/rafael2k/programs/devel/TLVC/cross/build/binutils-build/gas'
make[3]: *** [Makefile:814: all] Error 2
make[3]: Leaving directory '/home/rafael2k/programs/devel/TLVC/cross/build/binutils-build/gas'
make[2]: *** [Makefile:4864: all-gas] Error 2
make[2]: Leaving directory '/home/rafael2k/programs/devel/TLVC/cross/build/binutils-build'
[...]

PR #157, opened 2025-03-24
Fix compilation problems

Mellvik:
Fix compilation and build problems reported by @rafael2k in #156.


rafael2k:
Thanks @Mellvik. Today I'll boot up TLVC with IDE_DEBUG.

PR #159, opened 2025-03-24
[kernel] Add support for loading kernel in HMA

Mellvik:
This update expands on the Limited HMA Support introduced on #155 and enables the TLVC kernel to run in HMA when available. 

The problem (challenge) with HMA is that the segment register cannot change. The entire 64k block must be utilised by something that adheres to this limitation. Thus the first candidate, the external buffer headers, which when maxed pout with XMS buffers occupies a full segment. The next candidate is the kernel - bulging at the seams these days and easily occupying a full segment.

Given that 2975 buffers isn't really useful and that a 10th of that is more like it, kernel HMA would free up more application memory. Thus thus this capability, which indeed does improve the memory situation significantly. There are some caveats - and things to be aware of.

- Kernel in HMA cannot be selected in /bootopts for obvious reasons. So if compiled with `CONFIG_HMA_KERNEL`, setup will load the kernel high if possible. 
- 'Possible' means that the A20 line can be reliably turned on, which required the a20 handlers from `unreal.S` to be copied into `setup.S`.
- If the system cannot use unreal mode to access XMS, a HMA kernel prevents XMS buffers from being used: XMS via the BIOS turns off A20 after every access and the kernel is suddenly out of reach. So in such situations it's a choice - what's most beneficial, XMS buffers (and HMA bufheads) or HMA kernel?
- Setup will detect if XMS or A20 support is not available and load the kernel normally
- Likewise `buffer_init()` will check for a HMA loaded kernel before attempting to use high bufheads
- BIOS A20 manipulation turns out to be a rare thing, thus the code from @ghaerr's XMS buffer implementation in `unreal.S`.

```
ttyS0 at 0x3f8, irq 4 is a 16550A80x25 emulating ANSI (3 virtual consoles)
ttyS1 at 0x2f8, irq 3 is a 16550A
ttyS2 at 0x3e8, irq 7 is a 16550A
ttyS3 at 0x2e8, irq 10 is a 16550A
xms: 19456k available, using unreal mode, EXT bufheads
     128 xms buffers (base @ 0x110000), 8K L1-cache, 20 req hdrs
eth: ne0 at 0x340, irq 9 not found
eth: wd0 at 0x320, irq 15, ram 0xcc00, (wd8013) MAC 00:00:C0:BC:8F:4B, flags 0x0
eth: ee0 at 0x360, irq 11, (ee16) MAC 00:aa:00:55:e8:f5 (HWconf: IRQ 11, TP, 16-bit bus)
	 32k RAM (shmem 0xd000), flags 0x80
hd0: AT/IDE controller at 0x1f0
hda: IDE CHS: 993/16/63, Multisector I/O, max 4 sects
df: Floppy controller, FDC 8272A @ irq 6, DMA 2
df0: 1.44M (type 4), df1: 1.2M (type 2) [CMOS]
Floppy cache 5k, available 5k
Partitions: hda:(0,1000944)  hda1:(63,131985)  hda2:(132048,133056)  hda3:(265104,133056)  hda4:(398160,601776) 
PC/AT class, cpu 7, syscaps 0xff, 640K base ram, 16 tasks, 64 files, 96 inodes
TLVC 0.6.0 (64208 text, 17344 ftext, 11072 data, 7712 bss, 46750 heap)
Kernel text 0xffff, ftext 0x1d0, init 0x385, data 0x60c, top 0xa000, 551K free
MINIX: inodes 21845 imap 3 zmap 8, first data 696
VFS: Mounted root 0x0503 (minix filesystem).
Running /etc/rc.sys script
fsck -a /dev/hda3: /dev/rhda3 is clean, no check.
Starting networking on ee0
ktcp -b -p ee0 10.0.2.17 10.0.2.1 255.255.255.0
ktcp: ip 10.0.2.17, gateway 10.0.2.1, netmask 255.255.255.0
ktcp: /dev/ee0 mac 00.aa.00.55.e8.f5 mtu 1500
Mon Mar 24 19:53:17 2025


[tlvc17] TLVC 0.6.0

login:
```

### DIFF
--- a/tlvc/arch/i86/boot/crt0.S
+++ b/tlvc/arch/i86/boot/crt0.S
@@ -10,10 +10,9 @@
         .extern start_kernel
         .global early_putchar
 
-#define HMA_KERNEL
-#ifdef HMA_KERNEL
-	jmp	_start
-	.word	0,0,0,0,0,0,0
+#ifdef CONFIG_HMA_KERNEL
+	jmp	_start		// 16byte filler to move entry point to the
+	.word	0,0,0,0,0,0,0	// ... start of the HMA
 #endif
 _start:
 
--- a/tlvc/arch/i86/boot/crt0.S
+++ b/tlvc/arch/i86/boot/crt0.S
@@ -10,10 +10,8 @@
         .extern start_kernel
         .global early_putchar
 
-#ifdef CONFIG_HMA_KERNEL
 	jmp	_start		// 16byte filler to move entry point to the
 	.word	0,0,0,0,0,0,0	// ... start of the HMA
-#endif
 _start:
 
 /*
--- a/tlvc/arch/i86/boot/setup.S
+++ b/tlvc/arch/i86/boot/setup.S
@@ -62,6 +62,7 @@
 !	...
 !	0x1e2:	part_offset	long Sector offset of booted MBR partition
 !	0x1e6:	elks_magic	long "ELKS", written by build tool
+|	0x1ea:	xms_kbytes	word Size of XMS memory in Kbytes
 !	0x1ea-0x1ee:		5 bytes UNUSED
 !	0x1ef:	SETUPSEG	word UNUSED
 !	0x1f1:	setup_sects	byte in 512-byte sectors, written by build tool
@@ -512,10 +513,15 @@ sys_hdr_good:
 	mov %ax,%ds
 	xor %si,%si
 
-#ifdef CONFIG_HMA_KERNEL
+	push %ds
+	mov $SETUP_DATA,%ax
+	mov %ax,%ds
 	mov $0x8800,%ax
 	int $0x15	// get xms size
 	jc  no_xms
+	mov %ax,(xms_kbytes)
+	pop %ds
+#ifdef CONFIG_HMA_KERNEL
 	cmp $64,%ax
 	jl  no_xms
 	mov $0x2401,%ax
@@ -527,8 +533,10 @@ sys_hdr_good:
 2:
 	mov $HMA_SEG,%ax
 	jmp 1f
-no_xms:
 #endif
+no_xms:
+	movw $0,(xms_kbytes)
+	pop %ds
 	mov $REL_SYSSEG,%ax
 1:
 	mov %ax,%es
@@ -727,6 +735,8 @@ data_reloc:
 	.hex4sp	%cx,"entry:"
 	.if debug_output
 	.ifeq serial_output // if not serial output
+	mov $debug_prompt, %ax
+	call csputs
 	xor %ah,%ah         // read key press
 	int $0x16
 	.endif
@@ -874,6 +884,10 @@ hex4sp: call hex4
 
 hello_mess:
 	.ascii "\r\nTLVC Setup \0"
+.if debug_output
+debug_prompt:
+	.ascii "\r\nPress key to boot: \0"
+.endif
 
 //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 // Architecture specific routines for IBM PC
@@ -892,7 +906,7 @@ arch_get_mem:
 #endif
 	ret
 
-// A20 handlers - from unrteal.c
+// A20 handlers - from unreal.c
 // Even quite modern BIOSes don't support the A20 INT15 calls (QEMU does)
 // set_a20: Attempt to set the A20 line on 286 and higher
 // verify_a20: Check that a20 is set/working
--- a/tlvc/arch/i86/boot/setup.S
+++ b/tlvc/arch/i86/boot/setup.S
@@ -62,8 +62,9 @@
 !	...
 !	0x1e2:	part_offset	long Sector offset of booted MBR partition
 !	0x1e6:	elks_magic	long "ELKS", written by build tool
-|	0x1ea:	xms_kbytes	word Size of XMS memory in Kbytes
-!	0x1ea-0x1ee:		5 bytes UNUSED
+!	0x1ea:	xms_kbytes	word size of xms memory in kbytes
+!	0x1ec:	hma_kernel	byte hma=kernel seen in /bootopts
+!	0x1ed-0x1ee:		2 bytes UNUSED
 !	0x1ef:	SETUPSEG	word UNUSED
 !	0x1f1:	setup_sects	byte in 512-byte sectors, written by build tool
 !	0x1f2:	ROOTFLAGS	word UNUSED
@@ -81,8 +82,6 @@
 #include <linuxmt/boot.h>
 
 // Relocating loader debug option
-// NOTE: If running the kernel in HMA, setup hangs when jumping to the kernel if
-// debug_output is active and serial_output is NOT active.
 debug_output    =       0       // display various register values during execution
 serial_output   =       0       // use INT 14 serial instead of INT 10 console out
 debug_loader    =       0       // display relocations
@@ -481,11 +480,11 @@ sys_hdr_msg:
 
 sys_hdr_good:
 	mov %sp,%bp
-	mov 8,%dx     // -2(%bp) code (text) size
+	mov 8,%dx     // -2(%bp) .text code size
 	push %dx
 	add $15,%dx   // align on next paragraph (issue #209)
 	mov $4,%cl
-	shr %cl,%dx   // DX = code size in paragraph
+	shr %cl,%dx   // DX = code size in paragraphs
 	shr %cl,%bx   // BX = header size in paragraphs
 
 	mov 12,%ax    // -4(%bp) data size
@@ -512,49 +511,31 @@ sys_hdr_good:
 	add %bx,%ax      // skip header
 	mov %ax,%ds
 	xor %si,%si
+	xor %di,%di
 
-	push %ds
-	mov $SETUP_DATA,%ax
-	mov %ax,%ds
-	mov $0x8800,%ax
-	int $0x15	// get xms size
-	jc  no_xms
-	mov %ax,(xms_kbytes)
-	pop %ds
-#ifdef CONFIG_HMA_KERNEL
-	cmp $64,%ax
-	jl  no_xms
-	mov $0x2401,%ax
-	int $0x15	// Set a20
-	jnc 2f
-	call set_a20
-	call verify_a20
-	jc no_xms
-2:
+#ifdef CONFIG_ARCH_IBMPC
+// Check whether relocating .text to HMA is possible
+	call checkhma
+	jc 2f		 // no HMA
 	mov $HMA_SEG,%ax
 	jmp 1f
 #endif
-no_xms:
-	movw $0,(xms_kbytes)
-	pop %ds
-	mov $REL_SYSSEG,%ax
-1:
-	mov %ax,%es
+2:	mov $REL_SYSSEG,%ax
+1:	mov %ax,%es
 	mov %ax,-18(%bp) // save .text segment
 	.ifeq debug_output
 	call hex4sp
 	.endif
 
-	add %dx,%ax	// add code paragraphs = .fartext start
-#ifdef CONFIG_HMA_KERNEL
-	jnc 1f		// skip if it didn't overflow (no HMA) 
-	movw $REL_SYSSEG,%ax	// .fartext start here
-1:
-#endif
-	mov %ax,-20(%bp) // save .data start in case no .fartext
-	xor %di,%di
-
 	mov -2(%bp),%cx  // code size in bytes
+	add %dx,%ax      // add code paragraphs = .fartext start
+	jnc 1f		 // skip if didn't overflow (no HMA)
+	mov $REL_SYSSEG,%ax // .fartext starts at normal .text location
+	add $16,%si	 // skip first paragraph at FFFF:0000
+	add $16,%di
+	sub $16,%cx
+1:	mov %ax,-20(%bp) // save .data start in case no .fartext
+
 	.hex4sp	%ds,"\nCopy DEF_SYSSEG .text to  REL_SYSSEG DS:"
 	.hex4sp	%si,"SI:"
 	.hex4sp	%es,"ES:"
@@ -594,12 +575,9 @@ no_xms:
 
 	mov -18(%bp),%ax  // kernel .text segment
 	add %dx,%ax       // skip code paragraphs
-#ifdef CONFIG_HMA_KERNEL
 	jnc 1f
-	mov $REL_SYSSEG,%ax	// if text in HMA, start fartext here
-#endif
-1:
-	mov %ax,%es
+	mov $REL_SYSSEG,%ax // if text in HMA, start fartext here
+1:	mov %ax,%es
 	mov %ax,-20(%bp)  // save .fartext segment
 	xor %di,%di
 
@@ -735,7 +713,7 @@ data_reloc:
 	.hex4sp	%cx,"entry:"
 	.if debug_output
 	.ifeq serial_output // if not serial output
-	mov $debug_prompt, %ax
+	mov $debug_prompt,%ax
 	call csputs
 	xor %ah,%ah         // read key press
 	int $0x16
@@ -839,6 +817,9 @@ csputs:	push	%bx
 	jmp	1b
 2:	pop	%bx
 	ret
+
+debug_prompt:
+	.ascii	"\r\nPress key to boot: \0"
 .endif
 
 // Output hex nibble, byte and word. All registers saved.
@@ -883,11 +864,7 @@ hex4sp: call hex4
 	ret
 
 hello_mess:
-	.ascii "\r\nTLVC Setup \0"
-.if debug_output
-debug_prompt:
-	.ascii "\r\nPress key to boot: \0"
-.endif
+	.ascii "\r\nELKS Setup \0"
 
 //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 // Architecture specific routines for IBM PC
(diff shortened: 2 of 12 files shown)


ghaerr:
Nice work. This turned out to use less code than I would have thought, but given that setup.S has been heavily cleaned up over the last four years, and your nice trick of using "jnc jumps" for determining whether the the .text segment is 0xFFFF or not, this big enhancement looks very straightforward :)

I had an HMA patch put together a while back that moved the kernel data segment to HMA, but thanks to your recent finding that INT 15h AH=87h block move resets A20, that wouldn't have worked anyways! Given that the kernel data segment can be shrunk down quite a bit when memory is tight, but the .text kernel size is now almost alway 58k+ bytes, this is a better solution, and a great way to add another contiguous 64k for programs!

Speaking of which A20 resets, I imagined a workaround for the problem of A20 being switched off after XMS INT 15h block moves, on systems where unreal mode won't work: if the function that calls the INT 15h `block_move` were a FARPROC function (e.g. void FARPROC int15_fmemcpyw), then when the INT 15 BIOS interrupt returned, it wouldn't be returning to  the .text HMA CS segment, but instead to a .fartext segment somewhere in resident memory. The `enable_a20_gate` function could then be called before returning to the HMA CS segment calling code.

This would seem to work, unless an interrupt was pending, in which the interrupt execution would take precedence and `_irqit` would end up being called before A20 was switched back on, crashing the system. Interrupts could be turned off prior to calling INT 15 and only re-enabled after enable_a20_gate - but this might not work if the BIOS itself reenabled interrupts after reseting A20 but before IRET.

So it seems this could be worked around, but the "solution" would BIOS-dependent. More consideration leads me to believe that the BIOS wouldn't be disabling A20 if it didn't want to keep "users" out of XMS memory in the first place, so I suppose its best to just leave HMA kernel and XMS INT 15 as mutually separate. The Compaq 386 BIOS (the only BIOS I've actually tested and require INT 15h on) actually runs in protected mode, so I can imagine it keeps A20 off to protect itself.

It could be worth playing with though, if you are already running this and testing on your Compaq 386.

> // NOTE: If running the kernel in HMA, setup hangs when jumping to the kernel if
 // debug_output is active and serial_output is NOT active.

Do you know the reason for this? Setup.S uses INT 14h for serial out and INT 10h for console output, so it would seem that getting debug output on console should be able to be made to work with an HMA kernel being booted.

A final point (and possibly related to the above kernel startup problem): it appears that the kernel .text segment is copied to address starting at FFFF:0000 (real address FFFF0) which is actually BIOS ROM for 16 bytes, right? I saw your previous commit where the .text start is prefaced by a 16-byte paragraph with a 'jmp _start' (is that a 3 or 2 byte jump?) and 14 zeros. However, these 16 bytes aren't actually copied to XMS, they are copied to a ROM address, right? And your current code doesn't adjust the kernel a.out header "entry" point, which is `_start`, which is almost always 0, since crt0.S is the first object file given to the linker. So this code wouldn't work if the kernel entry point _start is != 0.

The HMA setup.S code doesn't actually update the kernel a.out entry point, which is assumed to be 0, and thus still jumps to CS:entry, which is CS:0, or actually FFFF:0 (still ROM!?). I would think you would want to adjust the entry point by 16 bytes, thus actually jumping to CS:0010, or FFFF:00010, or real address 100000. This should then start the kernel properly, for any entry point \_start value. I may be missing something but it appears that the current code may actually be running the top 16 bytes of ROM and the find itself at the kernel .text segment.

Mellvik:
> Nice work. This turned out to use less code than I would have thought, but given that setup.S has been heavily cleaned up over the last four years, and your nice trick of using "jnc jumps" for determining whether the the .text segment is 0xFFFF or not, this big enhancement looks very straightforward :)

Thanks @ghaerr, it has been an interesting and very educational journey (again), which I (correctly) figured would send me to places I've never been in the system. As is usually the case, it all started with a hardware problem - ironically on the Compaq Portable Plus, an XT clone, running too hot when fully populated with (old style) memory (chips). A long story for another day and time.

Anyway, yes, for a while there - before I moved from QEMU to real hardware, it looked like the implementation would be limited to just a few instructions in `setup.S`. Then came reality (that is A20, which is always on, regardless, on QEMU), and the code from `unreal.S` had to be moved across.

> Speaking of which A20 resets, I imagined a workaround for the problem of A20 being switched off after XMS INT 15h block moves, on systems where unreal mode won't work: if the function that calls the INT 15h `block_move` were a FARPROC function (e.g. void FARPROC int15_fmemcpyw), then when the INT 15 BIOS interrupt returned, it wouldn't be returning to the .text HMA CS segment, but instead to a .fartext segment somewhere in resident memory. The `enable_a20_gate` function could then be called before returning to the HMA CS segment calling code.
> 
> This would seem to work, unless an interrupt was pending, in which the interrupt execution would take precedence and `_irqit` would end up being called before A20 was switched back on, crashing the system. Interrupts could be turned off prior to calling INT 15 and only re-enabled after enable_a20_gate - but this might not work if the BIOS itself reenabled interrupts after reseting A20 but before IRET.
> 
> So it seems this could be worked around, but the "solution" would BIOS-dependent. More consideration leads me to believe that the BIOS wouldn't be disabling A20 if it didn't want to keep "users" out of XMS memory in the first place, so I suppose its best to just leave HMA kernel and XMS INT 15 as mutually separate. The Compaq 386 BIOS (the only BIOS I've actually tested and require INT 15h on) actually runs in protected mode, so I can imagine it keeps A20 off to protect itself.
> 
> It could be worth playing with though, if you are already running this and testing on your Compaq 386.

That's an interesting chain of thought. I agree, technically that should work and the question is how fast/slow would it really be. It's a lot of extra code in addition to the BIOS call itself, which potentially (although not in our case) can transfer quite large amounts of data.

There is no question that in the COMPAQ case, they want us to stay out of XMS except via the BIOS and the included DOS utilities/drivers - likely because of the remapping etc. they are doing with the UMB memory (not optional, always there) to just below 16M. Plus there is the LIM EMS implementation (the HW part which seems to be available, it's on my list to check that), which possibly can mess up things if used.

Interestingly, my AMI 386SX system, which is 15 years newer, has the same feature - not the mandatory remapping, but the BIOS XMS block transfers turning off A20, and A20 control via the BIOS not available. OTOH, this system doesn't prevent unreal mode.

Confusing but the gist of it seems to be this:
- BIOS XMS enforces tight A20 control
- Few if any BIOSes implement A20 control, likely for that reason (I'm tempted to #ifdef that out from both setup.S and unreal.S)
- Those doing memory mapping trickery with extended memory, block unreal mode - likely only the oldest of systems because memory was still really expensive at that time, making it worthwhile to optimize the usage of even a few hundred k.

> 
> > // NOTE: If running the kernel in HMA, setup hangs when jumping to the kernel if
> > // debug_output is active and serial_output is NOT active.
> 
> Do you know the reason for this? Setup.S uses INT 14h for serial out and INT 10h for console output, so it would seem that getting debug output on console should be able to be made to work with an HMA kernel being booted.

No, this is a mystery. My guess is that it has to do with the fact that the console output is dealing with video memory and that the INT10h handler - for good measure - ensures A20 is off. I decided not to pursue it for now.

> 
> A final point (and possibly related to the above kernel startup problem): 

I missed you there, what was the startup problem?

> ...it appears that the kernel .text segment is copied to address starting at FFFF:0000 (real address FFFF0) which is actually BIOS ROM for 16 bytes, right? I saw your previous commit where the .text start is prefaced by a 16-byte paragraph with a 'jmp _start' (is that a 3 or 2 byte jump?) and 14 zeros. However, these 16 bytes aren't actually copied to XMS, they are copied to a ROM address, right? And your current code doesn't adjust the kernel a.out header "entry" point, which is `_start`, which is almost always 0, since crt0.S is the first object file given to the linker. So this code wouldn't work if the kernel entry point _start is != 0.

This is not the way I see it, but I'm also not sure about this. First, the initial `jmp` and the padding changes  the address of `_start` to base+0x10. Second, yes the initial paragraph gets uselessly written to the ROM area,  _start is at 0xffff:0010. Since the address of _start is 0x10, that's where the setup `lret` sends us, otherwise it wouldn't work. I believe the following debug output confirms that:
```
Done SS:8000 DS/ES:13E8 .text size:FAD0 .fartext size:43C0 .data size:2B40 
Jump CS:FFFF entry:0010 Direct console, scan kbd 80x25 emulating ANSI (3 virtual consoles)
xms: 3072k available, using unreal mode, EXT bufheads
     128 xms buffers (base @ 0x110000), 8K L1-cache, 20 req hdrs
...
```

ghaerr:
> The initial jmp and the padding changes the address of _start to base+0x10.

Oops - I missed that, since that code isn't actually part of this commit, I remembered it incorrectly.

> // NOTE: If running the kernel in HMA, setup hangs when jumping to the kernel if
// debug_output is active and serial_output is NOT active.

Regarding this startup problem - when debug output and console output is active, does the system hang on both QEMU and real hardware, or just real hardware? And are we talking Compaq real hardware, or other real hardware?

> likely because of the remapping etc. they are doing with the UMB memory (not optional, always there) to just below 16M.

My understanding is that the BIOS is actually mapped high as you describe, and then shadow-mapped to segment F000. Or is it the reverse? I'm wondering whether the console video at A000/B800 is also remapped, somehow causing the hang and/or lack of console display when debug active.

Mellvik:
> > Those doing memory mapping trickery with extended memory, block unreal mode
> 
> I don't think the issue is a matter of "blocking" unreal mode - as there is no way to do so - as much as when the BIOS uses protected mode (or even unreal mode itself), it uses a different GDT which invalidates ours, so it doesn't work.

I agree with that and I didn't mean it literally. Whatever they do, as we've experienced, just causes the system to hang. That will keep intruders like us out.

Mellvik:
For the record, I just tested the (COMPAQ) 80286 w/6MB ram and IN15 xms support. Works as expected and reminded me that CONFIG_FS_XMS_BUFFER doesn't have to be set in order to enable a HMA kernel. If CONFIG_HMA_KERNEL is there and XMS memory is found (BIOS call for now), HMA kernel is used.

Now `meminfo` needs to be redone again (it never had either XMS nor extbufs), and more info is needed from the kernel I order to get things right. Now, instead of adding even more IOCTLs to the mem driver, it occurs to me that defining a `meminfo` structure that has everything is better. One call to /dev/mem, simpler etc. What's your take, @ghaerr?

The third thing that just popped up when I read about XMS ramdisk support in ELKS (congrats, this is good); the CONFIG_FS_XMS_BUFFER define is becoming a misnomer. In a while we may have LIM EMS support and more, and even the _FS_ part becomes misleading. CONFIG_XMS_SUPPORT or just CONFIG_XMS may be more appropriate.

With a 'soft' (as in careful) background sync mechanism that ramdisk could become root. That would be one fast (16 bit - sort of) system.

ghaerr:
> Since it take about 10 seconds, here's the output from QEMU with setup debug like this:

Thanks for the screenshot. I plan on looking further into this and debugging it. I have a feeling if not, it'll come back to bite.

> the CONFIG_FS_XMS_BUFFER define is becoming a misnomer.

> CONFIG_XMS_SUPPORT or just CONFIG_XMS may be more appropriate.

For ELKS, I don't want to break all my users' config files every time I add a new feature. So from my point of view, keeping the same CONFIG_xxx names is a good thing, at least until I get user feedback. 

When writing the XMS ramdisk, I also considered the same points you've brought up, and ended up with the following set of CONFIG_FS_xxx names, at least for what we're now discussing:
```
CONFIG_FS_EXTERNAL_BUFFER=y
CONFIG_FS_NR_EXT_BUFFERS=64
CONFIG_FS_XMS=y
CONFIG_FS_XMS_BUFFER=y
CONFIG_FS_XMS_RAMDISK=y
CONFIG_FS_NR_XMS_BUFFERS=2500
# CONFIG_FS_XMS_INT15 is not set
```
The fs/config.in ensures that if CONFIG_FS_XMS=y isn't set, none of the other XMS options will be set, and defconfig automatically updates older config files.

I didn't choose the coding route you took with HMA buffers (and using xms_size to indicate if XMS is active) because:
- The kernel may want to know the size of XMS before enabling it for a given feature.
- In my design, xms_init doesn't have to be called at kernel startup; thus XMS buffers may not be set but XMS ramdisk is wanted.
- The XMS ramdisk doesn't call xms_init at init time either. This is because the size of both main memory and XMS memory ram disks are set using the `ramdisk` utility. It may be that the kernel is compiled for XMS but the ramdisk is never initialized, and the system is more stable across wider PC environments if the XMS initialization is left until the `ramdisk` ioctl is executed, and/or when XMS is needed.
- IMO xms_size is best filled in only at setup.S time, and then passed to the kernel via a SETUP_xxx macro in config.h, rather than writing code to query it again in the kernel. This also has the advantage that a system can be configured to use less XMS than is present, for testing or otherwise, just like the other SETUP_ config.h variables customized per PC environment. Normally it wouldn't matter that much, but with the HMA enhancement setup.S must check for XMS anyways.

That said, yes, with the XMS ramdrive there is now an inconsistency with configuring the XMS ramdrive (/dev/ssd), done in the menuconfig filesystem section (with the rest of the XMS setup) versus the main memory ramdisk (/dev/rd0), which is configured as part of the menuconfig block devices. This will want to get resolved, but I'm thinking of doing that after combining the rd.c ramdisk driver with the ssd.c XMS driver. I used ssd.c as a base since it's much newer and cleaner, so much so it only took about an hour to write the XMS subdriver. (The SSD block driver now has three subdrivers - "test" for testing async I/O callbacks with variable-length delays, CF card support on 8018x and now XMS support. The subdriver concept would work well for NIC and other async CB drivers but that's another story) . The older rd.c driver is a bit of a mess. I'm thinking in the final version there ought to only be a single ramdisk driver and the user can specify where they want the ram to come from (main, xms, umb, whatever).

> With a 'soft' (as in careful) background sync mechanism that ramdisk could become root.

I'm not seeing the benefit there - why not just use XMS buffers and /bootopts sync= ?

ghaerr:
> // NOTE: If running the kernel in HMA, setup hangs when jumping to the kernel if
// debug_output is active and serial_output is NOT active.

Sorry I didn't understand this until just now - this isn't a bug: setup is waiting for a keystroke to continue:
```
        .hex4sp %ax,"\nJump CS:"
        .hex4sp %cx,"entry:"
        .if debug_output
        .ifeq serial_output // if not serial output
        xor %ah,%ah         // read key press
        int $0x16
        .endif
        .endif

        lret               // jump to REL_SYSSEG:_start to start kernel
```
The INT 16h read keypress was added so that users without a serial port could still see the debug output from the kernel booting - otherwise the entire output is scrolled off screen by the regular boot messages, which would be useless. 

Obviously it'd be better to prompt with "[press key to boot]", or something like that. With setup's size at 1374 (no debug) and 2871 bytes (debug without HMA enhancement), we can afford it. I'll add that to my next PR.

Mellvik:
> Sorry I didn't understand this until just now - this isn't a bug: setup is waiting for a keystroke to continue.

No sh.., I should have seen that immediately - I clearly wasn't looking. Thanks!

Mellvik:
>> With a 'soft' (as in careful) background sync mechanism that ramdisk could become root.

> I'm not seeing the benefit there - why not just use XMS buffers and /bootopts sync= ?

This is one of many examples where ELKS and TLVC differ in orientation. One of TLVC's ambitions is to be a tool to diagnose and/or investigate systems, usually vintage systems. Running off of a floppy is in many cases the normal in such scenarios, the disk is usually already 'occupied' with FAT partitions, and the user doesn't want to risk anything screwed up on the disk while investigating. IOW - not having to install anything and still having a decent system is very beneficial. So here's the idea I've been playing with:
- boot from floppy
- have setup use BIOS XMS block move to copy the entire floppy to XMS. That will take a few seconds and would require a console message indicating what's going on.
- make the xms image the root filesystem when booting
- block attempts to mount the floppy image
- have a sync mechanism that does lazy syncing back to the floppy, like a few blocks every second, which would be unnoticable when using the direct driver. If the blocks are small, even with BIOS IO. The latter being important for more recent systems where USB floppy is the only option. 

That would make a very fast and convenient 'just let me check out this system' tool.

Still just a thought, although the lazy sync is something we've discussed (without giving it a name) a number of times before, after XMS buffers - and with them some previously unthinkable challenges with syncing and delays - became available.

Mellvik:
Thanks, @ghaerr - another useful discussion.

> I didn't choose the coding route you took with HMA buffers (and using xms_size to indicate if XMS is active) because:
> 
> * The kernel may want to know the size of XMS before enabling it for a given feature.

Agree, that's what the new `setup.S` does.  HMA buffers came first, thus the `get_xms_size` code in `unreal.S`. It's going away.

> * In my design, xms_init doesn't have to be called at kernel startup; thus XMS buffers may not be set but XMS ramdisk is wanted.

Again as implemented in TLVC. The kernel will be loaded high if possible, regardless of the presence or absence of XMS code and buffers.

> * The XMS ramdisk doesn't call xms_init at init time either. This is because the size of both main memory and XMS memory ram disks are set using the `ramdisk` utility. It may be that the kernel is compiled for XMS but the ramdisk is never initialized, and the system is more stable across wider PC environments if the XMS initialization is left until the `ramdisk` ioctl is executed, and/or when XMS is needed.

Makes sense. I plan to take a look at the xms ramdisk implementation. Ramdisk was never an issue in TLVC, now it has an entirely different potential.

> * IMO xms_size is best filled in only at setup.S time, and then passed to the kernel via a SETUP_xxx macro in config.h, rather than writing code to query it again in the kernel. This also has the advantage that a system can be configured to use less XMS than is present, for testing or otherwise, just like the other SETUP_ config.h variables customized per PC environment. Normally it wouldn't matter that much, but with the HMA enhancement setup.S must check for XMS anyways.

Completely agree on that one. That's the way I'm heading too. The size query in xms_init is a leftover.

> 
> ...  The subdriver concept would work well for NIC and other async CB drivers but that's another story. 

Exactly! This is what I was referring to when we discussed network drivers and more the other day. Having a 'meta' network driver will simplify my life significantly. A continuation of our discussion about  NIC driver APIs etc from last yer.

> The older rd.c driver is a bit of a mess. I'm thinking in the final version there ought to only be a single ramdisk driver and the user can specify where they want the ram to come from (main, xms, umb, whatever).

Sounds good. Admittedly I never even looked at that one. There are still many dark corners in the system.

Mellvik:
Completeing the discussion on HMA usage, @ghaerr:

I did a run-through of the alternatives before starting with the external bufheads, and found - as stated in the PR notes - that there were really only two candidates: Bufheads and kernel text. 

Kernel data seemed dangerous: We would have to change (revert) the buffer code which is now addressing buffers by segment only, merging segment with offset. There may be other occurrences of segment register manipulations that I have either forgotten or never known, so it appears like a minefield. Using HMA as main memory is probably off the table for the same reason.

Kernel fartext is too small and HMA'ing it would prevent the release of INITSEG after boot.

Mellvik:
> When writing the XMS ramdisk, I also considered the same points you've brought up, and ended up with the following set of CONFIG_FS_xxx names, at least for what we're now discussing:
> ```
> CONFIG_FS_EXTERNAL_BUFFER=y
> CONFIG_FS_NR_EXT_BUFFERS=64
> CONFIG_FS_XMS=y
> CONFIG_FS_XMS_BUFFER=y
> CONFIG_FS_XMS_RAMDISK=y
> CONFIG_FS_NR_XMS_BUFFERS=2500
> # CONFIG_FS_XMS_INT15 is not set
> ```
> The fs/config.in ensures that if CONFIG_FS_XMS=y isn't set, none of the other XMS options will be set, and defconfig automatically updates older config files.

This is good, @ghaerr - I'm going to adopt this setup!

ghaerr:
> Kernel data seemed dangerous: We would have to change (revert) the buffer code which is now addressing buffers by segment only, merging segment with offset.

You're right - I'd forgotten about that optimization - that would not have worked! 

Given your A20 reset finding when using BIOS INT 15 also pretty much kills any good idea of using kernel data in HMA, especially since kernel text in HMA can still work on BIOS INT 15 machines with XMS buffers turned off - a large savings of main memory.

Since the kernel stack is also in the kernel data segment, even if the kernel is running in .fartext a single interrupt would crash the system with A20 off.

> Using HMA as main memory is probably off the table for the same reason.

I had thought of the idea of using add_seg to add the "memory region" from FFFF to 0000 to the main memory pool, allowing it to be allocated piecemeal, but suffers from the huge problem that the allocator always returns a segment with a zero offset.

So I agree, HMA kernel text is a great solution which maximizes memory gained with very little kernel code overhead!

I plan on bringing this enhancement over to ELKS and setting it ON by default; the memory savings are too great not too. That will also allow us to ensure that there are no other issues to system stability with it.

Mellvik:
> I plan on bringing this enhancement over to ELKS and setting it ON by default; the memory savings are too great not too. That will also allow us to ensure that there are no other issues to system stability with it.

Glad to hear it. The more beating it can get etc. I need to go and (re)fix meminfo :-). Then there is this other memory related 'advancement'. More later.

Thanks

Mellvik:
BTW @ghaerr, I'm sure the code in `setup.S` saving `xms_kbytes` to SETUP_DATA could be more compact. It seems clumsy to me.

ghaerr:
>  instead of adding even more IOCTLs to the mem driver, it occurs to me that defining a meminfo structure that has everything is better.

Somewhat tough call on this one. It likely depends on whether the variables/addresses that would be packed into a struct meminfo are also needed by other programs, e.g. ps. The simple benefit is that it's cleaner, but the drawbacks are the above, plus also the new application dependency on sizeof(struct meminfo) not changing, requiring recompilation. Another issue is that the packing code and added code for verified_memcpy_tofs passing it bloats the kernel, whereas now adding a new variable/address is very tiny.

> I'm sure the code in setup.S saving xms_kbytes to SETUP_DATA could be more compact. 

I don't think the setup variables have to be zeroed since IIRC the setup data segment is cleared beforehand, but will have to check. The initialization could probably be done in a better place, where DS is already set to SETUP_DATA. In each of the many setup.S rewrites I've done over the years I've tried to move all the hacks that had gone in over the years to specialized areas, that idea might work well for this too. I'll take a closer look at it and respond.

Mellvik:
> > instead of adding even more IOCTLs to the mem driver, it occurs to me that defining a meminfo structure that has everything is better.
> 
> Somewhat tough call on this one. It likely depends on whether the variables/addresses that would be packed into a struct meminfo are also needed by other programs, e.g. ps. The simple benefit is that it's cleaner, but the drawbacks are the above, plus also the new application dependency on sizeof(struct meminfo) not changing, requiring recompilation. Another issue is that the packing code and added code for verified_memcpy_tofs passing it bloats the kernel, whereas now adding a new variable/address is very tiny.

Good points, Thank you.

Mellvik:
The 2nd commit adds @ghaerr's reworked `setup.S` which among other things enables kernel HMA loading to be enabled/disabled via `bootopts`. A very important addition, in particular for testing. 

Also a fix for the `reboot`command not working plus a number of cleanups. The `CONFIG_HMA_KERNEL` is no longer needed and has been deleted. Code shared between the xms code in `lib/unreal.S` and  `boot/setup.S` has been separated out into its own file in `i86/lib/a20.inc`.

PR #160, opened 2025-03-29
[kernel] Fixes and enhancements for HMA usage

Mellvik:
As we're getting more experience with HMA usage, problems and opportunities arise. This PR addresses some of them:

- Fix boot failure on XT class systems when HMA support is CONFIGured in (setup.S) -  by checking cpu type before accessing some functions
- Allow external bufheads to use the HMA memory not occupied by the kernel, if possible. The ideal number of buffers for a practical system is between 128 and 256, which in many cases allows the bufheads to sneak in on top of the kernel.
- Simplify the selection logic: This  new functionality allows for many combinations and scenarios and equally many boot message combinations, even without thinking XMS ramdisk. 

What's really missing still, is a simple way to check whether `unreal mode` works without hanging the system if it doesn't. With that, all the `CONFIG_` stuff could be eliminated and everything autoconfigured.

### DIFF
--- a/tlvc/arch/i86/boot/setup.S
+++ b/tlvc/arch/i86/boot/setup.S
@@ -84,6 +84,7 @@
 // Relocating loader debug option
 debug_output    =       0       // display various register values during execution
 serial_output   =       0       // use INT 14 serial instead of INT 10 console out
+				// NOTE: Not supported on some/most XT class systems
 debug_loader    =       0       // display relocations
 
 
@@ -514,13 +515,21 @@ sys_hdr_good:
 	xor %di,%di
 
 #ifdef CONFIG_ARCH_IBMPC
-// Check whether relocating .text to HMA is possible
+// If 80286 or higher, check whether relocating .text to HMA is possible
+	push %ds
+	mov $INITSEG,%ax
+	mov %ax,%ds
+	mov cpu_type,%ax
+	cmpb $6,cpu_type
+	jl 2f
 	call checkhma
 	jc 2f		 // no HMA
 	mov $HMA_SEG,%ax
+	pop %ds
 	jmp 1f
+2:	pop %ds
 #endif
-2:	mov $REL_SYSSEG,%ax
+	mov $REL_SYSSEG,%ax
 1:	mov %ax,%es
 	mov %ax,-18(%bp) // save .text segment
 	.ifeq debug_output
@@ -864,7 +873,7 @@ hex4sp: call hex4
 	ret
 
 hello_mess:
-	.ascii "\r\nELKS Setup \0"
+	.ascii "\r\nTLVC Setup \0"
 
 //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 // Architecture specific routines for IBM PC
@@ -943,7 +952,11 @@ novga:	mov	%al,14		// CGA 25 rows
 
 // check for hma=kernel in /bootopts, return AX = 0 if not
 hmabootopts:
-	push	%ds
+	cmpb	$6,cpu_type	// ignore if < 80826
+	jae	1f
+	xor	%al,%al
+	ret
+1:	push	%ds
 	push	%es
 
 	push	%cs			// ES = our code segment
@@ -976,10 +989,8 @@ hma_string:
 	.ascii	"\nhma=kernel"		// size = 11
 
 // check whether HMA can be enabled, return NC if so with A20 enabled
+// Assumes DS=INITSEG
 checkhma:
-	push	%ds
-	mov	$INITSEG,%ax
-	mov	%ax,%ds
 	mov	xms_kbytes,%ax		//. must have xms size >= 64k
 	cmp	$64,%ax
 	jl	1f
@@ -992,8 +1003,7 @@ checkhma:
 	clc				// success
 	jmp 	2f
 1:	stc				// fail
-2:	pop	%ds
-	ret
+2:	ret
 
 // Write AL to console, save all registers
 putc:	push %ax
--- a/tlvc/arch/i86/mm/xms.c
+++ b/tlvc/arch/i86/mm/xms.c
@@ -19,6 +19,8 @@
 #ifdef CONFIG_FS_XMS_BUFFER
 
 extern int xms_size;
+extern int xms_avail;
+extern int hma_avail;
 
 /* these used in CONFIG_FS_XMS_INT15 only */
 struct gdt_table;
@@ -53,23 +55,31 @@ void xms_init(void)
 	}
 #else
 	if (kernel_cs == 0xffffU) {
-		printk("Not available with INT15: Kernel in HMA");
+		printk("Not available with INT15 and kernel in HMA");
+		xms_avail = 0;
 		return;
 	} 
 #endif
-	if (!xms_size) {
+	if (!xms_avail) {
 		printk("not available\n");
 		return;
 	}
-	printk("%uk available, ", xms_size);
+	printk("%uk available, ", xms_avail);
+#ifdef NOT_REQUIRED_ANYMORE
 	debug("A20 was %s", verify_a20()? "on" : "off");
 	enable_a20_gate();
 	enabled = verify_a20();
 	debug(" now %s, ", enabled? "on" : "off");
 	if (!enabled) {
-		xms_size = 0;
+		xms_size = 0;		/* No xms, no HMA */
+		xms_avail = 0;
 		printk("disabled, A20 error\n");
 	} else {
+#else
+	if (!hma_avail)  {
+		printk("disabled, A20 error\n");
+	} else {
+#endif
 #ifdef CONFIG_FS_XMS_INT15
 		printk("using int 15/1F");
 #else
--- a/tlvc/fs/buffer.c
+++ b/tlvc/fs/buffer.c
@@ -83,6 +83,8 @@ static struct wait_queue L1wait;                  /* Wait for a free L1 buffer a
 static int lastL1map;
 #endif
 extern int xms_size;		/* kbytes, 0 if not present */
+extern int xms_avail;		/* kbytes, usually xms_size - HMA */
+extern int hma_avail;
 static int map_count, remap_count, unmap_count;
 
 static int nr_free_bh, nr_bh;
@@ -137,7 +139,7 @@ static void INITPROC add_buffers(int nbufs, char *buf, ramdesc_t seg)
 	size_t offset;
 
         /* segment adjusted to require no offset to buffer */
-        offset = xms_size? ((n & 63) << BLOCK_SIZE_BITS) :
+        offset = xms_avail? ((n & 63) << BLOCK_SIZE_BITS) :
                               ((n & 63) << (BLOCK_SIZE_BITS - 4));
         ebh->b_L2seg = seg + offset;
 #else
@@ -196,8 +198,8 @@ int INITPROC buffer_init(void)
 #ifdef CONFIG_FS_XMS_BUFFER
     if (nr_xms_bufs)
         xms_init();       /* try to enable unreal mode and A20 gate*/
-    if (xms_size)
-        bufs_to_alloc = (nr_xms_bufs > (xms_size-64)) ? (xms_size-64) : nr_xms_bufs;
+    if (xms_avail)
+        bufs_to_alloc = (nr_xms_bufs > xms_avail) ? xms_avail : nr_xms_bufs;
 #endif
 #ifdef CONFIG_FAR_BUFHEADS
     if (bufs_to_alloc > 2975) bufs_to_alloc = 2975; /* max 64K far bufheads @22 bytes*/
@@ -222,17 +224,24 @@ int INITPROC buffer_init(void)
     if (!buffer_heads) return 1;
 #ifdef CONFIG_FAR_BUFHEADS
     size_t size = bufs_to_alloc * sizeof(ext_buffer_head);
-    seg_t hma_seg = 0xFFFFU; 
-    if (xms_size && kernel_cs != hma_seg) {		/* HMA available for ext headers */
+    seg_t hma_seg = 0xffffU; 
+    //printk("HMA space available: %u, need %u\n", 0xfff0 - (unsigned)_endtext, size); 
+    if (hma_avail && kernel_cs != hma_seg) {		/* HMA available for ext headers */
 	fmemsetw((void *)0x10, hma_seg, 0, size >> 1);
 	ext_buffer_heads = _MK_FP(hma_seg, 0x10);
 	printk(", HMA bufheads\n     ");
+    } else if (kernel_cs == hma_seg && (size < (hma_seg - (unsigned)_endtext))) {
+	/* kernel is loaded high, but there is still enough space for the buffer headers */
+	/* compact this later */
+	fmemsetw((void *)_endtext, hma_seg, 0, size >> 1);
+	ext_buffer_heads = _MK_FP(hma_seg, (unsigned)_endtext);
+	printk(", bufheads in high HMA\n     ");
     } else {
 	segment_s *seg = seg_alloc((size + 15) >> 4, SEG_FLAG_BUFHEAD);
 	if (!seg) return 1;
 	fmemsetw(0, seg->base, 0, size >> 1);
 	ext_buffer_heads = _MK_FP(seg->base, 0);
-	printk(", EXT bufheads\n     ");
+	if (xms_size) printk(", EXT bufheads\n     ");
     }
 #endif
     bh_next = bh_lru = bh_llru = buffer_heads;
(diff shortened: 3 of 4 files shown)


ghaerr:
> Fix boot failure on XT class systems when HMA support is CONFIGured in (setup.S) - by checking cpu type before accessing some functions

Where or what exactly is the boot failure on XT? I'm not seeing any issues with booting HMA kernel on XT systems. The requirements for HMA kernel are just 1) getting >= 64k for INT 15h AX=8800h get extended memory size, 2) A20 operational, and 3) hma=kernel in /bootopts, right?

> What's really missing still, is a simple way to check whether unreal mode works without hanging the system if it doesn't. 

Agreed that unreal mode can't be currently be "tested for" without knowing that the CPU is 386 or greater. Can't we just fix this with testing for "arch_cpu != 7 || check_unreal_mode <= 0" in xms_init?

HMA doesn't require 286 or 386 support, but XMS buffers and XMS ramdisk require 386+, correct? So that should allow XT systems to boot HMA (unless all XT systems don't have XMS, in which case xms_size == 0, which stop loading). That leaves 386+ systems to check for unreal mode for non-HMA XMS services, and 286 systems will likely have to enable INT 15 block move support.

I was thinking we could compile in CONFIG_FS_XMS_INT15 support and have xms_fmemcpy switch between linear and bios block moves, which should then allow HMA and XMS to be totally automated with the above arch_cpu check fix, right?

tlvc/arch/i86/boot/setup.S
```diff
@@ -514,13 +515,21 @@ sys_hdr_good:
...
 
 #ifdef CONFIG_ARCH_IBMPC
-// Check whether relocating .text to HMA is possible
+// If 80286 or higher, check whether relocating .text to HMA is possible
+	push %ds
+	mov $INITSEG,%ax
+	mov %ax,%ds
+	mov cpu_type,%ax
+	cmpb $6,cpu_type
+	jl 2f
```
ghaerr, line 524: This code should be moved to checkhma, shouldn't it? The sole purpose of checkhma is to determine whether HMA boot is a viable option or not. This would also eliminate the kluge DS push since it (previously) always referenced DS==INITSEG.

tlvc/arch/i86/boot/setup.S
```diff
@@ -943,7 +952,11 @@ novga:	mov	%al,14		// CGA 25 rows
 
 // check for hma=kernel in /bootopts, return AX = 0 if not
 hmabootopts:
-	push	%ds
+	cmpb	$6,cpu_type	// ignore if < 80826
+	jae	1f
```
ghaerr, line 956: Same here, why not just have checkhma do this instead of checking twice?

tlvc/arch/i86/boot/setup.S
```diff
@@ -976,10 +989,8 @@ hma_string:
...
 
 // check whether HMA can be enabled, return NC if so with A20 enabled
+// Assumes DS=INITSEG
 checkhma:
-	push	%ds
-	mov	$INITSEG,%ax
-	mov	%ax,%ds
 	mov	xms_kbytes,%ax		//. must have xms size >= 64k
 	cmp	$64,%ax
 	jl	1f
```
ghaerr, line 996: As mentioned in the main comment section, don't XT-class systems set xms_bytes to 0 from the INT 15/AX=8800 call? If so, that would preclude the need for checking cpu_type when XMS memory is not present.

If on the other hand, if this check is needed for 286 systems with HMA boot hanging because XMS is available and A20 gate works, but still the HMA area is not addressable  for some reason, then yes, cpu_type == 286 will have to be checked for here. Then our additional fix of checking for arch_cpu != 6 (80286) before calling check_unreal_mode should stop further hang in xms_init.

Am I on the right track here?

Mellvik:
> > Fix boot failure on XT class systems when HMA support is CONFIGured in (setup.S) - by checking cpu type before accessing some functions
> 
> Where or what exactly is the boot failure on XT? I'm not seeing any issues with booting HMA kernel on XT systems. The requirements for HMA kernel are just 1) getting >= 64k for INT 15h AX=8800h get extended memory size, 2) A20 operational, and 3) hma=kernel in /bootopts, right?

Agreed, I would have expected it to just run, but the boot process goes haywire in `hmabootopts`, and I haven't debugged it further than that for now, deeming it more important to get it running. 

IMG_0548

> 
> > What's really missing still, is a simple way to check whether unreal mode works without hanging the system if it doesn't.
> 
> Agreed that unreal mode can't be currently be "tested for" without knowing that the CPU is 386 or greater. Can't we just fix this with testing for "arch_cpu != 7 || check_unreal_mode <= 0" in xms_init?

I thought so too until a week ago and I tested the COMPAQ 386 (reported in a different thread). It doesn't do unreal mode and hangs exactly like the portable III (286). While the much more recent AMI 386SX system is fine. `check_unreal_mode` as implemented will hang those systems. I don't know if there are other ways to do it that are 'softer'.

> 
> HMA doesn't require 286 or 386 support, but XMS buffers and XMS ramdisk require 386+, correct? 

Not quite. HMA does require at least 286 (I don't know about the 186) because there has to be a A20 address line.

So that should allow XT systems to boot HMA (unless all XT systems don't have XMS, in which case xms_size == 0, which stop loading). That leaves 386+ systems to check for unreal mode for non-HMA XMS services, and 286 systems will likely have to enable INT 15 block move support.
> 
> I was thinking we could compile in CONFIG_FS_XMS_INT15 support and have xms_fmemcpy switch between linear and bios block moves, which should then allow HMA and XMS to be totally automated with the above arch_cpu check fix, right?

If the check_unreal_mode problem can be fixed, that's the way to go!

ghaerr:
> but the boot process goes haywire in hmabootopts

Something very strange in your screenshot - there are no dots after "TLVC Setup" , and the "Copy DEF_SETUPSEG code to high mem" and "Copy DEF_INITSEG data ..." are missing, while then it jumps back to copying .fartext.

I would say it seems as though a mismatched push/pop is happening, perhaps in hmabootopts? Does this happen only on XT and/or 286 systems, or always at this point? I haven't seen any funny stuff with XT systems and they have been reported working. 286 has not been tested. I will look closely at my hmabootopts function now.

Mellvik:
Actually- it seems to me -  there is a bug (unless it's too late in the evening) than may affect all this:
```
        mov     $INITSEG,%ax
        mov     %ax,%ds
        call    arch_get_mem    // save base memory size
        mov     %ax,mem_kbytes
        mov     $0x8800,%ax     // get xms size
        int     $0x15
        jc      1f
        mov     %ax,xms_kbytes
1:                  <---- this label is missing, so if the BIOS call fails, the hambootoipts code will be executed.
#ifdef CONFIG_BOOTOPTS
        call    bootopts        // load /bootopts into DEF_OPTSEG (0050:0000)
        call    hmabootopts     // check /bootopts for hma=kernel
        mov     %al,hma_kernel
#endif
        ret
```

Mellvik:
> > but the boot process goes haywire in hmabootopts
> 
> Something very strange in your screenshot - there are no dots after "TLVC Setup" , and the "Copy DEF_SETUPSEG code to high mem" and "Copy DEF_INITSEG data ..." are missing, while then it jumps back to copying .fartext.
> 
> I would say it seems as though a mismatched push/pop is happening, perhaps in hmabootopts? Does this happen only on XT and/or 286 systems, or always at this point? I haven't seen any funny stuff with XT systems and they have been reported working. 286 has not been tested. I will look closely at my hmabootopts function now.

Exactly my thinking too, but my cursory look don't reveal anything, but maybe the the bug above explains it all.

This happens only on the XT system, and I've tested only one, but that would be enough.

tlvc/arch/i86/boot/setup.S
```diff
@@ -976,10 +989,8 @@ hma_string:
 	.ascii	"\nhma=kernel"		// size = 11
 
 // check whether HMA can be enabled, return NC if so with A20 enabled
+// Assumes DS=INITSEG
```
ghaerr, line 992: Not sure about this assumption and removal of setting and saving DS: when checkhma is called,  DS is not set to INITSEG, its set to SYSSEG. So this likely falsely allows or disallows hma, it would seem to me.

Mellvik: Now that it works (more testing in the morning) I'll just revert these changes.

ghaerr:
> OK this is it, adding in the label and taking out the cpu_type stuff and it runs the XT just fine!

My copy of setup.S has the label:
```
        mov     $INITSEG,%ax
        mov     %ax,%ds
        call    arch_get_mem    // save base memory size
        mov     %ax,mem_kbytes
        mov     $0x8800,%ax     // get xms size
        int     $0x15
        jc      1f
        mov     %ax,xms_kbytes
1:
#ifdef CONFIG_BOOTOPTS
        call    bootopts        // load /bootopts into DEF_OPTSEG (0050:0000)
        call    hmabootopts     // check /bootopts for hma=kernel
        mov     %al,hma_kernel
#endif
        ret
```
Sounds like this was a translation error bringing it over from ELKS!

> Still, let's keep the discussion going on how to get by the unreal check & hang problem.

So then, can we ignore the XT issue then, I am not totally clear. Our testing at ELKS shows that XT systems work well. Likely because the XMS size is returned 0 which stops HMA loading.

With regards to check_unreal_mode, looking at it, it "checks" for a 32-bit CPU - that should rule out 8086 code from hanging. But perhaps this is allowing 80286 code to continue. In that case, lets add a "if (arch_cpu != 7 || ..." before the check_unreal_mode() call in xms_init. Do you agree with that?

> I thought so too until a week ago and I tested the COMPAQ 386 (reported in a different thread). It doesn't do unreal mode and hangs exactly like the portable III (286). While the much more recent AMI 386SX system is fine. check_unreal_mode as implemented will hang those systems. I don't know if there are other ways to do it that are 'softer'.

So even with my above suggested fix, the Compaq 386 would still hang, while the Portable III should not. Correct? That leaves us with only the problem of some 386 systems passing "unreal mode" but not able to use it. IMO, these are exactly the systems for which CONFIG_FS_INT15 is required, right? So that's the final check we need to figure out. ?

ghaerr:
To finalize our discussion, then:

1) none of this PR is actually needed, as it was found a translation error from bringing code over.

2) Adding arch_cpu check for 386 only before calling check_unreal_mode should get 286 systems disabled (and would switch on the compiled-in INT15 support, should we do that)

3) If CPU is 386, its still not safe to attempt setting unreal mode for certain Compaq 386 systems. We need to special case them somehow or determine a way to know their BIOS is running in protected mode. I'm leaning towards the former for the time being. I think there may be a way to know we're running on Compaq BIOS by looking at some stuff near FFFF:0000 (BIOS date and stuff).

ghaerr:
Given #2 and #3 above, one possibility is that we always compile in INT 15 support, and use that when 286 or 386 is present. The only problem is it's not nearly as fast, I presume, as unreal mode. Unreal mode could be set as a /bootopts option for 386 systems only, where it would be allowed for 386+ systems. Then everything would be automatic, although not necessarily fast. But all dynamically configurable.

ghaerr:
Back to 80286 for a minute - I might have gotten a bit confused about exactly where you're seeing a hang

>  I tested the COMPAQ 386 (reported in a different thread). It doesn't do unreal mode and hangs exactly like the portable III (286).

The Compaq 386 will pass the unreal mode check and then hang, as we learned long ago that setting unreal mode ON when using protected-mode BIOSes hangs the system, likely in later BIOS calls, but it doesn't matter - these systems require INT 15.

The Compaq Portable III 286 should not hang in check_unreal_mode, here's the code, and its pretty simple:
```
# Check if unreal mode capable. Currently requires 32-bit CPU (386+)
# Returns 1 if OK, otherwise error code (-1=not 386, -2=in V86 mode).
check_unreal_mode:
        pushf                   # check for 32-bit CPU
                pushf
                popw %bx        # old FLAGS -> BX
                movw %bx,%ax
                xorb $0x70,%ah  # try changing b14 (NT) or b13:b12 (IOPL)
                pushw %ax
                popf
                pushf
                popw %ax        # new FLAGS -> AX
        popf
        xorb %ah,%bh
        xorw %ax,%ax
        andb $0x70,%bh          # 32-bit CPU if we changed NT or IOPL
        je      not_32bit

# check if (32-bit) CPU is in V86 mode
        smsww %bx               # 'SMSW' is a '286+ instruction
        andb $1,%bl
        jne     in_vm86mode

        mov     $1,%ax          # unreal mode capable, return 1
        ret
not_32bit:
```
Since this routine uses an 286+ instruction SMSW, it's definitely not safe for 8086. But it won't ever be called from 8086 since xms_size == 0 from INT 15/AX=8800. Are you thinking that the Portable III hangs here? It should not, but an "arch_cpu == 7" check would prevent that.

Nonetheless, it seems to me that if we have XMS memory, then it would be safe for any CPU (i.e. 286+) to use INT 15 block moves compiled in by default, unless /bootopts specifies xms=unreal, in which case check_unreal_mode would be called, and then enable_unreal_mode if check_unreal_mode returned Ok.

We do have one big problem with using INT 15, instead of using unreal mode by default: the HMA kernel won't work on some INT 15 BIOSes because it disables A20! Given that default HMA kernel is a huge enhancement, I'm now thinking default INT 15 for XMS is not a great idea, and its slow anyways and only needed on Compaq machines. 

So perhaps compiling in INT 15 support, and enabling unreal mode by default unless /bootopts xms=int15 is set would be the best option. All systems except Compaq would run out of the box, and Compaq users would have to uncomment xms=int15 in order to boot. 

If it were possible to check for a Compaq BIOS, xms=int15 could be set by default, and not required in /bootopts, although xms=unreal or xms=int15 could force the issue. If you run `hd FFFF:0000` you should be able to read the last 16 bytes of BIOS ROM on your Compaq systems - we could then determine where the manufacturer code is and use that. I'll look into some BIOS source to find the standard address for that.

Mellvik:
> To finalize our discussion, then:
> 
> 1. none of the setup.S code of this PR is actually needed, as it was found a translation error from bringing code over.

not correct, that was one of several issues addressed in the pr, as pointer out on the pr-notes.

> 2. Adding arch_cpu check for 386 only before calling check_unreal_mode should get 286 systems disabled (and would switch on the compiled-in INT15 support, should we do that)

correct.
> 3. If CPU is 386, its still not safe to attempt setting unreal mode for certain Compaq 386 systems. We need to special case them somehow or determine a way to know their BIOS is running in protected mode. I'm leaning towards the former for the time being. 

Agreed.

Mellvik:
> Given 2) and 3) above, one possibility is that we always compile in INT 15 support, and use that when 286 or 386 is present. The only problem is it's not nearly as fast, I presume, as unreal mode. Unreal mode could be set as a /bootopts option for 386 systems only, where it would be allowed for 386+ systems. Then everything would be automatic, although not necessarily fast. But all dynamically configurable.

True, this is a viable solution - unless we can figure a way to 'fix' the unreal check.

Mellvik:
> We do have one big problem with using INT 15, instead of using unreal mode by default: the HMA kernel won't work on some INT 15 BIOSes because it disables A20! Given that default HMA kernel is a huge enhancement, I'm now thinking default INT 15 for XMS is not a great idea, and its slow anyways and only needed on Compaq machines.

We don't know that this is a compaq thing, it's just the only systems we've tested on. The AMI is a different animal, almost 15 years newer.

I need to revisit the boot process on the compaqs with check_unreal, to identify exactly  what happens when. Now that we're diving into it, I'm optimistic that we'll figure something that works. Tomorrow.

tlvc/arch/i86/lib/unreal.S
```diff
@@ -67,40 +67,68 @@ in_vm86mode:
...
+	push %fs
+	push %gs
+
+	xorl %eax,%eax
+	mov %cs,%ax		# set up the GDT entry for the code seg
+	shll $4,%eax
+	mov %ax,code+2
+	ror $16,%eax
+	mov %al,code+4
+	mov %ah,code+7
```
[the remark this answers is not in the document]
Mellvik: No, this code is creating a suitable GDT entry for the code segment.

The `lret` idea seems to work, thanks.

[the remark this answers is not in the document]
ghaerr: I'm still a bit confused about how this is actually works, from your findings that CS has to be changed in order for protected mode to be entered/exited. I can't quite tell from the above code, from which you said might have been accidentally pushed.

My question is: do we need a new CS selector (like LINEAR_SEL, but something like CODE_SEL) to load into CS when entering protected mode? That would imply that a LJMP wouldn't work because CS would the segment value of CS used in real mode, rather than say "LJMP $CODE_SEL:$1f" which would actually load a new selector.

For returning to real mode, of course a real mode segment value would have to be loaded, for which my LRET idea would work.

What's crazy about all this is that my initial implementation of enable_unreal_mode was from the writings of Chris Giese, who wrote a number of articles and published UNREAL.ASM, which doesn't contain any resetting of CS... ?! It almost seems that "his" unreal mode is in fact some special mode whereas what we're doing now as a result of your findings is a full-blown entry/exit protected mode which is required for the Compaq. What do you think?

PR #161, opened 2025-04-15
[kernel] Fix for enable_unreal_mode

Mellvik:
This PR fixes a long time problem with `enable_unreal_mode`, as discussed in #160 and https://github.com/ghaerr/elks/pull/2274#issuecomment-2804893775. A clip from the latter describes to changes:

> The problem is this: As is, the enable_unreal_mode() routine does nothing. As it turns out, flipping the PE bit does enable, but not turn on protected mode. Until the prefetch queue has been emptied and most importantly, the CS changed, the processor stays in real mode. This is why we never - in the BIOS sources for example - see a switch to/from protected mode without an ljmp immediately following it. The ljmp takes care of both. And the ljmp was missing from our enable_unreal_mode routine.

> The question is: Why does this work in some cases, like on QEMU and my AMI386SX system? The answer seems to be lucky circumstances. QEMU is easy. I commented out the calls to enable_unreal_mode completely, and xms still works just fine. What's presumably happening is that the SEA BIOS is using a full, flat GDT entry for DS and ES, and leaves those in the segment descriptor cache. Since our memory transfer routines expect that simple mapping, everything is fine. And A20 is always on anyway on QEMU.
The explanation for the AMI386 system is likely similar, but on this system behaviour without calling enable_unreal was different, and before I figured out why, I got the new enable_unreal working, so I left it.

> IOW, getting the long jumps into place, the Compaq 386 runs unreal like a charm. I have yet to demonstrate it with HMA kernel, but there is not reason it should not work.

### DIFF
--- a/tlvc/arch/i86/lib/unreal.S
+++ b/tlvc/arch/i86/lib/unreal.S
@@ -95,10 +95,12 @@ enable_unreal_mode:
 
 	lgdt gdt_ptr
 
-	mov %cs,%ax		# self-modifying code, ugh..
-	mov %ax,%ds
-	mov %ax,ur_ret		# set the CS for protected mode return
+	//mov %cs,%ax		# self-modifying code, ugh..
+	//mov %ax,%ds
+	//mov %ax,ur_ret		# set the CS for protected mode return
 
+	push %cs
+	push $unreal
 	movl %cr0, %eax		# CR0.PE=1: enable protected mode
 	orb $1,%al
 	movl %eax, %cr0
@@ -116,9 +118,10 @@ pmode:
 
 	decb %al        # CR0.PE=0: back to (un)real mode
 	movl %eax, %cr0
-	.byte 0xea	# ljmp
-	.word unreal	# address
-ur_ret:	.word 0		# CS, changed by the code above
+	lret
+	//.byte 0xea	# ljmp
+	//.word unreal	# address
+//ur_ret:	.word 0		# CS, changed by the code above
 
 # loading segment registers in (un)real mode changes their base address
 # but not the segment limit; which remains 4GB-1
@@ -412,7 +415,7 @@ code:   .byte 0xff, 0xff, 0, 0, 0, 0b10011011, 0b00001111, 0
 
 # linear data segment descriptor:
 LINEAR_SEL = . - gdt
-        .word 0xFF    # limit 64k - either bytes of 4k pages
+        .word 0xFFFF    # limit 64k - either bytes of 4k pages
         .word 0         # base 0
         .byte 0
         .byte 0x93      # present, ring 0, data, expand-up, writable, accessed
--- a/tlvc/arch/i86/lib/unreal.S
+++ b/tlvc/arch/i86/lib/unreal.S
@@ -31,7 +31,7 @@
 	.global	linear32_fmemcpyb
 	.global	linear32_fmemset
 
-#include "a20.inc"              # A20 gate functions shared with setup.S
+#include "a20.inc"
 
 # Check if unreal mode capable. Currently requires 32-bit CPU (386+)
 # Returns 1 if OK, otherwise error code (-1=not 386, -2=in V86 mode).
@@ -67,40 +67,68 @@ in_vm86mode:
 
 # Put CPU in unreal mode. Requires 32-bit CPU (386+) to call!
 enable_unreal_mode:
-# point gdt_ptr to gdt
+
 	xorl %eax,%eax
 	movw %ds,%ax
 	shll $4,%eax
 
-	addl $gdt, %eax
+	addl $gdt, %eax		# point gdt_ptr to gdt
 	movl %eax, gdt_ptr+2
 
 	pushf			# save interrupt status
 	cli                     # interrupts off
-	pushl %ds
-	pushl %es
-	pushl %fs
-	pushl %gs
-		lgdt gdt_ptr
-		movl %cr0, %eax # CR0.PE=1: enable protected mode
-		orb $1,%al
-		movl %eax, %cr0
+	mov $0x80,%ax
+	out %al,$0x70		# for good measure, turn off RTC NMI
 
-		movw $LINEAR_SEL, %bx # selector to segment with 4GB-1 limit
-		movw %bx,%ds    # set segment limits in descriptor caches
-		movw %bx,%es
-		movw %bx,%fs
-		movw %bx,%gs
+	push %ds
+	push %es
+	push %fs
+	push %gs
 
-		decb %al        # CR0.PE=0: back to (un)real mode
-		movl %eax, %cr0
+	xorl %eax,%eax
+	mov %cs,%ax		# set up the GDT entry for the code seg
+	shll $4,%eax
+	mov %ax,code+2
+	ror $16,%eax
+	mov %al,code+4
+	mov %ah,code+7
+
+	lgdt gdt_ptr
+
+	mov %cs,%ax		# self-modifying code, ugh..
+	mov %ax,%ds
+	mov %ax,ur_ret		# set the CS for protected mode return
+
+	movl %cr0, %eax		# CR0.PE=1: enable protected mode
+	orb $1,%al
+	movl %eax, %cr0
+
+	ljmp $code-gdt,$pmode
+	//.byte 0xea
+	//.word pmode,code-gdt
+	
+pmode:
+	movw $LINEAR_SEL, %bx # selector to segment with 4GB-1 limit
+	movw %bx,%ds    # set segment limits in descriptor caches
+	movw %bx,%es
+	movw %bx,%fs
+	movw %bx,%gs
+
+	decb %al        # CR0.PE=0: back to (un)real mode
+	movl %eax, %cr0
+	.byte 0xea	# ljmp
+	.word unreal	# address
+ur_ret:	.word 0		# CS, changed by the code above
 
 # loading segment registers in (un)real mode changes their base address
 # but not the segment limit; which remains 4GB-1
-	popl %gs
-	popl %fs
-	popl %es
-	popl %ds
+unreal:
+	pop %gs
+	pop %fs
+	pop %es
+	pop %ds
+	xor %al,%al
+	out %al,$0x70
 	popf			# restore interrupt status
 
 	mov	$1,%ax		# system is in unreal mode, return 1
@@ -117,8 +145,7 @@ linear32_peekw:
 	call   setgpf
 	mov    %si,%dx
 
-	//pushf               // save interrupt status
-	//cli                 // uncomment if extended registers used in interrupt routines
+	//cli			// uncomment if extended registers used in interrupt routines
 	xorl   %esi,%esi
 	mov    %sp,%si
 
@@ -132,7 +159,7 @@ linear32_peekw:
 
 	cld
 	addr32 lodsw          // AX = [DS:ESI++]
-	//popf                // restore interrupt status
+	//sti			// restore interrupt status
 
 	mov    %dx,%si
 	mov    %ss,%bx
@@ -151,7 +178,6 @@ linear32_fmemcpyw:
 	mov    %si,%ax
 	mov    %di,%dx
 
-	//pushf               // save interrupt status
 	//cli                 // uncomment if extended registers used in interrupt routines
 	xorl   %esi,%esi
 	mov    %sp,%si
@@ -177,7 +203,7 @@ linear32_fmemcpyw:
 	addr32 rep movsw      // word [ES:EDI++] <- [DS:ESI++], ECX times
 	addr32 nop            // 80386 B1 step chip bug on address size mixing
 
-	//popf                // restore interrupt status
+	//sti		      // restore interrupt status
 
 	mov    %ax,%si
 	mov    %dx,%di
@@ -197,8 +223,7 @@ linear32_fmemcpyb:
 	mov    %si,%ax
 	mov    %di,%dx
 
-	//pushf               // save interrupt status
-	//cli                 // uncomment if extended registers used in interrupt routines
+	//cli		      // uncomment if extended registers used in interrupt routines
 	xorl   %esi,%esi
 	mov    %sp,%si
 
@@ -228,7 +253,7 @@ linear32_fmemcpyb:
 	addr32 rep movsb      // byte [ES:EDI++] <- [DS:ESI++], ECX times
 	addr32 nop            // 80386 B1 step chip bug on address size mixing
 
-	//popf                // restore interrupt status
+	//sti		      // restore interrupt status
 
 	mov    %ax,%si
 	mov    %dx,%di
@@ -247,7 +272,6 @@ linear32_fmemset:
 	push   %es
 	mov    %di,%dx
 
-	//pushf               // save interrupt status
 	//cli                 // uncomment if extended registers used in interrupt routines
 	xorl   %esi,%esi
 	mov    %sp,%si
@@ -277,7 +301,7 @@ linear32_fmemset:
 	addr32 rep stosb      // byte [ES:EDI++] <- AL, ECX times
 	addr32 nop            // 80386 B1 step chip bug on address size mixing
 
-	//popf                // restore interrupt status
+	//sti		      // restore interrupt status
 
 	mov    %dx,%di
 	mov    %ss,%ax
@@ -383,16 +407,19 @@ gdt:    .word 0         # limit 15:0
         .byte 0         # b7-4=flags, b3-0=limit 19:16
         .byte 0         # base 31:24
         
+code:   .byte 0xff, 0xff, 0, 0, 0, 0b10011011, 0b00001111, 0
+//flat:   .byte 0xff, 0xff, 0, 0, 0, 0b10010010, 0b11001111, 0
+
 # linear data segment descriptor:
 LINEAR_SEL = . - gdt
-        .word 0xFFFF    # limit 0xFFFFF
+        .word 0xFF    # limit 64k - either bytes of 4k pages
         .word 0         # base 0
         .byte 0
         .byte 0x93      # present, ring 0, data, expand-up, writable, accessed
 # putting a zero byte here (instead of 0xCF) effectively disables unreal mode:
 #       .byte 0         # byte-granular, 16-bit, limit=64K-1
-        .byte 0xCF      # page-granular, 32-bit, limit=4GB-1
-        .byte 0
+        .byte 0xCF      # page-granular, 32-bit, limit=4GB-1 (was CF)
+        .byte 0		# upper (4th) byte of base address, not used
 gdt_len = . - gdt
 
 gdt_ptr:.word gdt_len - 1
--- a/tlvc/arch/i86/lib/unreal.S
+++ b/tlvc/arch/i86/lib/unreal.S
@@ -94,21 +94,13 @@ enable_unreal_mode:
 	mov %ah,code+7
 
 	lgdt gdt_ptr
-
-	//mov %cs,%ax		# self-modifying code, ugh..
-	//mov %ax,%ds
-	//mov %ax,ur_ret		# set the CS for protected mode return
-
-	push %cs
+	push %cs		# prepare for return to real mode
 	push $unreal
 	movl %cr0, %eax		# CR0.PE=1: enable protected mode
 	orb $1,%al
 	movl %eax, %cr0
 
-	ljmp $code-gdt,$pmode
-	//.byte 0xea
-	//.word pmode,code-gdt
-	
+	ljmp $code-gdt,$pmode	# Set the CS to enter protected mode
 pmode:
 	movw $LINEAR_SEL, %bx # selector to segment with 4GB-1 limit
 	movw %bx,%ds    # set segment limits in descriptor caches
@@ -118,10 +110,8 @@ pmode:
 
 	decb %al        # CR0.PE=0: back to (un)real mode
 	movl %eax, %cr0
-	lret
-	//.byte 0xea	# ljmp
-	//.word unreal	# address
-//ur_ret:	.word 0		# CS, changed by the code above
+	lret		# Effectively an ljmp to the next line
+			# to set the CS
 
 # loading segment registers in (un)real mode changes their base address
 # but not the segment limit; which remains 4GB-1


tlvc/arch/i86/lib/unreal.S
```diff
@@ -383,16 +407,19 @@ gdt:    .word 0         # limit 15:0
         .byte 0         # b7-4=flags, b3-0=limit 19:16
         .byte 0         # base 31:24
         
+code:   .byte 0xff, 0xff, 0, 0, 0, 0b10011011, 0b00001111, 0
+//flat:   .byte 0xff, 0xff, 0, 0, 0, 0b10010010, 0b11001111, 0
+
 # linear data segment descriptor:
 LINEAR_SEL = . - gdt
-        .word 0xFFFF    # limit 0xFFFFF
+        .word 0xFF    # limit 64k - either bytes of 4k pages
```
ghaerr, line 415: I think this still wants to be 0xFFFF, which are the low 16 bits of the segment limit register. This will be the same whether the limit is 4GB or 64K, with the high 4 bits of the limit set in lower 4 bits of the 0xCF byte below. Right?

tlvc/arch/i86/lib/unreal.S
```diff
@@ -67,40 +67,68 @@ in_vm86mode:
...
+	push %fs
+	push %gs
+
+	xorl %eax,%eax
+	mov %cs,%ax		# set up the GDT entry for the code seg
+	shll $4,%eax
+	mov %ax,code+2
+	ror $16,%eax
+	mov %al,code+4
+	mov %ah,code+7
```
ghaerr, line 94: (copied from earlier PR)

I'm still a bit confused about how this is actually works, from your findings that CS has to be changed in order for protected mode to be entered/exited. I can't quite tell from the above code, from which you said might have been accidentally pushed.

My question is: do we need a new CS selector (like LINEAR_SEL, but something like CODE_SEL) to load into CS when entering protected mode? That would imply that a LJMP wouldn't work because CS would the segment value of CS used in real mode, rather than say "LJMP $CODE_SEL:$1f" which would actually load a new selector.

If the above is the case, couldn't we just use the above LJMP rather than all this complicated SHL/ROL code?

For returning to real mode, of course a real mode segment value would have to be loaded, for which my LRET idea would work.

What's crazy about all this is that my initial implementation of enable_unreal_mode was from the writings of Chris Giese, who wrote a number of articles and published UNREAL.ASM, which doesn't contain any resetting of CS... ?! It almost seems that "his" unreal mode is in fact some special mode whereas what we're doing now as a result of your findings is a full-blown entry/exit protected mode which is required for the Compaq. What do you think?

Mellvik: If you look at the first GDT entry, labeled 'code:' you'll see how this works. The code simply fills in the CS content into the GDT. IOW this is the CODE_SEL you're referring to.

tlvc/arch/i86/lib/unreal.S
Mellvik: My first comment to this one seemed to just disappear. 
Yes, should be 0xFFFF, I have been used GDT entry above ('flat:'), commented out now, for testing, just enabled this one before the commit (and it works BTW). Obviously I've had some fun experimenting with GDTs along the way.

Mellvik:
I tested the `lret` method some more, and it's reliable. Included in the last commit.

Mellvik:
BTW @ghaerr - I didn't mention that: The compaq is now running kernel in HMA and XMS buffers just fine.

tlvc/arch/i86/lib/unreal.S
ghaerr: > I have been looking for that particular file 

Here it is, on the Wayback Machine: https://web.archive.org/web/20180414165746/http://my.execpc.com/~geezer/code/unreal.asm

> so the missing ljmps are indeed strange

Indeed, and looking at his original code, there aren't any changes to CS. I am thinking that perhaps there is an "unreal mode" after all - one that sets the limit registers but doesn't actually enter protected mode - but for reasons we don't fully yet understand, don't work on Compaq. This could have to do with the Compaq's using very early steppings of the 386 which don't support an unreal mode as Chris writes about, or perhaps luck on our part that some systems previously setup flat 32-bit segments, or that Chris's code works on later 386's...

In any case, I would say far better for us to fully enter protected mode for whatever reasons, given your having got that method to work on all real hardware you have as well as QEMU!

ghaerr: > The code simply fills in the CS content into the GDT. IOW this is the CODE_SEL you're referring to.

Got it. I think your longer sequence of code entering protected mode might be able to be shorted by using 'CODE_SEL = . - gdt' as a fixed offset (unlike code: now which is a real mode address) from gdt (just like LINEAR_SEL) and then using LJMP $CODE_SEL,$1f; much shorter and quicker.

ghaerr: > then using LJMP $CODE_SEL,$1f

Oops - missed your latest commit... looks you've already effectively done that in a slightly different way.

Mellvik:
> Indeed, and looking at his original code, there aren't any changes to CS. I am thinking that perhaps there is an "unreal mode" after all - one that sets the limit registers but doesn't actually enter protected mode - but for reasons we don't fully yet understand, don't work on Compaq. This could have to do with the Compaq's using very early steppings of the 386 which don't support an unreal mode as Chris writes about, or perhaps luck on our part that some systems previously setup flat 32-bit segments, or that Chris's code works on later 386's...

> In any case, I would say far better for us to fully enter protected mode for whatever reasons, given your having got that method to work on all real hardware you have as well as QEMU!

I think of this as an accidental omission from Kiese, a bad one but still just that. From hundreds of tests - and when we look closely, common sense - there is no other way than the CS replacement - `ljmp` & co - to achieve the mode switch. An easy test (which I did) is to flip the PE bit at some random place in the OS code and let it run for a 'long' time to see what happens. Nothing. Easiest at kernel startup where interrupts are 'under control'. The PE bit is a status bit, not a switch. What I didn't test but would be the ultimate test, is to give garbage to `lgdt`, then set the PE bit. If the memory management system is active, the processor will hang really fast.

The way I picture the mechanism is like this: The PE bit is consulted when segment registers change. If on, the memory management unit is on, otherwise not. Part of the revelation when studying this is that protected mode doesn't turn the processor into '32bit mode'. This is (elegantly) done via the 'D' bit (bit 6) in the segment descriptor, which means that while in protected mode we - the developers - can flip back and forth between 32bit operand code and 16 bit operand code at will and at minimal cost (loading a GDT entry from memory).
 
To me, the 'switching mechanism' makes technical/practical sense. The key difference between real mode and protected mode is memory addressing, ie. using or not using the memory management subsystem. In protected mode, segment registers take on an entirely new meaning, a new role compared to real mode. How, while still running code loaded from/addressed via the current code segment register, can a controlled transfer between modes happen? Flipping a bit in a status register could do it, but only after careful (possibly complex) preparations of where to go next, akin to what we need to do when using INT15 to do XMS block transfers.  A clean break is necessary - and a deliberate jump - whether via ljmp, lcall or lret, is possibly not the only choice, but it's an obvious choice. Quite elegant actually.

What I'm saying is that IMHO this newly acquired understanding is firm and correct, and corroborated but thousands of code examples on the net. It makes sense and it works. And it makes me think (with a smile) - what if we set up a valid IDT, and a GDT that reflects the current state of the system, would TLVC (and ELKS) run in protected mode? It would take a little more than that, but it's an interesting thought. What's interesting about it isn't its usefulness which is probably dubious, but the understanding (of practical protected mode) that would be required to get it working. How tempting.

ghaerr:
@Mellvik: I'm not in (full) agreement with your analysis, and there's a lot of different things being discussed in this thread.

I thought to respond at length on the discrepancies/disagreements, so you can see where I'm coming from, so we don't end up both making stuff up that proves to be false, as has happened at least twice in the recent past, trying to make sense of what we're seeing.

First, there's no question you've found a solution to getting the prevously-stubborn Compaq XMS working, and that is through enabling and fully switching to protected mode, setting the DS and ES segment limits through the GDT, and then switching back to real mode.

Second, I take back my previous statement that "unreal mode isn't a mode at all, its the way all 386+ CPUs use the shadow segment limit registers values on add32r prefix instructions". There seems to be no question that in real mode the 386 always uses the shadow limit register values on addr32 prefixed instructions. But I believe there is a separate "unreal mode", separate from real mode and protected mode.

> The problem is this: As is, the enable_unreal_mode() routine does nothing.

Disagree. 

> I think of this as an accidental omission from Kiese, a bad one but still just that. 

Disagree.

> and when we look closely, common sense - there is no other way than the CS replacement - ljmp & co - to achieve the mode switch. 

Agree, but not common sense. Fully agree that the prefetch queue needs to be empty and CS reloaded - to enter protected mode. But not to enter "unreal mode". There is a difference, this is not a "mistake" on Chris Giese's part.

> What I'm saying is that IMHO this newly acquired understanding is firm and correct

No.

I believe there are many things happening here, and lots of factors, especially when we're talking about the differences between emulators, real hardware, CPU versions, BIOSes, etc.

There is an "unreal mode" which absolutely works without reseting CS/LJMP, etc. This is firmly demonstrated in the ELKS application program elkscmd/sys_utils/unreal16.S (`/bin/unreal16`). The program, which uses my (Chris's) enable_unreal_mode code, sets up unreal mode, then draws to B8000 on the PC screen using addr32 and no CS reset/LJMP. 
Try it.

> Why does this work in some cases, like on QEMU and my AMI386SX system? The answer seems to be lucky circumstances. QEMU is easy. I commented out the calls to enable_unreal_mode completely, and xms still works just fine. 

The QEMU case could be a QEMU bug, or could be a case of a previous flat descriptor having been loaded in protected mode. I don't know. When I first wrote XMS support, the 386 desktop I tested it on did not work until I got the routine right. The enable_unreal_mode is in fact setting "unreal mode" which does in fact work on certain systems. Agreed on other systems it may not be needed, perhaps on your AMI386SX where a previous flat descriptor was loaded. I don't know, lots of details.

So, what is going on here? Here are my latest thoughts:
- For reasons yet unknown, the Compaq requires a protected mode switch to increase the segment limits, unreal mode alone won't work.
- Enable_unreal mode works on most systems, reference /bin/unreal16. The routine is not a NOP.
- This PR isn't a "fix" to enable_unreal_mode. It's another routine entirely, using protected mode to setup the segment limits used later in real mode. (Thus the routine should be renamed, but I don't really care).

It appears that "unreal mode" is that period of time when the CPU has PE set, but has not cleared the prefetch queue nor reloaded CS. When segment registers are (re)loaded during this period of time, the "shadow segment limit registers" are set with the loaded GDT's limit value, and remembered past when the PE bit is turned off. Note that during this time, the segment registers are not acting as selectors, although a GDT "selector value" is loaded into a segment register and then popped after PE is turned off. Protected mode is never entered, but unreal mode is effective. 

Why unreal mode is ineffective on Compaq - I don't know.

I'm not sure why some folks recommend using unreal mode, rather than going into protected mode and back, for setting up the shadow limit registers. There is much talk about unreal mode, so there's probably a good reason. Perhaps its faster. Not sure.

More on your protected mode thoughts in the next post...

ghaerr:
> A clean break is necessary 

> Quite elegant actually.

I'm leaning more towards the idea that the design choices that were made during the transition out of real mode into a (actually too complicated combined segmentation and paged memory management) hardware protected system were actually necessary, and not elegant, when you get into the nitty gritty of it. Agreed, if the CPU didn't wait until CS was reloaded to change modes, it'd require loads of complicated and possibly not possible software setup just to get a CS selector value that happened to match the executing CS without crashing. Intel definitely over designed the 386 with all the complex segmentation and selectors/gates/tasks etc, with the end result that very few operating systems used them; they just setup flat mode and used paging. I'm not sure what this over-design might have to do with early chip bugs and/or undocumented unreal mode issues we're seeing on Compaq.

> what if we set up a valid IDT, and a GDT that reflects the current state of the system, would TLVC (and ELKS) run in protected mode?

No. There's far too much code and assumption regarding segment register values being set as a result of address arithmetic, rather than being a selector which has no builtin address component.

However, if you're talking about V86 protected mode, or "8086 protected mode", yes. This CPU mode seems to be a variation of having an GDT with the G bit (4k/byte granularity) = 0 and D bit (32/16-bit address default) = 0, combined with using segment registers as addresses rather than selectors and protected instructions. I recently found source to a complete V86 monitor to implement something like this. The biggest problem is that a large number of instructions are "protected" and cause an interrupt that is required to be handled in the monitor, then adjusting the "real mode" stack frame with the allowed results, and RETI to allow the guest kernel to continue.

> This is (elegantly) done via the 'D' bit (bit 6) in the segment descriptor

From what I've read, there are all sorts of exceptions to this "rule" when it comes to code (CS) and stack (SS) segments. There are related complexity problems when pushing or popping IP vs EIP and which stack to use which override a simply elegant D bit.

ghaerr:
> for good measure, turn off RTC NMI

I'm not familiar with port 0x70: is this port number standard on most/all PC's, or is this something related to an add-on RTC card?

Mellvik:
> > for good measure, turn off RTC NMI
> 
> I'm not familiar with port 0x70: is this port number standard on most/all PC's, or is this something related to an add-on RTC card?

Wow @ghaerr, this is becoming really interesting. I'm still devouring your thinking above, let me just 'eliminate': 0x70 is the CMOS/RTC, which is connected to the processor NMI and may be used as a low res clock, watchdog timer etc. I forget the details and put it in there as a reminder. The COMPAQ BIOSes does this every time they switch modes - obviously a precaution, so I decided it's too important not to keep until investigated.

Mellvik:
@ghaerr, I see your points and agree with some of them (and I enjoy the discussion). I'll address most of it later when I've given the issues some more thought. My first line of reasoning is this:

- I think your initial 'characterisation' of unreal-mode was fair. There is no unreal-'mode', we're simply talking about a controlled way to take advantage of processor capabilities outside the realm of the '16 bit processor emulation mode'. An unintended but convenient consequence of processor design choices. 
- I take back the statement that enable_unreal_mode does (did) nothing. It did. It loaded the GDT. And that's where the grey area starts.
- I think this whole discussion comes down to that simple question: Does `lgdt` get acted upon without a switch to protected mode? My previous post assumed not, your points and observations make me having 2nd thoughts about that. In fact, most of what you describe can likely be explained that way (and tested, which I'm looking forward to). It is entirely possible that some, maybe all ia32 and up processors actually process and use a new GDT after loading it - it may or may not require the PE bit to be flipped, and it may or may not be processor dependent. And it may be code dependent - as in the distance in code bytes between the `lgdt` and reloading CR0. In the latter case it may even be dependent on the size of the prefetch pipeline. It will be interesting to test some of these theories. 


More later.

ghaerr:
Yes, more testing on real hardware and emulators hopefully will illuminate more things. As far as emulators and BIOSes go, there's likely quite a lot of variability compared to different processor revs when it comes to unreal mode. Nonetheless a table of results will come in handy when comparing results across differing emulators or BIOSes.

> to that simple question: Does lgdt get acted upon without a switch to protected mode

Too simple: I'm sure the internal LGDT register gets updated, but in unreal mode, I think the action happens when a segment register is loaded AND the PE bit is on (reading that from the unreal.S code). There separate variables in this one statement.

> dependent on the size of the prefetch pipeline

Improbably IMO, but testing will show. My guess is the prefetch queue clearing is required since there may be an addr32 prefix in it which was changed when new GDT was loaded for the segment with the applicable D bit set differently. Thinking about it, all these little details have to be handled in a known orderly fashion to achieve computational determinism. The CPU has to deliver the same results based on the published ISA, and the internal pipeline architecture is not part of the that, other than having to clear it along with other invalidations, like TLB, in certain circumstances. This is one of them. It is not in the manufacturer's interest to publish too much as ISA, as then application programs that use the feature have to be maintained compatible in future CPU revisions, which might be very costly.

Mellvik:
> Yes, more testing on real hardware and emulators hopefully will illuminate more things. As far as emulators and BIOSes go, there's likely quite a lot of variability compared to different processor revs when it comes to unreal mode. Nonetheless a table of results will come in handy when comparing results across differing emulators or BIOSes.

I don't know about emulators - my guess is rhat there is quite a bit of variation at thislevel. BIOSes are simpler, the don't deal with unreal mode. They (those I've seen) switch into and out of protected mode, always using same procedures.

> 
> > to that simple question: Does lgdt get acted upon without a switch to protected mode
> 
> Too simple: I'm sure the internal LGDT register gets updated, but in unreal mode, I think the action happens when a segment register is loaded AND the PE bit is on (reading that from the unreal.S code). Three separate variables in this one statement.

@ghaerr, Let me remind you, I'm approaching this from a factual side, corroborated by Intel's documentation and scores of other sources. You are speculating - which I am open to, but it is really unhelpful to pretend assumptions are facts. It's better to stop right here and let the facts rule. 

I'll be back when I've verified what can and cannot be done with the PE bit on and no CS change. Then we can take it from there.

ghaerr:
> What I'm saying is that IMHO this newly acquired understanding is firm and correct

>  your points and observations make me having 2nd thoughts about that. 

> I'm approaching this from a factual side, corroborated by Intel's documentation and scores of other sources. 

> You are speculating 

I'm pretty sure we're both speculating here, very few "facts" on the table. That's the whole point of the continued investigation. Please don't assume everything you've found is "fact" and that I'm the only one speculating, there are many still unanswered questions in this thread. :)

rafael2k:
It is really interesting and informative all these findings about the unreal mode. It is crazy how much engineering Intel put in the development on their first 32-bit x86 while allowing many execution modes.

Mellvik:
> It is really interesting and informative all these findings about the unreal mode. It is crazy how much engineering Intel put in the development on their first 32-bit x86 while allowing many execution modes.

Indeed, @rafael2k - I think it is safe to say, as @ghaerr alluded to above, that Intel went overboard with some/many of the  functions in the 386. Very sophisticated functionality that ended up not being used - either because it wasn't all that useful, too complicated or there was a simpler way of doing it. Not as good by some metric, but good enough seen from the developer side. This 'overdosing' is also understandable - driven by competition. Intel had a lot of catching up to do at the time - to match the 68010 in particular, but other processors as well. And there managed to pull it off, thanks to their stubbornness about backwards compatibility.

Actually, the going overboard with advanced functionality thing is more like the rule than the exception in processor design, even relatively simple microcontrollers like the 8390 that powers the ne1k/ne2k/WD80x3/3c503 ethernet controllers, not to mention the Intel 82586 family powering the EtherExpress line. Or the 68030 processor - the list goes on. What happens/happened in many cases was that the world changed faster than the design-to-chip.to-product process - even back in the late 80s. Like, most of the mentioned examples have great (as in sophisticated) memory management functions designed to optimize the use of a very precious resource. Which by the time they entered the market, weren't as precious anymore. Etc.

What's fascinating about really understanding how these things work at the lowest level is that only then can we take full advantage of them in a sophisticated way - like 'unreal mode', which is inspiring to some of us.

ghaerr:
> It seems to me that even facts are being questioned or subjective these days

I hadn't thought of it that way, but I think you're right - there's a societal side to this going on nowadays, where traditions and previously held norms are being questioned.

Regarding facts vs speculating, and to close the discussion on that topic, I want to add that my work on ELKS is pretty much speculative all the time, and not necessarily corresponding to the real world, or "factual". This is because almost all of the design and enhancements I've made to ELKS over the past five years have been almost entirely done on top of a software emulator, with very few excursions onto real hardware. Heck, when I first found ELKS, it was impossible to even build on macOS, ELKS was rarely run on an emulator, and it was completely unreliable on real hardware. Contrast that to the work being done here on TLVC, where development and testing on real hardware is paramount, with real hardware testing continuing to bridge the gap from emulation. So we're sometimes coming from opposite ends of the spectrum, but I think that's a very good thing, and I feel lucky I can continue to develop and learn how my work actually fares in reality, on real hardware, through our association. In summary, TLVC and ELKS are a good combination. Thanks @Mellvik!

>  here's what this small world looks like from this end:

Nice summary, agree with all points. The comment that protected mode has likely already been entered on most systems by the time our kernel gets control stands out as an important additional factor in trying to get the bottom of the current investigation.

I think we're all agreed on how protected mode needs to be entered, and how the segment descriptor cache continues after exit. More fuzzy to me, and likely unreliable in emulation, is the "unreal mode" detail of whether the segment descriptor cache can be altered by just setting the PE bit and writing a segment register: at cold power-on, versus after having previously entered/exited protected mode, for instance.

The fact that Chris Giese wrote UNREAL.ASM, which now exists as a binary unreal16 (and works), doesn't necessarily prove that his "unreal mode", that is, setting the segment cache descriptors w/o entering protected mode, works at cold power-up. It might only work after the CPU was once in protected mode. Or it might actually work all the time. Or it might not actually do anything, with his code working only because it was tested on real hardware that had already set the limits. 

To me, the above "unreal mode" issue is the biggest outstanding issue that really matters, which I don't think we know what is happening on real CPUs. And we need to keep in mind that early CPUs, like those of the Compaq, had a number of bugs that were later changed. I'm not really sure how to effectively test unreal mode on real hardware without having the BIOS interfered with jumping into protected mode first, unless a system can be found with a 386 CPU that doesn't support extended memory through BIOS.

Getting to the bottom of whether the segment cache can be set without entering protected mode will help in answering why the Compaq didn't work before the full PM entry/exit code was added. IMO, it will also answer whether there's a separate "unreal mode" that people keep writing about, or whether that's ultimately a mischaracterization of the fact that the segment cache remains valid after PM exit.

Since I am also interested in learning just for knowledge's sake, I plan on (re)reading about unreal mode(s), and will also look at Shanley's book, thanks.

Mellvik:
Thanks @ghaerr - that should put us back on the same page. And I agree as to the benefits of the 'dual platform' scenario created by TLVC and ELKS.

I've put some more effort into the still open questions in this scenario:
- The AMI386SX does indeed allow access to the segment register cache without entering protected mode, explaining the Kiese code and why the `enable_protected_mode` code worked on some systems before. It is reasonable to assume, but still speculative(!), that similar behaviour applies to all later x86 processor variants. It is also possible, although less likely, that this applies to the 386SX only, and even to this specific version of it which isn't a processor per se, but a SoC made by AMI. It may be argued that this IS protected mode, since the PE bit is set. It may also be argued, and this is my position, that the processor is still running real mode code from a real-mode-addressed code segment, which fail to qualify as 'Protected Mode'. IOW, we're still running real mode. Naming aside, the fact is that some processor-variants allow this, some don't, and entering protected mode is required to cover all. It's 'cheap' and safe.
- After spending some time thinking about and experimenting with the GDT in lieu of our BIOS-related discussion, the conclusion is that whatever the BIOS does in or with protected mode, GDTs and segment caches are of cursory interest at best and not practically useful. For a while there I had the misguided idea that since even the 286 BIOS will have to have entered protected mode if it has extended memory, the segment register cache could be reused and we'd have 'free' access to extended memory w/o `loadall` or a full PE-switch followed by a reset. Misguided because even if we can look at the last used GDT via the `sgdt` instruction (I tested that, it works fine - with caveats), there is no way to find out which GDT entry is associated with which segment register. So it's a dead end regardless.   
- A final point about the NMI: I mistakenly associated this with the RTC which used port 0x70 for all commands. This quote from the OSdev Wiki explains it: "On the XT the NMI can be masked by setting bit 7 on I/O port 0xA0. On the AT the NMI can be masked by setting bit 7 on I/O port 0x70. This port is shared with the CMOS RAM index register using bits 0 through 6 of I/O port 0x70. The CMOS RTC expects a read from or write to the data port 0x71 after any write to index port 0x70 or it may go into an undefined state. There may also need to be an I/O delay between accessing the index and data registers. ... " --- I've seen the claim that port 0x71 should always be accessed after port 0x70 in order to ensure a predictable state, before. I've ignored it in the code because this is what COMPAQ does in their BIOSes. That may, however, not be a smart choice. Adding an `in 0x71,%al` is cheap insurance against unpredictable behaviour on some systems - even if they're few and far between. IOW I plan to add it to the code.

That concludes my work on unreal and protected mode for now. Some pending UMB cleanups - and this more or less accidental marathon that started with UMB support has passed the finish line. Thanks for joining - and participating.

ghaerr:
> The AMI386SX does indeed allow access to the segment register cache without entering protected mode

Very good to know, thanks. I'll call that unreal mode.

> It may also be argued, and this is my position, that the processor is still running real mode code from a real-mode-addressed code segment, which fail to qualify as 'Protected Mode'. 

Yep.

> It is reasonable to assume, but still speculative(!), that similar behaviour applies to all later x86 processor variants. 

I would agree.

> the fact is that some processor-variants allow this, some don't, and entering protected mode is required to cover all. 

I would agree.

My take is that every 386 we've seen and experimented on, except the Compaq 386, works in unreal mode. I am not sure, but guessing it may also be one of the earliest Intel 386 processors to hit the market. Compaq's own technical reference documents many 386 problems which were later fixed in later steppings.

Finding that the Compaq would set the segment register cache via full protected mode was very important, and a big surprise to me, as I had never considered just fully enabling PM instead of using the undocumented unreal mode. Using an undocumented CPU feature is never guaranteed to work.

I suppose the reason for this is that I didn't know that entering PM allowed the segment register cache to be accessed after return to real mode. I had read and thought that that feature was an undocumented unreal mode feature only. Now I believe from your research quoting various sources that this latter feature is documented by Intel, thus guaranteed to work.

So there we are. Thanks for your findings. I am curious though - did you ever find what actually happens when running the original enable_unreal mode on the Compaq? Does the CPU crash on first access using addr32, or something else?

There is a simple answer to my curious question which would give us a good reason to see an immediate crash on the first REP MOVSB execution though. And that is, on the Compaq 386 chip, attempting to set the segment cache via unreal mode does exactly nothing. The first XMS buffer to be used, in normal configuration is the one at exactly 1M; this address is all zeroes except for bit 20. The ramdesc_t address for that buffer would be 0x00100000. With the segment register limit not changed, that is, set to the default of 0xFFFF, the ADDR32 REP MOVSB would immediately copy to low core, wiping the interrupt vector table. The first hardware timer interrupt would crash the system. Since we're done with researching this rabbit hole, I'm going to go with that explanation for now.

On a final note - I notice in your merged code there is a bit set seemingly incorrectly in the code segment descriptor, in particular the first of two bytes specified in binary, the access descriptor the byte. In particular, bit 3 is documented as "must be 0", but it's a 1. The normal value is 0x93, yours is 0x9B:
```
code:   .byte 0xff, 0xff, 0, 0, 0, 0b10011011, 0b00001111, 0
```
Is there a reason for this?

Mellvik:
> Heck, when I first found ELKS, it was impossible to even build on macOS, ELKS was rarely run on an emulator, and it was completely unreliable on real hardware. 

Indeed @ghaerr. It has been an incredible journey. You've turned a soap box car into an efficient, well equipped while still small sports car.
 
> So we're sometimes coming from opposite ends of the spectrum, but I think that's a very good thing, and I feel lucky I can continue to develop and learn how my work actually fares in reality, on real hardware, through our association. In summary, TLVC and ELKS are a good combination. Thanks @Mellvik!

I fully agree. The differences in environments, viewpoints, experience, even understanding come in extremely handy, not only when solving problems but when making choices, decisions. A great combination. Thanks.

Mellvik:
> On a final note - I notice in your merged code there is a bit set seemingly incorrectly in the code segment descriptor, in particular the first of two bytes specified in binary, the access descriptor the byte. In particular, bit 3 is documented as "must be 0", but it's a 1. The normal value is 0x93, yours is 0x9B:

> `code:   .byte 0xff, 0xff, 0, 0, 0, 0b10011011, 0b00001111, 0`
> Is there a reason for this?

this bit IIRC indicates that it is a code segment. Check the osdev.wiki web page on unreal mode, it's really good (I use that source quite a bit). In this context, I  don't think it matters much, but I haven't tested. As is, it is correct.

ghaerr:
> this bit IIRC indicates that it is a code segment. 

Ok, finally got it figured out. Turns out the UNREAL.ASM documentation on the access descriptor byte is incorrect, saying that the bit value 0 is required. OSDev documentation shows this as the E)executable bit. I'll be updated unreal.S and will link you to it so you can see the documentation I'm correcting in that file.

Mellvik:
Thanks @ghaerr - this is useful as it identifies some areas which we view things differently. So in order to avoid future misunderstandings, let me clarify my position on these points or issues.

> My take is that every 386 we've seen and experimented on, except the Compaq 386, works in unreal mode. I am not sure, but guessing it may also be one of the earliest Intel 386 processors to hit the market. Compaq's own technical reference documents many 386 problems which were later fixed in later steppings.

First, I don't know of any 386DX system this has been tested on other than the Compaq. Second, the Compaq - and the 386DX works flawlessly in unreal mode. Unless you're redefining 'unreal mode'. I'm sticking with the 'traditional' definition as per the likes of Wikipedia and OSDev etc. Not really a mode but an 'enabled capability'.

> Finding that the Compaq would set the segment register cache via full protected mode was very important, and a big surprise to me, as I had never considered just fully enabling PM instead of using the undocumented unreal mode. Using an undocumented CPU feature is never guaranteed to work.

Again, this has nothing to do with Compaq. And I agree, there is only one reliable way to have a 386+ enter unreal mode. I consider the 'Kiese method' an anomaly, as discussed before - useful knowledge, then put aside as 'not practically useful'. I have not seen any code-example anywhere except his, that miss/omit the ljmp/lret. 

> I am curious though - did you ever find what actually happens when running the original enable_unreal mode on the Compaq? Does the CPU crash on first access using addr32, or something else?

It simply hangs at the first (32 bit mode) memory access. As we've seen, it might have worked accidentally had the BIOS used a GDT entry with the 'right' content in the 'right' spot. It doesn't, and the reference either goes off into never-ever land, or triggers an exception - either way, the machine needs power-cycling every time. As does the AMI machine in similar situations - I've experimented quite a bit with various GDT entry size values and flags, segment bases etc. There have been hundreds of power cycles, mainly using the AMI because it's faster. Very educating. Very tedious. And no, there is no difference in behaviour between the processors - except that the AMI accept the 'Kiese method'.

Mellvik:
One more thing @ghaerr, before completely closing this 'project'.

I don't know about ELKS, but TLVC is querying the BIOS to get the available XMS size and we are incorrectly assuming that it starts at 1M (0x100000). However, the Compaq - and presumably other systems from that day and age - provide XMS even on systems with only 1MB installed. As part of the remapping of all ram above 640k to just below 16M is the A0000-E0000 256k block which becomes XMS. So the response to our size query will be 256, but is it possibly double-mapped like the ROMs?

In order to find out, I really need to dissect a machine and pull most of the memory out. That's going to take a while, but - although slightly less reliable - the XMS_START_ADDR can be set to 0xFA0000, which turns out to work. Is this memory also mapped to 0x10000? The machine has 10MB total, so there will definitely be memory at 0x10000, otherwise HMA would not work. So looking at the HMA segment, currently unoccupied, I should either see all zeroes (unused memory) or the contents of the L2 buffers allocated at FA0000. It does the former. 

While not an unequivocal proof, the reasonable conclusion is that while the 'get XMS size' BIOS call is reliable, there is no guarantee that XMS memory starts at 1M. A test is needed in the XMS code to verify where its location in physical memory, and - to be sure - whether it is a contiguous segment. Also, using the reported XMS size to enable the use of HMA will fail in such cases.

ghaerr:
Hello @Mellvik,

> this is useful as it identifies some areas which we view things differently. So in order to avoid future misunderstandings,

Thank you so much for bringing this up! Please see below for explanation.

> Unless you're redefining 'unreal mode'.

I finally figured out what has happening with our recent conversation(s): yes, I am defining 'unreal mode' differently than your definition, and that has caused all sorts of confusion. I am so sorry I didn't pick up on this earlier!!! I actually think we are in agreement as to the results of both our findings, but my communication has used the wrong terms and I believe our communication has thus been messed up.

If you replace all of my statements using "using/in unreal mode" with "using/after the Giese method", then that better communicates what I have been trying to say. That is, my definition of "unreal mode" has been "the Giese method". And I've finally realized that is definitely not your (or others) use of the term "unreal mode". My apologies.

Some background, and then I'll stop, but I think worth explaining: When I first added XMS support to ELKS, I knew very little about "unreal mode" and less about segment cache registers. After searching the net for code, I came up with Chris Giese's (BTW not Kiese) UNREAL.ASM which implemented "unreal mode". This is where I got and continue to use the term "unreal mode". As you well know, his routine and ELKS original routine does not reset CS, but that's what I've been calling "unreal mode" this whole time. I am going to stop using the phrase "unreal mode" entirely to avoid future confusion.

Now, I will (re)explain what I believe are our conclusions without using the phrase "unreal mode":

To me, it goes without saying that once the internal segment limit cache registers have been actually set, they remain set in real mode, and addr32 prefixes can be used to take advantage of large offsets to access XMS memory. Without this, XMS could never work on ELKS they way it's been done.

So that only leaves *how* to set the segment limit cache registers, not "what mode" do they work in. I believe we have seen two ways the segment limit cache registers can be set on real hardware:
- Method 1: Load GDT, Set PE bit in CR0, Push then load segment register with data selector, unset PE bit, pop segment register.
- Method 2: Load GDT, Set PE bit in CR0, Execute LJMP to reset CS, Push then load segment register with data selector, unset PE bit, Execute LRET to reset CS, pop segment register.

The only difference between the two methods of setting the segment limit cache registers are the LJMP and LRET which clears the prefetch cache and reloads CS.

Finally, the important part: All 386 processors support setting the segment limit cache registers to be later used in real mode. Until now, all 386 processors we've tested on, except those in early Compaqs, have worked using Method 1. The CPU in the Compaq requires Method 2. Other processors may too, I have no idea. In conclusion, since both ELKS and TLVC desire to guarantee that XMS works on all 386 processors, only Method 2 will be used for enabling XMS.

Does that cover it?

>  I have not seen any code-example anywhere except his, that miss/omit the ljmp/lret.

Can you point me to other sources setting the segment limit cache registers using LJMP/LRET? I'm sure you've seen that in the Compaq BIOS listing but I haven't found that and have not found any other sources at all. I'm continuing to research protected mode and GDT's in general and would love to see whatever you've found.

On a similar topic, I've found a new book "Design of OS/2 by Michael Kogan", the lead architect of OS/2. This has some seriously interesting information on 16-bit protected mode and all sorts of stuff that was designed into OS/2 v1.0, which was essentially all 16-bit protected mode, not 32-bit. We shall have to discuss that later, it sheds tons of history on the partnership between IBM and Microsoft with Intel, and kind of shows how lots of the GDT bits came into being, trying to get this OS operational.

> TLVC is querying the BIOS to get the available XMS size and we are incorrectly assuming that it starts at 1M (0x100000)

Yes, ELKS does the same. Agreed this needs to be enhanced to find actual start address of XMS memory instead of assuming 1M.

> but is it possibly double-mapped like the ROMs?

I didn't quite follow this - did your experimentation with mapping XMS to 0xFA0000 and probing HMA show the high memory is doubled mapped to 1M?

> A test is needed in the XMS code to verify where its location in physical memory, and - to be sure - whether it is a contiguous segment. 

A big step forward here would be to use BIOS function E820h which returns a full system memory map as an array, showing the start, length and type of memory for the entire system. The bad news is E820 is not implemented on earlier systems, so for those systems we still have the same problem as today. I don't think the Compaq supports E820, so what kind of fallback do you think would work, or are we back to specifying this somewhere (not another /bootopts I hope lol :)

Mellvik:
Thank you @ghaerr, that was very clarifying. 

>> I have not seen any code-example anywhere except his, that miss/omit the ljmp/lret.

> Can you point me to other sources setting the segment limit cache registers using LJMP/LRET? I'm sure you've seen that in the Compaq BIOS listing but I haven't found that and have not found any other sources at all. I'm continuing to research protected mode and GDT's in general and would love to see whatever you've found.

Actually, the `lret` is ours, I haven't seen that anywhere. The Compaq BIOSes use `ljmp`s only - with segment override (cs:) when entering, literal immediate when returning. I couldn't get the lret to work at first, thus the self modifying ljmp code. Your example made my ('$') error obvious and we had a much better solution. Switching between Intel/MS assembler syntax and `gas` does bad things with my head. BTW I might post the Portable BIOS, rough as it is, for reference.

>  Until now, all 386 processors we've tested on, except those in early Compaqs, have worked using Method 1.

I would be interested in which processors this might be - and the origin (like Intel or other), for reference/my curiosity. Also it may be worth commenting that the processors in the Compaq Portable 386 series are 3rd generation, not really early. The 1st generation  ran at  12.5MHz (which ncludes the first Compaq Deskpros), the 2nd clocked in at 16MHz, also used in Deskpros, while our Portable have the 20MHz version. While these version may be production process refinements only, we can be sure that they would fix errors at the same time. What I'm saying is that I think the 386 bug lists we both have looked at from time to time (e.g. Robert Hummel's The Processor and Coprocessor, which I use a lot, devotes a full chapter to "Incompatibilities and Bugs") is complete. That said, we haven't encountered any errors, so the observation is more like a curiosity.

> On a similar topic, I've found a new book "Design of OS/2 by Michael Kogan", the lead architect of OS/2. This has some seriously interesting information on 16-bit protected mode and all sorts of stuff that was designed into OS/2 v1.0, which was essentially all 16-bit protected mode, not 32-bit.

Thanks for the hint about the book! This is indeed interesting. I'm assuming its focus is on the 286, since that was OS/2's target and market? That's incidentally exactly where I've been spending the last 24 hours of reading. Trying to find a shortcut to include the 286 into our 'new unreal world' without going the LOADALL route. Doesn't look good BTW, but a lot of learning, including that there is in fact a (secret) STOREALL instruction too. Which does exactly what you would think it does. 
What I've concluded for now, is that the INT15 block moves we use for 286 systems, need to move to the fartext segment, like you alluded to once before. Then a20 can be reactivated before returning to the HMA segment. I'll do a test implementation of that next week.

> > but is it possibly double-mapped like the ROMs?

> I didn't quite follow this - did your experimentation with mapping XMS to 0xFA0000 and probing HMA show the high memory is doubled mapped to 1M?

No, this is just speculation, and since writing it, I have confirmed that that is not the case.

>> TLVC is querying the BIOS to get the available XMS size and we are incorrectly assuming that it starts at 1M (0x100000)

> Yes, ELKS does the same. Agreed this needs to be enhanced to find actual start address of XMS memory instead of assuming 1M.

I'm working on this right now (and noticed the thread that mentions this on ELKS). First, the BIOS call returns a 16 bit value in kbytes above 1M, so the max it can report is 64M. I don't know how BIOSes on newer systems deal with this, but TLVC for sure cannot handle it, and I'm tempted to just cap the return value at 32M. For now anything above this is barely useful anyway. 
More importantly though: The returned value from the BIOS may be correct and below 32M, but the memory may be unavailable (to us). My 386SX system (errata: I've called it an AMI system, it's not. The BIOS is AMI, the processor, really a SoC, is the ALI M6117C) when booting TLVC, reports 
`xms: 19392k available, a20 status 1, using unreal mode, EXT bufheads`
Interesting given that this is an SX with only 24 address lines and thus max 16M. I've been wondering about that for a while and took a dive to find out last night.  It turns out many of these systems - including this one, inside the SoC -  use a 74LS612 memory mapper to extend the address range - beyond the address lines of the processor. 
Nice, but it obviously needs a mechanism to manipulate/drive that mapper, and TLVC does not have that mechanism, so the max is 16M-1M=15M, regardless of what the BIOS 'has access to'. Thus the XMS 'sizer' I wrote yesterday came in even more handy than expected.

ghaerr:
> Interesting given that this is an SX with only 24 address lines and thus max 16M.

Wow, not sure about how to handle this?! Why/how could the BIOS report more memory than the CPU can handle, especially since it's supposed to have tested and counted memory beforehand. Are you thinking we might require a routine to tell us the specific type of 386 we're running on, (not sure this exists), or whether it'd be safer to just cap XMS at 16M?

The other option that needs to be tried is to use E820 for those systems that have it and display the memory map produced, to see whether the BIOS knows more than it can tell us via the basic INT 15 8800h extended memory function.

I have some other comments from above when I have a bit more time, thank you!

Mellvik:
> Wow, not sure about how to handle this?! 

@ghaerr - I find this more amusing than anything else. I have ideas about how they do this, but it doesn't really concern us. We just stop at the 'physical' limit and leave the rest. No big deal really, just important to know - and 'take down' so to speak.

btw, and on a different (related) note, one of the things fixed today was the assumption in setup.S that the HMA exists if the size returned from the xms-size bios call is > 64k. The plan is to put together 1M portables, III and 386, for testing in the next few days.

ghaerr:
> the hint about the book! This is indeed interesting. I'm assuming its focus is on the 286

Actually its got large chapters on both the 286, history and design, as well as the 386, since the book covers a number of OS/2 releases well into the 386 era. They dive into very esoteric things like LDT tiling and 16:32/32:16 thunks to move between real, protected 16, and protected 32-bits, all required back in the day when trying to emulate DOS as well as DOS extenders all the while moving to a 32-bit OS. Very very interesting and detailed [here's a link for it](https://ia801308.us.archive.org/5/items/DesignOfOS2/Design%20of%20OS2.pdf).  Its very fitting for our current conversation with wonderful history on why and how everything happened.

Interesting sections:
            Ch 2.3.5 Address Translation - description of cache registers
            Ch 10.3 276 DOS compatibility, 10.3.1 Mode Switching
            Ch 10.3.3 Tiled Memory and 10.9 tiled descriptors
            Ch 10.4 386 DOS compatibility
            Ch 10.6


>  I'm tempted to just cap the return value at 32M

> Interesting given that this is an SX with only 24 address lines and thus max 16M. 

It seems arbitrary to limit to 32M, perhaps a 16M limit would be more fitting until its found that more memory is actually useful or needed for something (including ram disks).

>  that the INT15 block moves we use for 286 systems, need to move to the fartext segment, like you alluded to once before.

Can you remind me again why we need fartext for INT 15? I have so many things going on right now I'm losing track.
Was this something to do with the A20 gate being turned off after calling it, thus a need to keep the routine itself out of HMA?

>  one of the things fixed today was the assumption in setup.S that the HMA exists if the size returned from the xms-size bios call is > 64k

Also again sorry to be obtuse, but why did this need fixing? Which systems report XMS memory == 0 but still have HMA available? I am also assuming that almost all systems (except Compaq) report XMS memory >= 1024 meaning 1M+ memory available, right?

What do you think of the BIOS E820 idea, have you already written code to use that?

Mellvik:
> > I'm tempted to just cap the return value at 32M
> 
> > Interesting given that this is an SX with only 24 address lines and thus max 16M.
> 
> It seems arbitrary to limit to 32M, perhaps a 16M limit would be more fitting until its found that more memory is actually useful or needed for something (including ram disks).

The idea was to retain ample space for a RAM disk, but TLVC doesn't have that yet. Given that, yes, 16M is just as good.

> 
> > that the INT15 block moves we use for 286 systems, need to move to the fartext segment, like you alluded to once before.
> 
> Can you remind me again why we need fartext for INT 15? I have so many things going on right now I'm losing track. Was this something to do with the A20 gate being turned off after calling it, thus a need to keep the routine itself out of HMA?

That's it. HMA and XMS are mutually exclusive on 286 systems because of the A20 problem. Moving those routines to the fartext segment would solve it. Having the hardware available, I'm going to test that.

> 
> > one of the things fixed today was the assumption in setup.S that the HMA exists if the size returned from the xms-size bios call is > 64k
> 
> Also again sorry to be obtuse, but why did this need fixing? Which systems report XMS memory == 0 but still have HMA available? I am also assuming that almost all systems (except Compaq) report XMS memory >= 1024 meaning 1M+ memory available, right?

It's the other way around. As I mentioned a couple of days ago, some systems, including the Compaqs, when equipped with only 1M memory, report 256k or more XMS available,  all mapped to the below-16M region,  nothing at 1M. The current setup.S will hang such systems because it doesn't test for the actual presence of memory in the HMA area.

> 
> What do you think of the BIOS E820 idea, have you already written code to use that?

Honestly, I've taken a pass on that one this far. I don't yet know what it is. I'll be back on that.

Thanks for the additional details (pointers inside) about the OS/2 book. I look forward to it.

rafael2k:
One question about unreal mode: could we use it in an "unreal far" pointer from a C code and have the compiler to do the magic to generate the memory accesses? Of course then we'd need kernel interfaces for orchestrating the access (or not?).

Mellvik:
> > Wow, not sure about how to handle this?!
> 
> @ghaerr - I find this more amusing than anything else. I have ideas about how they do this, but it doesn't really concern us. We just stop at the 'physical' limit and leave the rest. No big deal really, just important to know - and 'take down' so to speak.
> 
> btw, and on a different (related) note, one of the things fixed today was the assumption in setup.S that the HMA exists if the size returned from the xms-size bios call is > 64k. The plan is to put together 1M portables, III and 386, for testing in the next few days.

I take this back - while scratching my head. I accidentally forgot to enable the 16M cap in my xms_sizer() which turns out to happily read and write memory above 16M on the SX!! IOW the before mentioned (from the ALI spec sheet) 'memory extender' is somehow directly connected to the (usually unavailable) processor address lines above A23, making (presumably) the entire 32 bit address space available. Apparently that's the kind of magic you can pull when designing a SoC. IOW - the xms-size returned from the BIOS call is reliable even in this case.

Never a dull moment.

ghaerr:
@rafael2k:

> could we use it in an "unreal far" pointer from a C code and have the compiler to do the magic to generate the memory accesses?

I can think of a way to do this in a fairly straightforward manner: just use a standard 32-bit x86 C compiler, and modify it to produce the address-size prefix (0x67) before each 32-bit pointer access. However, the compiler will also have to insert operand-size prefixes (0x66) before most other instructions that reference 32-bit registers. (These are required in real mode. In protected mode, a bit in the selector's GDT determines the default operand/address size and these prefixes effectively flip that bit).

So there's going to be a helluva lot of prefixes being emitted. The ELKS native C86 compiler could probably be modified to support this without too much trouble, but those object files aren't compatible with ia16-elf-gcc's object files. Modifying GCC to support this kind of thing would be a nightmare, IMO.

As far as kernel interfaces for all that - no system calls would be allowed for any 32-bit pointers without effectively rewriting the entire system, so this kind of thing would be limited to internal application programs memory management. As @Mellvik has suggested though, just using unsigned longs (aka linear32 ramdesc_t pointers internal to the kernel) and an ASM wrapper for accessing XMS memory would work without all the extra complexity. But then of course the application program would be competing with kernel XMS so they're need to be new system calls to support that - perhaps xmemalloc/xmemfree.

rafael2k:
This is very interesting, thanks @ghaerr and @Mellvik.
So in case of an user which has just one memory hog process (lets say, C86 compiler, or arachne Web browser), the XMS memory could be disabled in the kernel, and the process can be free to check size and allocate how much of the XMS memory space the process wants?

Mellvik:
Interesting thoughts, @rafael2k.

It's not quite that easy though. Your (huge) app would need to run alongside the OS, which is clocking in at least 100 times per second, it would need OS services (disk IO, terminal/console IO, maybe networking), no BIOS access and short time slices in order to not block interrupts for too long while in unreal mode.

Again possibly doable, but really messy. I believe a more viable approach would be to get TLVC or ELKS to run in V86 mode. Not that I have any practical experience with it, but it does eliminate some of the limitations of 'real' real mode while maintaining (much of) real mode compatibility. My impression is that those loosened constraints apply to memory access in particular.

I'm sure @ghaerr has more insight to add to this.

rafael2k:
Thanks for the explanation @ghaerr - and I can say that it is exactly what is needed IMHO for userland - a simple way to get the to get the XMS address, length, and a way to copy data to and from the XMS location (can be in assembly, no problem). Bobcat (lynx fork) is not well designed for this, but microweb and arachne browsers, for example, can definitely be adapted to benefit from unreal mode. I'll open a new issue about this soon.

rafael2k:
Thank you both for bringing unreal mode to life with ELKS and TLVC!

rafael2k:
> It seems I 'overestimated' what you're actually requesting @rafael2k . My comment was made under the assumption that you wanted to run code in extended memory. I now understand that that is not the case, which makes the picture viable, even tempting.

What I'm thinking are two possible use cases: compile large C sources (C86 memory allocations will need to be modified to cope with memory swapping to XMS) and browsing the web (for leaving in base memory for core structures, while html document tree structures and media in XMS - this is implemented for Arachne's DOS version and MicroWeb browser already I think, need to check the details).

PR #163, opened 2025-04-21
[kernel] unreal mode enhancements and fixes

Mellvik:
This PR has a number of fixes and enhancements to TLVC's XMS support - many discussed in detail and at length in #161 (discussion). This is a 'development (work in progress) release.

- Autodetection of the XMS 'structure' has been added which accommodates the following scenarios:
    - XMS does not start at physical memory 1M, which among other things means that XMS present does not automatically imply HMA support.
    - XMS memory may not be contiguous, which means that the XMS size as returned from the BIOS call may not be reliable. TLVC will search for the start and end of the lowest (by start  address) block.
    - Available and thus useable XMS size is capped  at 32M - 1M base memory.
    - In order to implement the 'xms_sizer', `linear32_peekw` and `linear32_pokew` functions have been added.
    - `linear32_peekw()` reveals a possible side effect of using unreal mode: The ES register must be saved even though it isn't used.
- In order to keep the code as small and efficient as possible, `kbytes` is being used as a metric in much of the new code, avoiding longs whenever practical.
- The xms memory allocator has been enhanced to catch overallocation. Still no `free`, and there is no need for one at this time.
- A very simple `dump_gdt()` function has been added for debugging only, which - if enabled - will dump the contents of the processor GDT table as left by the BIOS. 

There is quite a bit of cleanup needed, including moving more/all of the xms-related code away from `init/main.c` (to `xms.c`). Also needed is a cleanup related to CONFIG options. Finally, if it turns out to be possible to get HMA kernel and XMS buffers to coexist on a 286, that would simplify things even more. Those are coming in a later commit. 

This code has been tested on QEMU, the ALI386SX and the Compaq 386DX.


tlvc/arch/i86/lib/unreal.S
```diff
@@ -127,38 +129,75 @@ unreal:
...
 # word_t linear32_peekw(void *src_off, addr_t src_seg)
+# void linear32_pokew(void *src_off, addr_t src_seg, unsigned val)
 # WARNING: Requires 32-bit CPU, with unreal mode and A20 gate enabled!
 #          Trashes EBX and ESI without saving.
 #
 	.global	linear32_peekw
 linear32_peekw:
-	call   setgpf
-	mov    %si,%dx
+	push	%es		// Required: ES is somehow invisibly used, must be saved
```
ghaerr, line 149: > linear32_peekw(): ... The ES register must be saved even though it isn't used.

That is very strange! Just for kicks, have you tried changing the push/pop ES to push/pop DS instead? That would then save the spot on the stack and not require rewriting the 4(SI) and 6(SI) offsets below to 2(SI) and 4(SI) when no push/pop is performed, but not touch ES, just to see what happens.

Another idea to try, would be adding a push ES pop ES just before the modified PUSH DS ... POP DS around the routine. That would test whether for some very strange reason ES has to be rewritten in order to work in unreal mode (even though of course it already has been, twice, in the unreal mode setup routine).

What is the code that's calling these two peek/poke functions? Do they use any far pointers, or just ramdesc_t? It is possible that ES or DS is being reset by code near the calling function if far pointers have been manipulated. 

Yet another idea is to CLI/STI around the entire routine from the very start, and not push/pop ES.

When the routine does not save ES, what does peekw return - always the same value, bad random values, or system crash?

I ask all these questions because, sure as heck, when we do something that we don't understand, it always seems to come back and bite!

Mellvik: Wow, that's indeed a shipload of questions, @ghaerr. Exactly what I needed. There is no mystery, just a stupid mistake. While moving the push around, I forgot to adjust the parameter offsets accordingly. Thanks!

Mellvik:
Next step, @ghaerr - about the 286 and moving XMS code to fartext, this is foreign territory:

1) How to make an assembly function fartext (`block_move()`)?
2) Does the linker enable functions to make near calls to other functions in the same fartext segment?

ghaerr:
> How to make an assembly function fartext (block_move())?

In general, it's not a big deal - the code section will need to be designated to be linked into the .fartext.far section using a .section directive, and then the function needs to be defined FARPROC in C. When called, it will execute normally except that its parameter offsets will have to be adjusted for the extra CS pushed and return with RETF. I'll take a look at the specifics later today.

> Does the linker enable functions to make near calls to other functions in the same fartext segment?

The GCC linker doesn't know anything about near/far segments, so no. One has to be extremely careful with which functions are called from where, as the called function's RET or RETF is constant and needs to be called the way it was coded. I think the answer to what your asking is: make near calls to near functions, and far calls to far functions, that'll work. And calling far from near will work. But calling near from far requires a thunk, which is currently only emitted by the C compiler. I'd stay away from that for the time being, if possible.

The good news is that block_move is a BIOS call (i.e. uses INT 15) and doesn't require calling another function.

I'l try to get something written up so that you can test on real hardware.

Mellvik:
Thanks, appreciated!

Mellvik:
@ghaerr,

> Thanks for bringing the A20 issue to my attention, I had forgotten about the memory test involved. This will not work and needs to be redone because there may be no memory at 1M and still XMS available.

Now that I've looked at the code, I see no reliable way to verify that A20 has been set - since we don't know for sure where the memory is located. And - given all the hoops created by the possibility that XMS doesn't start at 1M, I'm questioning whether the benefit is worth the effort. Whether I'm putting too much into a seemingly esoteric case of a really old clunker with very little RAM. Considering all the code required to handle this rare scenario, maybe  a simple  warning is good enough. When - at boot time - xms size > 0 and A20 verify fails, just issue a warning, just disable HMA and enable INT15 xms, which is sure to work. The user can take it from there via bootopts. 
What do you think?

ghaerr:
Hello @Mellvik, thanks for your comments. Indeed this is a handful keeping everything in mind, given all the differences in some systems we're studying. I want to take it a little slower by dissecting it further.

>  I see no reliable way to verify that A20 has been set - since we don't know for sure where the memory is located.

Lets start with this one. I can now see from your points that determining A20 being set is tricky, but I think it can be reliably done by enhancing verify_a20 a bit more. However, with the help of a20 verification and looking at low core 0:0, I think this can be determined:

With A20 supposedly "enabled", read FFFF:0010, negate it, then write it again. If 0000:0000 changes, then A20 is not enabled, game over for HMA or XMS. However, if FFFF:0010 changes to the negated value, then A20 is working and memory is present at 1M; HMA is OK and XMS starts at 1M. The current verify_a20 routine only checks wrap but doesn't check whether the write works to 1M. 

What do you think?

In a sense, an updated verify_a20 routine can check for both HMA available as well as XMS memory present at 1M. The xms_bytes can be used for the size of XMS memory. This is exactly what we want (for now), since the xms.c doesn't presently handle XMS not starting at 1M (XMS_START_ADDR). Right?

> And I believe the litmus test for A20 should be just that, XMS size > 0.

I don't follow: since there are a couple different (incompatible) ways of enabling/disabling the A20 gate, how is it that a simple XMS > 0 test figures out which method to use for enable/disable of A20?

> When - at boot time - xms size > 0 and A20 verify fails, just issue a warning, just disable HMA and enable INT15 xms, which is sure to work. The user can take it from there via bootopts.
What do you think?

Have you ever seen xms size > 0 and A20 verify fail? That sounds like the case where XMS memory is not present at 1M. If that's the case, the enabling INT 15 still won't work, since XMS_START_ADDR is still at 1M.

In setup.S, HMA is already currently disabled if xms size = 0 *or* A20 verify fails.

> it may be a good time to reasess the entire idea: HMA kernel may not be worth it 

I still think HMA kernel is a great enhancement: there's no other way to increase the amount of memory available to user applications, especially when not running with lots of buffers (e.g., really tight memory situations but not much I/O needed - like running Doom).

Sorry for so many questions - it's hard to keep all this detail in my head with all that's going on!

ghaerr:
@Mellvik: It's pretty much game over for FARPROC INT 15 and any associated A20 handling, because the first hardware interrupt will immediately crash the system when HMA kernel and A20 disabled. See https://github.com/ghaerr/elks/pull/2320#issuecomment-2828191980 for the full analysis. 

I'm going to revert the FARPROC block_move and A20 changes, and all this means it doesn't matter how fast the first (and only) call to enable_a20_gate is: IMO we should continue to use both BIOS and keyboard mechanisms, in order to be safe.

> Which brings me to a point that IS still relevant: Which A20 mechanism to use. 

I'm in agreement with your analysis, and thanks for the pointer about the FAST A20 switch. But at this point, given the total failure of being able to reliably switch A20 when using BIOS INT 15/1F block moves, there seems to be no point in trying to determine the *best* A20 enable method, since the routine will be called exactly once.

This has been a hell of a journey... and there's some small details to handle if we care about XMS starting > 1M. But I think I'm about done with this too, since at least ELKS and I presume TLVC will boot and run on all known systems including Compaq and PC-98, and just auto-disable XMS when 286 CPU and HMA kernel since INT 15/1F can't be made to reliability work. As you've pointed out, this whole issue now only affects 286 CPUs w/HMA. With your full protected mode mechanism for setting the segment cache limit registers, the PC-98 and Compaq now join the club for working XMS using unreal mode.

Regarding LOADALL etc from previous conversations: not worth it, since 80286 can't reset itself out from protected mode back to real mode without external hardware - so we'd be back into more ancient IBM specifics for how to reliably do that. I'm not going to worry about it.

While it is tempting to remove INT 15/1F entirely, I'm going to leave it in, so that 286 users can still enjoy XMS, just not HMA kernel. The kernel stays a little bigger, but Compaq/286 users get to run without rebuilding.

Phew, glad this is over! :)

Thank you!

Mellvik:
Wow, that explains the weird behaviour I've been experiencing this afternoon. Thanks for the  immediate update, that certainly saves a lot of debugging time. I'll work some mor3 on it in the morning to make sure I understand (and can reliably place) the symptoms.

An intense marathon indeed. I look forward to dive into your analysis - will be back with comments and questions tomorrow.

Thanks

ghaerr:
> You agree?

Well - no. Your modification will still work, but it's slower. In the original, a SHR instruction shifts all bits right, with bit 0 going into CF (carry flag). That bit is recovered with the RCL later for the final byte (if any).

In your version, CF is cleared via CLC, and then all bits are shifted right, with bit 0 going into CF (all same as SHR), *except* that with RCR the CF flag is shifted into the high bit of ECX, which we definitely don't want. But it works since CLC cleared CF beforehand.

So, no, the change isn't really what is wanted, but does work. BTW, this same mechanism is used in linear32_fmemcpyb, just above this routine, as well as in the kernel fmemsetb and fmemcpyb routines in memory.S.

In summary, both SHR and RCR shift bit 0 into CF. But RCR shifts CF into the high bit of the result, whereas SHR shifts 0 in. We definitely want the latter, since the whole trick is converting bytes to words first with a right shift (not rotate), then possibly a single byte operation afterwards.

Mellvik:
> Well - no. Your modification will still work, but it's slower. In the original, a SHR instruction shifts all bits right, with bit 0 going into CF (carry flag). That bit is recovered with the RCL later for the final byte (if any).

No, the docs specifically points out that the rightmost bit is lost in the shift. That's the bug.

ghaerr:
> the docs specifically points out that the rightmost bit is lost in the shift.

What docs are you reading? The x86 instruction set reference [says the opposite](https://c9x.me/x86/html/file_module_x86_id_285.html). All right shifts shift the rightmost bit into CF:
SHR

Mellvik:
I think that applies to left shift only. The Robert Hummel book says about SHR: "... the bit that occupied the low-order position of the destination is lost. If it is necessary to preserve this bit, use the RCR instruction."

ghaerr:
> I think that applies to left shift only. 

Both left and right shift shift the bit you're calling "lost" into CF.

> the bit that occupied the low-order position of the destination is lost. 

Melvik! The whole point here is to perform a divide by two, to convert words to bytes! We don't want that bit rotated back into the result. The lost bit is saved in CF and for the later. 

Sorry, but is sounds like you are not understanding how SHR (or SHL) works, nor the subtleness of this routine (and all the others that depend on it in the working kernel). Get a book that shows a visual of the register bits and CF (for any processor, they're all the same(), so you can see what I'm talking about.

And to repeat, the only reason your version works is that you CLC first, thus guaranteeing that the high bit is 0, while *still* rotating the lower bit OUT of ECX in CF! Do you see it?!

You are misinterpreting Hummel's statement about whether or not one wants to keep the original bits (rotated), or shift in a 0 or 1 (the difference between SHR shift right and ASR arithmetic shift right).

Every processor, from the 8008, 8088, 8086, 386, ARM, etc works in exactly the way I'm describing, with regards to the differences between rotate, shift and arithmetic shift - they all involve CF. Upgrading 16->32, 32->64 bit ADDs, SUBs, etc all depend on it.

ghaerr:
>  how SHR (or SHL) works, 

[Here are some visuals](https://home.adelphi.edu/~siegfried/cs174/174l7.pdf) to show how each of the R/L rotate and shift instructions work, to help you. 

> nor the subtleness of this routine 

In particular , the above link has following slide, which just happens to use exactly the same code as the routine we're talking about to restore the previous value in the shifted register:
RCL

Mellvik:
You're right, @ghaerr. It took a full 30 seconds with DOS DEBUG to verify that. Robert Hummel's text is wrong, for both SHL and SHR - at least the way I read it. However, the algorithms presented with each instruction are correct. 

Thanks for clarifying that!

Mellvik:
The most recent commit includes the following fixes/enhancements:
- Adds `xms=` to the boot options using the same values as ELKS, `on` and `unreal`.
- CONFIG_FS_XMS_INT15 has been eliminated, Int15 block move supoort is is now always included with CONFIG_FS_XMS_BUFFERS (which is soon to be renamed)
- Adds tentative support for INT15 block moves and HMA kernel concurrently (intended for 286 systems). In its current form, it does not work, as discussed previously in this thread and in the corresponding thread on ELKS, and is subject to removal. See also the next comment in this thread about this issue.
- Setup (`setup.S`) now checks for the physical presence of HMA memory before relocating the kernel there.
- The 'bug fix' in `linear32_fmemset` discussed above is included and will be removed.
- The validation/configuration code in `xms_init()` (`xms.c`) has been completely reworked and now includes the code formerly located in `init/main.c`, including the (commented out) `find_xms_start()` and `find_xms_end()` routines.
- `unreal.S` has new `linear32-peekw` and `...pokew` routines, well tested, but not currently used (and thusly commented out).

Mellvik:
**HMA kernel + INT15 XMS buffers now working on AT/286 machine**

Admittedly I didn't believe my own eyes, yet here it is. This is from the COMPAQ portable III, 286 w/ 6784K memory, running off of a floppy, with serial console and plenty of interrupt load. Notice the initial 'xms:' line and the kernel ds later.
I figured it had to fail eventually, but it has not this far.

```
ttyS0 at 0x3f8, irq 4 is a 16450
ttyS1 at 0x2f8, irq 3 is a 16750
xms: 6080k available, using int 15/1F, FAR bufheads
     256 xms buffers (base @ 0x110000), 8K L1-cache, 20 req hdrs
eth: ne0 at 0x300, irq 12 not found
eth: wd0 at 0x280, irq 2, ram 0xcc00 not found
hd0: AT/IDE controller at 0x1f0
hda: IDE CHS: 823/4/38 (from /bootopts)
df: Floppy controller, FDC 8272A @ irq 6, DMA 2
df0: 1.44M (type 4) [CMOS]
Floppy cache 5k, available 5k
Partitions: hda:(0,125096)  hda1:(1,6374)  hda2:(6375,1275)  hda3:(7650,41565)  hda4:(492
15,34000) 
PC/AT class, cpu 6, syscaps 0xff, 640K base ram, 18 tasks, 64 files, 96 inodes
TLVC 0.6.0 (61520 text, 17120 ftext, 10000 data, 7744 bss, 47790 heap)
Kernel text 0xffff, ftext 0x1d0, init 0x392, data 0x5fe, top 0xa000, 552K free
MINIX: inodes 256 imap 1 zmap 1, first data 12
MINIX-fs: mounting unchecked file system 0x200, running fsck is recommended.
VFS: Mounted root 0x0200 (minix filesystem).
Running /etc/rc.sys script
Mon Apr 28 10:27:15 2025


[tlvc15] TLVC 0.6.0

login: root
tlvc15# 
```

Just as I was preparing for removal of the experimental __far code, I decided to give it a final shot: Instead of just disabling interrupts before and after the BIOS block move call, I masked all interrupts on both 8259 controllers. Before that I shortened the time spent in `enable_a20_gate()` by having it return immediately if successful instead of running the validate code. I don't know if this makes a difference, probably not. That will become apparent as I start to remove the 'scaffolding' in order to see what's really needed to keep this going.

Possibly most surprising is that it wasn't enough to mask all interrupts on the first controller. I would have expected that since IRQs from the second is gated through the first.

Anyway, while it still feels like a dirty hack, it works reliably this far, and will get more of a beating with networking active and the whole shebang while the code is cleaned up before being committed.

Mellvik:
More testing and some cleanups delivered the last commit. INT15 XMS buffers with HMA kernel works flawlessly on the 286Compaq and the 386SX/ALI system. Only one anomaly seen, IRQ7 doesn't work when running on INT15 xms. Moving the `ttyS2` line from IRQ7 to IRQ10 and it works fine. Network is running fine on IRQ15 regardless. The IRQ7 issue may be system dependent.

Mellvik:
> > Setup (setup.S) now checks for the physical presence of HMA memory before relocating the kernel there.
> 
> Did you find this was necessary for the Compaq in certain memory configurations? Or was this just put in for safety? I agree it is probably needed but didn't know if the code has in fact been tested with a system that reports XMS > 64 but doesn't have XMS memory starting at 1M.

You're right, this code has not been tested (yet). I have yet to pull apart the compaq 386 and remove memory in order to test it. It does not apply to the 286 which has a different memory configuration (basic memory is 640k, not 1M, and expanded mem presumably starts at 1M).

> 
> > I masked all interrupts on both 8259 controllers
> 
> I must admit I never really thought of literally turning off all external interrupts at the PIC so that the block_move isn't otherwise interrupted by CLI/STI. I assume you've also got the NMI interrupt off as well (although that might not actually be needed since it won't vector to FFFF:XXXX in HMA).

I did, and I haven't tried without it (that is, without NMI enabled). I'll do that while at it.

> 
> > Moving the ttyS2 line from IRQ7 to IRQ10 and it works fine
> 
> My take is that since the original slow block move already executed with interrupts off, having 20-100 more instructions to externally disable the PIC isn't likely to affect timing that much, except for the big problem with edge-triggered interrupts, for which this solution likely won't work at all. So it would probably pay to research which device hardware is issuing edge-triggered, vs level-triggered interrupts: the level triggered hardware would be holding its interrupt request line active the entire period and the request finally granted when the PIC mask is re-enabled, while the edge-triggered could possibly be lost forever, it would seem. I would guess both the serial and network cards are using level-triggered interrupts, to support this theory of operation (except this doesn't explain the ttyS2 line issue - is that a separate serial card from the first two ports?)

Agreed, I don't think we have any edge triggered interrupts or I would have seen the effects by now. The serial IRQ is a mystery. This is the ALI system, there are 4 serial lines, all soft configurable. Why the h... would it work with one IRQ and not the other? And fail on IRQ7 only if block_move is used? Doesn't make much sense. No theory about that yet.

> 
> > while it still feels like a dirty hack, it works reliably this far,
> 
> Yes, it is possibly a dirty hack, but I would say the big issue is figuring out why IRQ 7 isn't working: without knowing that, the result could be deemed unreliable, since hardware had to be changed in order to stay working, for reasons not known. 

It could be, but like I said, it's even more likely to be somehow system related. Like, IRQ5 is available and unused on the system, but has never worked on serial lines. That is different from working in one setting and not in a different one, but for all I know - not having experimented with IRQ5 much at all - that may be the case here too. I may put a dual serial card configured for IRQ 7 into one of the compaqs to figure it out.

> This could easily creep into other systems or devices. As a result, I would offer the recommendation that perhaps (yet another) XMS mode that would differentiate between those systems wanting a working INT 15 (and no HMA) without having to reconfigure their hardware card IRQ settings, and the disabled interrupts INT 15. (XMS_INT15DI?)

I've been thinking about that, but now that the LOADALL options seems to be coming closer by the minute, the whole exercise may turn out to be just that. A proof that it's actually possible.

ghaerr:
> A proof that it's actually possible.

I think you've proved that: we both realized that a FARPROC block_move was necessary for to get the code out of HMA with A20 off. I then discarded the idea of using the FARPROC block_move because of the possibility of the PIC holding previously requested interrupts until interrupts were reenabled in the CPU, and then you disabled interrupts completely at the hardware level. But in the process, IMO, you've uncovered another rabbit hole that should probably be investigated.

> I may put a dual serial card configured for IRQ 7 into one of the compaqs to figure it out.

That would be worth doing. 

Since a good part of the whole point of this process is to educate ourselves about how things really work, I would say that getting to the bottom of why IRQ 7 doesn't work w/PIC disabled, and/or IRQ 5 never worked would be worth the effort. Granted there might be something very strange about your particular AMI system, so playing with the same cards on other hardware might also be worthwhile.

> but now that the LOADALL options seems to be coming closer by the minute

It looks like LOADALL is fully working. I've reviewed the code and am probably going to help by posting a PR with it all in shortly. Far better - but remember that 286 LOADALL is an invalid instruction on 386, and that 386 LOADALL was removed from the mask on 486+, so I'll be slightly recoding xms.c for 286 only, otherwise using existing INT 15, which I'll probably leave in for a while for testing, etc.

Which brings up another small issue - I'm thinking that ELKS should possibly process INT 6 interrupts - invalid opcode. It might be nice to know when the system has executed bad code, just like our DIV0 handler.

Mellvik:
Seems we're in the fast ale these days, that's inspiring!

> > A proof that it's actually possible.
> 
> I think you've proved that: we both realized that a FARPROC block_move was necessary for to get the code out of HMA with A20 off. I then discarded the idea of using the FARPROC block_move because of the possibility of the PIC holding previously requested interrupts until interrupts were reenabled in the CPU, and then you disabled interrupts completely at the hardware level. But in the process, IMO, you've uncovered another rabbit hole that should probably be investigated.
> 
> > I may put a dual serial card configured for IRQ 7 into one of the compaqs to figure it out.
> 
> That would be worth doing.

I took the easy way out, remembering that I've seen that serial line not working from time to time before:
I commented out the PIC disable code and the hma=kernel statement in bootopts and rebooted with the serial line set til IRQ7 and xms=int15 with plenty buffers active. ttyS2 on IRQ7 doesn't work. So it has to do with the use of this particular BIOS call, it has been there forever and it's not related to the PIC hack. I may still test a serial card on a Compaq though.

More interesting and equally useful: Removing the NMI disable code prevents the reboot command from working. Without looking, that vaguely makes sense. So I'll keep it in.

> 
> > but now that the LOADALL options seems to be coming closer by the minute
> 
> It looks like LOADALL is fully working. I've reviewed the code and am probably going to help by posting a PR with it all in shortly. Far better - but remember that 286 LOADALL is an invalid instruction on 386, and that 386 LOADALL was removed from the mask on 486+, so I'll be slightly recoding xms.c for 286 only, otherwise using existing INT 15, which I'll probably leave in for a while for testing, etc.

Wouldn't testing for cpu type take care of that completely?

> 
> Which brings up another small issue - I'm thinking that ELKS should possibly process INT 6 interrupts - invalid opcode. It might be nice to know when the system has executed bad code, just like our DIV0 handler.

This is probably a good idea regardless.

ghaerr:
> ttyS2 on IRQ7 doesn't work. So it has to do with the use of this particular BIOS call, it has been there forever 

Running BIOS INT 15 call (unmodified) causes ttyS2 to fail, but otherwise does not fail? Strange indeed. 

> Removing the NMI disable code prevents the reboot command from working. Without looking, that vaguely makes sense. So I'll keep it in.

This is on the AMI, not Compaq, right? I'd never heard of the NMI I/O port until last week: do you think that I/O port may not be implemented on all machines? What exactly is the NMI line hooked up to on the ATI hardware? We know it's doing something special on Compaq but not sure about standard IBM PC.

Also, the NMI should normally be enabled at all times, except for during the present dirty hack, right? Why would reboot have any effect if you're in the middle of an INT 15 memcpy? The reboot command executes other INT 15 code, IIRC, but NMI should be enabled at the time that's called. Better double check that.

> Wouldn't testing for cpu type take care of that completely?

Yes - except we want to be able to allow INT 15 for testing on 386, or possibly on 286, for testing. This gives us something to compare against when problems arise, rather than having no option.

Mellvik:
> > ttyS2 on IRQ7 doesn't work. So it has to do with the use of this particular BIOS call, it has been there forever
> 
> Running BIOS INT 15 call (unmodified) causes ttyS2 to fail, but otherwise does not fail? Strange indeed.

It is. I just changed the IRQ on that line to IRQ5, then the ALI system hangs during boot. What I've noticed before is that there are combinations that don't work on this system, presumably a mismatch between the BIOS and the hardware. This is an industrial SoC, just about everything is inside the chip. What's outside is line drivers/buffers, some IO decoding, DiskOnChip support and BIOS ROMs (plus RAM of course). And a via chip if I remember correctly. Floppy and hard disk controllers are inside the SoC.

> 
> > Removing the NMI disable code prevents the reboot command from working. Without looking, that vaguely makes sense. So I'll keep it in.
> 
> This is on the AMI, not Compaq, right? 

That's right. I haven't tested the reboot issue on the COMPAQ yet.

> I'd never heard of the NMI I/O port until last week: do you think that I/O port may not be implemented on all machines? What exactly is the NMI line hooked up to on the ATI hardware? We know it's doing something special on Compaq but not sure about standard IBM PC.

Yeah, this needs some more investigating. And me too. While familiar with the NMI line and its purpose, a port to disable it was news to me. I haven't found any references to it on the net yet, and I'm sure it wasn't on pre-AT machines.

> 
> Also, the NMI should normally be enabled at all times, except for during the present dirty hack, right? Why would reboot have any effect if you're in the middle of an INT 15 memcpy? The reboot command executes other INT 15 code, IIRC, but NMI should be enabled at the time that's called. Better double check that.

Actually, reboot is looping, interruptible. So it's most likely looping in try_unmount(). Doesn't make much sense. To be investigated.

> 
> > Wouldn't testing for cpu type take care of that completely?

> 
> Yes - except we want to be able to allow INT 15 for testing on 386, or possibly on 286, for testing. This gives us something to compare against when problems arise, rather than having no option.

That sounds similar to the setup we have today (disabled, macro AUTODISABLE), which will disable INT15 XMS if HMA kernel is active. So, if arch_cpu!=6, no LOADALL regardless, but xms=int15 may still be honoured unless hma=kernel, in which case unreal is the choice (or if my int disabled hack is available, use that). And yes, we might need another xms= option to turn off LOADALL even if running a 286. wow, another variant to consider!

ghaerr:
> Removing the NMI disable code prevents the reboot command from working. 

The reboot code looks like this, which involves a jump to FFFF:0000:
```
    asm("mov $0x40,%ax\n\t"
        "mov %ax,%ds\n\t"
        "movw $0x1234,0x72\n\t"
        "ljmp $0xFFFF,$0\n\t"
    );
```
We're running in HMA and it could be possible that's not real ROM but RAM? IIRC your kernel is loading at FFFF:0000, not FFFF:0010, so that could explain things. I think I fixed that in my setup.S but have to check. My crt0.S code doesn't have a `jmp _start` as it's not needed. I seem to remember you had something like that in there (wouldn't matter either way, since executing 0's will still get you to \_start).

Not sure what any of this has to do with NMI though.

ghaerr:
NMI articles:

https://retrocomputing.stackexchange.com/questions/11848/did-the-ibm-pc-use-the-8088s-nmi-line

https://wiki.osdev.org/CMOS

https://wiki.osdev.org/Non_Maskable_Interrupt

Interesting reading, especially that the NMI I/O port is used to control lots of things. Not sure to what degree any of this is standard, but in general, I would say fiddling with that port is potentially risky and could cause system instability on some systems.

Mellvik:
I'm going to investigate the reboot issue more, after reading your links. Like the irq7 issue this may be a fals3 alarm. Reboot works fine on the compaq w/o the nmi blocking.

Irq issue: the compaq works fine with 3 serial lines, 2 of them at irq 7 and 10. with kernel high and interrupt blocking active. Which closes that page.

Mellvik:
Thanks again for the links, @ghaerr. Turns out they were familiar - and while educating, there is a lot of speculation out there, on many issue including the NMI port.
I believe - from reading BIOSes and other sources and from own experience, that using the NMI blocker without going through all kinds of hoops related to the CMOS chip is safe. From the schematics I've looked at, it seems unlikely that the clock chip is even selected unless there is data on the D0-D5 lines. This is an 'educated opinion' and there may be settings (systems) that require 'tricks' to make it work. 
Still, after running more tests on the reboot issue discussed yesterday and found it a transient and unrelated to the issue at hand, I'm going to leave the NMI disable/enable code commented out - inside the `#ifdef CONFIG_FS_XMS_INT15DI` framing for now.

PR #164, opened 2025-05-02
[kernel] fd cache was ignored when using xms buffers

Mellvik:
Running a physical system off of floppy disk is really important some times.

While testing/debugging  the `loadall` implementation on the 286, which incidentally runs off of floppy (because the HD is busy with other OSes), it became clear that the floppy bounce buffer was used all the time, not only when it should be. A minimal restructuring of the `setup_DMA` routine was required to get it right.

Running root on floppy introduces delays that influence many parts of the system, including how `kctp` and the network deamons interact and run, which in this case revealed  a couple of problems - to be addressed separately.

### DIFF
--- a/tlvc/arch/i86/drivers/block/directfd.c
+++ b/tlvc/arch/i86/drivers/block/directfd.c
@@ -678,6 +678,7 @@ static void setup_DMA(void)
     struct request *req = CURRENT;
 
 #pragma GCC diagnostic ignored "-Wshift-count-overflow"
+    use_bounce = 0;
     physaddr = (req->rq_seg << 4) + (unsigned int)req->rq_buffer;
     dma_addr = _MK_LINADDR(req->rq_seg, req->rq_buffer);
     count = nr_sectors<<9;
--- a/tlvc/arch/i86/drivers/block/directfd.c
+++ b/tlvc/arch/i86/drivers/block/directfd.c
@@ -678,31 +678,33 @@ static void setup_DMA(void)
     struct request *req = CURRENT;
 
 #pragma GCC diagnostic ignored "-Wshift-count-overflow"
-    use_bounce = req->rq_seg >> 16;		/* XMS buffers active, always bounce */ 
     physaddr = (req->rq_seg << 4) + (unsigned int)req->rq_buffer;
     dma_addr = _MK_LINADDR(req->rq_seg, req->rq_buffer);
-
     count = nr_sectors<<9;
-    if (use_bounce || (physaddr + (unsigned int)count) < physaddr) { /* 64k phys wrap ? */
-	use_bounce++;
-	dma_addr = _MK_LINADDR(FD_BOUNCESEG, 0);
-	if (raw) {	/* The application buffer spans a 64k boundary, split it into
+
+#if CONFIG_FLOPPY_CACHE
+    if (use_cache) 			/* set if it's a read, not raw, cache OK */
+	dma_addr = _MK_LINADDR(FD_CACHESEG, 0);
+    else
+#endif
+    {
+	/* Need bounce buffer if spanning a 64k boundary and if buffer is in XMS memory */ 
+	if ((req->rq_seg >> 16) || (physaddr + (unsigned int)count) < physaddr) { /* 64k phys wrap ? */
+	    use_bounce++;
+	    dma_addr = _MK_LINADDR(FD_BOUNCESEG, 0);
+	    if (raw) {	/* The application buffer spans a 64k boundary, split it into
 			 * 2 or three parts, using the first k of the sector cache
 			 * as a bounce buffer */
-	    int sec_cnt = (0xffff - physaddr) >> 9;
-	    if (sec_cnt < 1) {	/* the single block with wrap problem */
-		nr_sectors = BLOCK_SIZE >> 9;	
-	    } else {		/* get the sectors before the wrap */
-		nr_sectors = sec_cnt;
-		use_bounce = 0;
-		dma_addr = _MK_LINADDR(req->rq_seg, req->rq_buffer);
+		int sec_cnt = (0xffff - physaddr) >> 9;
+		if (sec_cnt < 1) {	/* the single block with wrap problem */
+		    nr_sectors = BLOCK_SIZE >> 9;	
+		} else {		/* get the sectors before the wrap */
+		    nr_sectors = sec_cnt;
+		    use_bounce = 0;
+		    dma_addr = _MK_LINADDR(req->rq_seg, req->rq_buffer);
+		}
 	    }
 	}
-    } else {
-#if CONFIG_FLOPPY_CACHE
-	if (use_cache) 
-	    dma_addr = _MK_LINADDR(FD_CACHESEG, 0);
-#endif
     }
     if (raw) {	/* ensure raw access doesn't span cylinders */
 	int rest = (floppy->sect<<1) - sector - head*floppy->sect;
@@ -712,7 +714,7 @@ static void setup_DMA(void)
 	    nr_sectors = rest;
 	}
     }
-    count = nr_sectors<<9;
+    count = nr_sectors<<9;	/* recalculate count, nr_sectors may have changed */
     DEBUG("setupDMA:-%x:%x-", req->rq_seg, req->rq_buffer);
 
 #ifdef INCLUDE_FD_FORMATTING


Mellvik:
Worth noting in terms of floppy performance: Because of the nature of things (!), xms-buffers and and floppies are a bad combination. While the fix above obviously helps, there is nothing like multisector writes when using xms (well, I mean multiblock, all writes are 2 sectors). Since the data is coming from XMS everything needs to go through the bounce buffer. If deemed important, the floppy cache could be used, but it's too much work for a rather esoteric issue. 
This is more like a fact of life - just something to be aware of. 

Floppy use will always be faster with buffers in low memory.

Mellvik:
Hi @ghaerr, how interesting you should mention it. When fixing this problem, I got the exact same idea. To the extent that I was about to do some coding to check it out, books open on the desk and the whole shebang - while wondering why no one had done that before.

Then came the partykiller: DMA above 1M is possible but requires word access. IOW an interface with 16bit IO. I don't know of any such interfaces in existence, possibly a few sound cards. I guess at the time when the AT became available, the processors beat the bus/dma speed with PIO, and that was it...

ghaerr:
> DMA above 1M is possible but requires word access. IOW an interface with 16bit IO. 

What exactly are you thinking the issue is, is it the 8237 DMA controller that must access memory 16-bits at a time, must be I/O-addressable 16-bits at a time, or something else? I'm curious, since why would IBM have added an extra page register if that wouldn't work with the system's XMS memory?

Also the [JEMM](https://github.com/Baron-von-Riedesel/Jemm) (add-on DOS EMM/EMS memory manager based on FreeDOS) mentions support for XMS-based DMA buffers; I haven't read the source as to exactly how they're getting the I/O done.

ghaerr:
Thanks for your comments, I haven't been studying any schematics (yet), and am still trying to come to a full understanding of the IBM PC hardware.

> The second 8237 beats this limitation by being wired with skewed (shifted if you like) address lines, no connection to A0. 

Yes, but the FDC is continued to be wired to DMA channel 2, which is the "first" chip, running in slave mode from the new "second" chip, which is wired for word transfers. Thus, only DMA from channels 5-7 are subject to the word transfer issue you're referring to. 

The FDC controller works fine now with 8-bit-only transfers and normal A0 addressing, and would continue to do so with just external page register set to address the memory at 1-15MB, using the same DMA channel #2.

The direct FD floppy driver is currently running in just this configuration, but with the external page register set to 0. Setting the external page register from 1-15 would change the memory address gate to an address in the first 15MB of XMS.

> This makes it capable of transferring 128 bytes max per transfer etc.

My research shows that address line A15 is actually not wired up (discarded), so for the word transfer DMA chip, only 32768 16-bit word transfers are allowed, still resulting in a max 64KB transfer, not 128k.

ghaerr:
Looking a bit further into this, I now think not only will floppy read/writes into DMA be possible (using the direct FD driver only), but the driver itself needs only a few lines changed. The driver is already using the lower 4 bits of the external page register, labeled DMA_TOP / DMA_PAGE_2 (at least in my driver) to set A16-A19 in the lower megabyte. The upper 4 bits will be used to set A20-A23, with nonzero meaning XMS:
```
    dma_addr = LINADDR(req->rq_seg, req->rq_buffer);
...
    outb((unsigned)(dma_addr >> 16), DMA_TOP);
```
The bigger issue is that the buffer I/O system currently specifies an address using segment/offset, but it will need a ramdesc_t type instead of segment number, and/or a separate 4-bit XMS "megabyte bank number". I'll have to think about the most appropriate design, should it make sense to even support XMS floppy buffers. 

Currently, I only really see a couple big benefits: no floppy cache in low memory, and faster floppy performance when XMS is enabled. Your previous work last year showed little benefit for floppy caches in the first place, but getting rid of any cache in low memory is nice for maximizing memory; speeding up floppy transfers w/XMS is the comment which got me thinking about this in the first place. The downside will be using another byte of memory in each buffer struct. However, in ELKS, there is an extra byte available for free, right after b_mapcount, so this wouldn't cost anything.

Of course, moving the entire floppy disk cache to XMS means that more time will be spent doing XMS buffer copies. It sounds like the biggest speed bottleneck is losing the multisector read operations, with the XMS copies not being that time consuming. But at some point the track cache's usefulness itself becomes less important, depending on CPU speed, from your last year's research. Sounds like more thinking is needed on when an optional XMS DMA track cache might be enabled.

Mellvik:
Unfortunately this cannot work. The 16M address space (the expanded page register) is available only to DMA controller 2, which comes with the word only access 'limitation'. The 765 cannot handle that regardless.

As to the channel, like you say it's normally hardwired, but it depends on the controller card and the PC. ATs typically had the floppy controller on the main board and the DMA channel was hardwired. ISA-based floppy controllers may or may not have jumpers for the DMA chan, newer PC with north bridge controllers can assign the DMA in a BIOS setting.

ghaerr:
> This has nothing to do with the wiring that allows the above 1M memory access, which is handled by the page registers. 

> The 16M address space (the expanded page register) is available only to DMA controller 2, which comes with the word only access 'limitation'. 

These two statements seem contradictory, sorry - I'm trying to get to the bottom of this.

What you're saying is that page register ports 81-83h, corresponding to channels 1-3 have only 4 bits wired to A16-A20 with the upper 4 bits unwired, while page register ports 89-8Bh are wired to A16-A23? Which machine schematic are you reading regarding the page register connections? I haven't yet identified the chip responsible for implementing the 8 page registers on a dual 8237 system.

My research is showing that both master and slave page registers are wired to support a 24-bit address space, although on early PCs, the first (slave) controller would not initially have wired A20-23, since there was no XMS memory. But on systems with XMS memory, it may be system-dependent (or not) as to whether the upper 4 bits are connected. If a single chip were responsible for implementing the 8 page registers, then all 8 A16-A23 address lines would be connected to the address bus.

I now understand the previously pointed out issue regarding 16-bit I/O, but that's irrelevant to this discussion because the FDC is always wired to channel 2 on all PC compatibles, or our direct DF driver wouldn't work.

ghaerr:
Working POC implemented in https://github.com/ghaerr/elks/pull/2339.

> be systems out there that expand the page registers for the first dma controller, but somehow I doubt it because it doesn't make sense.

Doesn't make sense? It would seem to me that the first reason to make the first (slave) DMA controller work with all addresses < 16MB would be exactly what we're discussing now - ways to speed up floppy disk I/O without having to force the use of bounce buffers within an operating system.

Mellvik:
Well, I'll be dam... . I admire your tenacity, @ghaerr. And as is so often the case, we're approaching the issue from very different angles. Which isn't so bad although it does create misunderstandings along the way.

So let me rewind a little to put perspective on my angle.

> Working POC implemented in [ghaerr/elks#2339](https://github.com/ghaerr/elks/pull/2339).

Great, I'm looking forward to check it out on real hardware.  Needless to say, I didn't think that was possible.

> Since my whole point in responding was that you were complaining about the lack of performance using floppies and XMS buffers, I will put together a proof of concept if you are interested.

Your take entirely, @ghaerr. I never complained, I pointed out the facts. And honestly, whatever comes out of this, it's proving (or disproving) a point much more than delivering usefulness. Which is fine BTW. There are plenty of such examples in ELKS and TLVC, some still in the code, some discarded after proving them possible. The floppy cylinder and track cache are great examples. Beating the A20 problem related to XMS block transfers and HMA kernel is another. The fun of making things work beats usefulness. And there is always knowledge to be gathered along the way.

>> Unfortunately this cannot work. The 16M address space (the expanded page register) is available only to DMA controller 2, which comes with the word only access 'limitation'. The 765 cannot handle that regardless.

Yes, you were right and I was wrong on this one. A different IBM source (the AT Tech Reference 1985) confirms that unequivocally:

```
DMA controller 1 contains channels 0 through 3. These channels
support 8-bit data transfers between 8-bit I/O adapters and 8- or
16-bit system memory. Each channel can transfer data
throughout the 16M system-address space in 64K blocks.
```

IOW it's very likely that your POC will work on real hardware too. Great!

> The bigger issue is that the buffer I/O system currently specifies an address using segment/offset, but it will need a ramdesc_t type instead of segment number, and/or a separate 4-bit XMS "megabyte bank number". I'll have to think about the most appropriate design, should it make sense to even support XMS floppy buffers.

I may be missing something here: The address provided to the DMA subsystem (controller + page reg) is always the physical address, and the driver - in `setup_DMA` - has this already, whether XMS or main memory.

So now we're here, we know it's possible. Is it useful? 
There are some considerations to be discussed, all mentioned above, and keeping in mind that the driver knows nothing about the buffer system, it does IO of the requested size, to/from the address provided.
1) writes may now go directly from buffer to device even if the buffer is in xms. No bounce.
2) similar for reads: caching is marginally useful for 286 systems, not useful for 386+, so it makes sense to just forget about the cache in this context: XMS buffer = no cache. Reads go to the buffer, and slightly simpler, see next point. 
3) possible simplification in the fact that when memory is xms, allocation may be made so that there is no chance a 64k phys boundary may be crossed, thus simplifying the path through the driver.

It occurs to me that the obvious thing to do is 'nothing', simply adjust the driver to no longer distinguish between xms addresses and regular memory addresses. The cache is gone already (selectable, but always gone on 386+ systems in TLVC), raw access is not affected and if applications at some point should have data in XMS memory and do access to/from it, that would be fine too.

Issue #166, opened 2025-05-04
Floppy boot problem if DEF_OPTSEG changes


Mellvik:
If a kernel built with a DEF_OPTSEG value that is different from the one used when a floppy image was created, that new kernel will not load `/bootopts` at startup.

Let's say the 1.44M floppy image is built with a DEF_OPTSEG setting of 0x90. It boots and loads bootopts fine. Then the DEF_OPTSEG is changed to 0x50 (config.h) and a new kernel I generated (make clean; make). The new kernel is downloaded to or copied to the floppy. The new kernel boots fine, but `setup` does not find `bootopts`.

I've traced this all the way down to display the first two bytes from the buffer given to INT13 to read bootopts into. They are zero. So it seems likely that the code locating bootopts in the file system (minix) is somehow making assumptions tied to DEF_OPTSEG.
 
Confusingly, it always works on QEMU, but on QEMU the new image, not only the new kernel, is booted all the time. When repeating the scenario above, changing only the kernel and not the image, it fails on QEMU too.

@ghaerr, can you possibly shed some light on this?

ghaerr:
> So it seems likely that the code locating bootopts in the file system (minix) is somehow making assumptions tied to DEF_OPTSEG.

I haven't had time to fully go through the full scenario(s) you discussed, but yes, the MINIX boot loader (sector 1 track 0) has DEF_OPTSEG compiled into it (see bootblocks/boot_minix.c). So the MINIX boot loader loads /linux as well as /bootopts.

However, when booting a FAT filesystem, since there was no more room in the boot loader (at all!), setup.S locates and loads /bootopts on FAT filesystems. And in both cases, the kernel uses DEF_OPTSEG after boot (MINIX) or setup (FAT) and copies it to the kernel data segment.

I'm not sure how/whether this solves the bug you're talking about, except that 1) for MINIX, the FD or HD boot block has to match the kernel, and 2) for FAT, the setup block (prefixed to the kernel image) has to match the kernel.

Does that help, or something more sinister going on? I'm concerned about the potential QEMU issue, since that's my primary development emulator.

> a new kernel I generated (make clean; make). 

A full make should rebuilt the boot blocks in elks/bootbocks. However, if you use `make kclean; make kimage` like I do, that may not rebuild the boot sector - which could be a culprit.

PR #167, opened 2025-05-07
[kernel] Setup will now find where bootopts is loaded and relocate

Mellvik:
This PR fixes the problem described in #166: In a development setting, when OPTSEG changes, the bootblock and the kernel may be out of sync with regards to where `bootopts` (loaded by the bootblock) is located.

The fix is to have the bootblock code save the DEF_OPTSEG address used in a location where the kernel setup code can find it later. Not trivial since the SETUP_DATA address may also have changed. The solution is to use the unused top 16 bytes of the BIOS Data Area (BDA). Normally described as 'Intra-Applications Communications Area', this block is rarely if ever used, end definitely not used when TLVC is running. As such, these 16 bytes make up a convenient place to pass data during boot, possibly other settings too.

 The DEF_OPTSEG address occupies the first word of this block, at 0x4f:0.

`setup.S` checks this location - conveniently called BDA_IAC_SEG, compares to its own DEF_OPTSEG, and relocates `bootopts` if they don't match.

Also included some source cleanups (coding style) in `boot_minix.c` and - given the space constraints in said file, a few bytes were moved to `boot_block.S` instead. It is now possible to revert to the old 'Linux found' message during boot, but the current version is deemed good enough. With more space available, the `strcmp` implementation now has the end-of-string test of the second parameter reinstated: The function now works as expected in the rare case that the second string parameter is shorter than the first.

### DIFF
--- a/bootblocks/boot_minix.c
+++ b/bootblocks/boot_minix.c
@@ -118,9 +118,9 @@ void load_prog()
 
 static int strcmp(const char *s, const char *d)
 {
-	int c1 , c2;
+	int c1, c2;
 
-	while ((c1 = *s++) == (c2 = *d++) && c1 /* && c2*/);
+	while ((c1 = *s++) == (c2 = *d++) && c1 && c2);
 	return c1 - c2;
 }
 
--- a/bootblocks/boot_minix.c
+++ b/bootblocks/boot_minix.c
@@ -8,10 +8,14 @@
 #include "linuxmt/config.h"
 #include "minix.h"
 
+#define STR(x) #x
+#define STRING(x) STR(x)
+
 // Global constants
 
 #define LOADSEG DEF_INITSEG
-#define OPTSEG	DEF_OPTSEG             // bootopts copied here
+#define OPTSEG	DEF_OPTSEG		// bootopts copied here
+#define BOOTSEG 0x7C0			// Where we run
 
 // Global variables
 
@@ -19,7 +23,7 @@
 // as the context is not a user process
 // but code loaded in uninitialized memory
 
-static byte_t sb_block [BLOCK_SIZE];  // super block block buffer
+static byte_t sb_block[BLOCK_SIZE];  // super block block buffer
 #define sb_data	((struct super_block *)sb_block)
 
 static int i_now;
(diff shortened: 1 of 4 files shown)


bootblocks/boot_minix.c
```diff
@@ -74,18 +79,22 @@ static void load_file ();
 void load_prog()
 {
 	// Avoid reuse of an old copy of /bootopts in memory if we're rebooting w/ no /bootopts */
-	int __far *optseg = _MK_FP(OPTSEG, 0);	/* Expensive, is there a better way? */
-	*optseg = i_boot = i_now = 0;
+	//int __far *optseg = _MK_FP(OPTSEG, 0);	/* Expensive, is there a better way? */
+	i_boot = i_now = 0;
+	//*optseg = i_boot = i_now = 0;
```
ghaerr, line 84: Since you've commented out the previous fix to zero out the first bytes of the OPTSEG, how is it that this still works when /bootopts is deleted and then the system rebooted?

bootblocks/boot_minix.c
```diff
@@ -107,14 +116,11 @@ void load_prog()
...
-static int strcmp (const char * s, const char * d)
+static int strcmp(const char *s, const char *d)
 {
-	const char * p1 = s;
-	const char * p2 = d;
-
 	int c1, c2;
 
-	while ((c1 = *p1++) == (c2 = *p2++) && c1 /* && c2*/);
+	while ((c1 = *s++) == (c2 = *d++) && c1 && c2);
```
ghaerr, line 123: In reviewing this PR for possible enhancements to ELKS, research shows that none of the changes actually save any space except for the move of puts to the boot sector \_puts (which definitely saves space).

> With more space available, the strcmp implementation now has the end-of-string test of the second parameter reinstated: The function now works as expected in the rare case that the second string parameter is shorter than the first.

Actually, 'c2' was commented out in the strcmp while loop because it's not needed - not because strcmp was buggy. You'll notice that the while loop continues while c1 == c2, so only c1 needs to be tested for non-NUL to end the loop. Adding the c2 back in is superfluous and only adds code that's not needed. Also, with the optimizing compiler, using local variables p1/p2 don't change the code emited at all.

bootblocks/boot_minix.c
```diff
@@ -201,7 +207,7 @@ static void load_file ()
 
 	// Indirect zones
 	load_zone (1, &(i_data->i_zone [ZONE_IND_L1]), &(i_data->i_zone [ZONE_IND_L2]));
-	if (f_pos >= i_data->i_size) return;
+	//if (f_pos >= i_data->i_size) return;
```
ghaerr, line 210: Probably nice to have this commented out (but moved below near the double-indirect zone so as to not forget when needed), but FYI commenting this out also saved no space, since the optimizing compiler knows the next statement is an implicit return anyways.

bootblocks/boot_minix.c
Mellvik: Thanks @ghaerr, 

> In reviewing this PR for possible enhancements to ELKS, research shows that none of the changes actually save any space except for the move of puts to the boot sector _puts (which definitely saves space).

I know. The changes were made for clarity. The extra 'stuff' got me curious enough to check whether it had any effect. In a year or two I may have forgotten, or someone else is asking themselves the same question. Thus the simplification.

As to the change to `strcmp` - you're right, it is superfluous. There was a reason for the change, now completely uninteresting. Thanks. 
While turning seemingly every single byte of the code around - on a quest to squeeze a few more bytes out, I wrote a new strcmp using inline `asm()` macros and saved half a dozen bytes, only to find that GCC removed the string literal in the argument list  because they were deemed unused. Waste of time - except I learned quite a bit about GCC's optimization capabilities. The way it turns `strcmp` into less than 30 bytes blew my mind - until I found that it actually has special provisions for the `str...` family of functions. Still very impressive ...

bootblocks/boot_minix.c
Mellvik: Again, changed for clarity only. 
While on the subject of squeezing code, it may be an idea to move the strcmp function to the boot sector - to take off some of the pressure in boot_minix, possibly allowing the double-indirect code to be included (again). 
As usage evolves, files get larger and file system life span longer, that may be a good thing.  I'm seeing file systems lasting for months, close to a year at times, now that the system has become stable and rarely screws up the filesystems anymore. System updates are passed via the net, untarred locally, copies are kept around and there are at times thousands of files in a minix file system. This reality motivated the changes to the boot code in the first place, completely disconnecting the boot code from the kernel.

ghaerr: I wouldn't worry about double-indirect block code ever being needed, as it'd only be required for kernels >= 265,728 bytes. Given ia16-elf-gcc's max of 64K near text, 64k fartext and 64 data, I don't see how such a kernel could possibly be built.

PR #168, opened 2025-05-07
[kernel] replaced fmemcpyw and xms_fmemcpyw with equally fast byte versions

Mellvik:
Originally discussed in  https://github.com/Mellvik/TLVC/pull/165#issuecomment-2848625138, this PR removes word size memory copy routines for xms and far memory, instead using the byte versions of the same routines, which happens to use word size whenever possible. So there is no performance difference between the two, and eliminating the word-routines saves an estimated  500+ bytes of code space.

While the entry point for `fmemcpyw` has been kept (no code), the `xms_fmemcpyw` and `linear32_fmemcpyw` entries have been completely removed (commented out).

In a separate commit, the source code will be removed.

Thanks for the discussion, @ghaerr.

### DIFF
--- a/tlvc/arch/i86/lib/fmemory.S
+++ b/tlvc/arch/i86/lib/fmemory.S
@@ -26,7 +26,7 @@
 //		size_t count)
 
 fmemcpyb:
-	clc
+	or     $1,%al		// clear ZF
 2:	mov    %si,%ax
 	mov    %di,%dx
 	mov    %sp,%si
@@ -35,7 +35,7 @@ fmemcpyb:
 	les    ARG0(%si),%di  // far destination pointer
 	lds    ARG2(%si),%si  // far source pointer
 	cld
-	jc     3f
+	jz     3f
 	shr    $1,%cx         // copy words
 3:	rep
 	movsw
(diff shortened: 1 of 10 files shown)


tlvc/arch/i86/lib/fmemory.S
```diff
@@ -51,22 +53,8 @@ fmemcpyb:
...
-	cld
-	rep
-	movsw
-	mov    %ax,%si
-	mov    %dx,%di
-	mov    %ss,%ax
-	mov    %ax,%ds
-	mov    %bx,%es
-	ret
+	stc
```
ghaerr, line 56: I'm confused as to how this is supposed to work - setting CF and jumping to the fmemcpyb routine will do word moves, then always RCL the CF back into CX in Line 142, causing a final byte to always be copied: IMO one more byte than wanted every time?

Mellvik: Heck, you're right! Thanks for reviewing, I missed that one and testing didn't catch it (yet). Appreciated. Fix coming.

Mellvik:
The latest commit fixes the 'last byte' bug identified by @ghaerr, and adds a formerly discussed optimization to the `loadall_block_move` function.

PR #173, opened 2025-05-25
[kernel] Allow floppy DMA to/frm XMS memory, probe fixes

Mellvik:
Adds floppy DMA to/from XMS memory as discussed in #164 and in ELKS https://github.com/ghaerr/elks/pull/2339#issue-3084933312. Floppy IO no longer needs to pass via the bounce buffer when XMS buffers are involved.

Also included a bug fix (and some cleanups, optimizations) in the probe code that in some cases prevented probes to succeed after diskette changes.

### DIFF
--- a/tlvc/arch/i86/drivers/block/directfd.c
+++ b/tlvc/arch/i86/drivers/block/directfd.c
@@ -128,8 +128,6 @@ void (*DEVICE_INTR) () = NULL;
 			 * it back than pretending it's possible to save anything - detecting,
 			 * flushing, umounting etc. */
 			/* Always use off for PC/XT/8086/8088 8bit ISA systems */
-#else
-#undef CHECK_MEDIA_CHANGE /* Now defined in linuxmt/fs.h */
 #endif
 
 #define DEBUG_DIRECTFD 0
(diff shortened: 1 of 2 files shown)


tlvc/arch/i86/drivers/block/directfd.c
```diff
@@ -806,11 +813,10 @@ static void bad_flp_intr(void)
 #endif
     errors = ++CURRENT->rq_errors;
     if (errors > MAX_ERRORS) {
-        printk("df%d: Max retries (%d) exceeded\n", CURRENT->rq_dev&3, errors);
+        if (!probing) printk("df%d: Max retries (%d) exceeded\n", CURRENT->rq_dev&3, errors);
 	request_done(0);
 	return;
     }
-    if (probing) return;
```
ghaerr, line 813: Removing this return when probing will cause an FDC controller recalibrate each probe sub-type - is that really what you want? A recalibrate is needless and slow, don't some controllers also reseek the head on a recal?

Mellvik: That's attention to detail, @ghaerr. And you're right. This statement was removed only after changing the code in `rw_interrupt` to set the error count to MAX_ERRORS when probe is found to fail. That way we'll never get to this point in `bad_flp_intr` when probing.

Thanks.

PR #182, opened 2025-06-11
[kernel] Optimizations in init/main.c

Mellvik:
Instead of allocating a 1k buffer for the local copy of `bootopts` in `init/main.c`, the copy is now moved to the high end of the kernel data segment, which eventually becomes the heap (unless shrunk with the `heap=` `bootopts` directive, which doesn't matter for this particular temporary use). The change effectively reduces the kernel bss memory footprint, by keeping the heap area contiguous.

The assumption that the system RAM conversion the entire kernel data segment is a safe one, as we're always below 200k and the system will not be useful in less than 256k anyway.

Also in this PR is a change to the `bootopts` memory size directive, formerly `memory=`, now `mem=`,  saving 4 bytes of precious `bootopts` space. 

Apropos precious bootopts space, here's a related thought: What if the initial loading of `bootopts` at boot time could be done to top of memory instead of low memory? Right now that's a non-issue for TLVC since the floppy bounce buffer is used, so at its current size it doesn't take up any space. However, with `bootopts` continuously bursting at the seams, having the flexibility to increase its size to 1.5 or 2k seems advantageous. Moving the copies away from both low memory and BSS opens up the floodgates - `bootopts` may be literally any size with minimal consequences/changes.

@ghaerr, what's your take on this line of thinking?

### DIFF
--- a/tlvc/init/main.c
+++ b/tlvc/init/main.c
@@ -110,7 +110,13 @@ static char *argv_init[MAX_INIT_SLEN] = { NULL, bininit, NULL };
 #if ENV
 static char *envp_init[MAX_INIT_ENVS];
 #endif
+
+#define OPTIONS_IN_TOPHEAP
+#ifdef OPTIONS_IN_TOPHEAP	/* working copy of bootopts at top of heap */
+static unsigned char *options = (unsigned char *)(0xfffe - OPTSEGSZ);
+#else
 static unsigned char options[OPTSEGSZ];
+#endif
 
 extern int boot_rootdev;
 extern int dprintk_on;
@@ -370,8 +376,10 @@ static void INITPROC do_init_task(void)
 	seg_add(REL_INITSEG + (fdcache<<6), XD_BOUNCESEG);
 
 #ifdef CONFIG_BOOTOPTS
+#ifndef OPTIONS_IN_TOP_HEAP
     /* Release /bootopts parsing buffers */
     heap_add(options, OPTSEGSZ);
+#endif
 
     /* pass argc/argv/env array to init_command */
 
@@ -577,7 +585,8 @@ static int INITPROC parse_options(void)
 	char *next;
 
 	/* copy /bootops loaded by boot loader, possibly relocated by setup */
-	fmemcpyb(options, kernel_ds, 0, DEF_OPTSEG, sizeof(options));
+	printk("\nmoving bootopts to %x:%x\n", kernel_ds, line);
+	fmemcpyb(options, kernel_ds, 0, DEF_OPTSEG, OPTSEGSZ);
 
 #pragma GCC diagnostic ignored "-Wstrict-aliasing"
 	/* check file starts with ##, one or two sectors, max 1023 bytes */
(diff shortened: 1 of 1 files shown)


ghaerr:
> the init/main.c part is done and works well so far. It's really simple since all is done before the heap has been even initialized.

No, it's only working because the change hasn't been regression tested. While the options[] buffer is parsed early in kernel init and all 'int' values saved elsewhere, all 'char *' values, such as where init_command, argv_init[] and envp_init[] point are all contained in the opt.options[] buffer. This means that, for instance, init=, all init arguments and the environment variables all stand the chance of being overwritten should a heap allocation (low or high) end up on top of them. Without a rewrite, the options[] buffer needs to be present all the way through when init_task finally execs /bin/init or its designated replacement. In other words, options[] must be valid until that point in the original code where it is released by adding it to the kernel heap in the heap_add() late in the kernel init.

There are some ways this can be made to work - but the init_command/argv/envp code will have to be rewritten, or a way to protect a heap area, as suggested above. Is it worth it, considering that the options buffer is already released as a separate heap segment, which nicely matches some smaller allocations and keeps them out of fragmenting the main kernel heap? That's what I am wondering about. It would seem not much of this is really useful unless a larger /bootopts file size is supported. I don't recall the data on exactly the allocations occurring in the options heap segment, but I thought they solved a reported fragmentation problem.

Are you still using the high-heap kernel allocation mechanism for some allocations, or did that become less useful when the options[] buffer ended up being added separately?

> The FAT variant is a good point. I never use that - and need to decide whether to support it or not ...

Well, I guess that depends on whether you're writing TLVC only for yourself, or whether you're adding features for other users :) I think it would be a mistake to drop FAT /bootopts support, especially at the "expense" of programmer time.

Mellvik:
Very good points @ghaerr, and you're right, there has been no regression testing (yet). 

What makes this work is the fact that the high-heap allocation for small chunks is disabled because the heap_add of the options buffer turned out to have about the same effect. Thus - in my testing - all heap allocations are from the bottom, and using the top for bootopts is safe: We won't get there until the system has been running for a while, maybe never. Enabling the top-down allocation and having a 'full' bootopts will break it.

IOW - since the separation of small allocations from the 'rest' is quite useful, the system is now worse off than before -- with increased heap fragmentation. Not acceptable. So there are two options:
- Just forget it. Keep the options buffer as is and increase its size when/if `bootopts` expands. This will lessen the benefits of the buffer when released to the heap, while possibly make the heap a little less efficient.
- Add a padding to the block where `bootopts` is copied, moving it an extra .5 or 1k down from the heap top. No extra code, easy to expand when/if bootopts grows and the 'gravitate to top' for small heap allocations can be re-enabled. There are typically 4 small (3 seg_descriptors, 1 tty buffer) allocations to be concerned about.

While looking into whether moving the initial `bootopts` load to high memory instead of low is possible and viable. - and out of curiosity, I took a look at heap behaviour using  the 'old' (static buffer) scheme with a 2k buffer. It still catches most of the small allocations, IOW it works well. So odds are the original scheme will be kept.

Thanks, useful discussion.

PR #186, opened 2025-06-23
[kernel] Allow 'unlimited' bootopts filesize on FAT boot volumes

Mellvik:
With this relatively modest change, FAT boot volumes now handle 'large' `bootopts` files like minix file systems do (#184 ). The implementation is not failsafe in that it does assume the `bootopts` file to be contiguous. 

However, in practical testing, the assumption that `bootopts` is located early in the directory structure is as much if not more of a problem. Both need to be fixed.

In normal use, and unless the file system is very full, the stability is good.

PC98-related code has not been updated.

### DIFF
--- a/bootblocks/boot_minix.c
+++ b/bootblocks/boot_minix.c
@@ -31,7 +31,7 @@ static unsigned loadaddr;
 static int ib_first;                 // inode first block
 
 // locals made global to save code space
-//static int i_size;
+static int i_size;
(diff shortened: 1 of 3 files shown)


Mellvik:
More squeezing of `minix_boot.c` making it possible to save `bootopts` file size, simplifying `setup.S` somewhat. Also some related optimizations in `init/main.c`.

PR #189, opened 2025-06-27
[ktcp] A work-in-progress checkin, lots of enhancements & fixes

Mellvik:
This PR fixes several bugs and long time performance issues with `ktcp`, most importantly its inability to talk to itself reliably on slow hardware. A summary of the changes and the problems solved follows (mostly developer notes).
- The timer (frequent calls to `gettimeofday()`) is now running on kernel jiffies, simpler and much more efficient. The main loop timer is almost unchanged, now at 60ms instead of 62.5. I'm still experimenting with the possible benefits of using a shorter time tick for the main loop. A higher clock resolution may be beneficial in the still ongoing work to improve the retransmission mechanism. 
- The RTT convergence algorithm (in `tcp_output.c`) has been changed (changes in `tcp.h`) to eliminate the need for longs, and is now dramatically smaller and more efficient without changing its functionality.
- A simplified implementation of the (TCP mandatory) 'slow start' and 'congestion avoidance' algorithms is included. A more complete implementation has been evaluated and found useless under the circumstances. What this implementation does is to enable `ktcp` to talk to other really slow (usually `ktcp` based) systems, which didn't work before. That said, the implementation needs more tuning to become better at adapting to 3 orders of magnitude difference in peer speed (which is admittedly somewhat eased by the fact that we're only supporting 10Mbps NICs).
- The handling of retransmits has been improved but the results are not yet satisfactory. The remaining problem is that when a packet in the retransmit queue times out and gets retransmitted, all the other packets belonging to that connection will have timed out too - and will be dumped onto the wire immediately without waiting to see if the first retransmit generates an ack for the entire bunch of packets in the buffer (which is usually the case). In almost all cases when there is a retransmit, it will be `ktcp` that lost an ack. I've been experimenting with various simple algorithms to make sure there is time for a possible incoming ack to be read and processed before we ship off the whole shebang. Work in progress.
- The 'good old' double ack problem (or feature, since it was put there on purpose) has been worked on but its completion has yet to be verified. It's related to the silly window syndrome and needs some smartness because of the rather odd timings involved when slow systems are talking. Stability is not a problem, the challenge is optimization. When the other end is `ktcp`, unnecessary double acks disturbed the flow. For fast peers, it doesn't matter all that much.
- An old ARP problem that caused the ARP table to be filled with unanswered requests and eventually blocking new requests, has been fixed.
- Also fixed is a long time bug that in some cases caused `kcp` to loop in its main loop (while still working) - as if there were more data to process. This usually happened when a transfer failed, and most often an incoming transfer (like running out of space for an incoming file). What would happen was that the connection Wass terminated properly, but the `tcpcp_need_push` didn't get decremented. This now works, but there may still be other situations that cause the same behaviour, so we may not be completely out of the bushes yet.

With the exception of the two bug fixes mentioned above, this code has been running and in use daily for about 12 months, and - with the caveats mentioned below - is considered stable.

Problems pending
- There are still occasional hangs under heavy load, not predictably repeatable (yet) and possibly driver related.
- There are still situations on XT class systems that may cause long waits at seemingly random intervals during a (significant sized) transfer - similar to what we see when the shot is busy syncing to floppy.
- `in_atoi` (really `in_gethostbyname`) has no sanity check on input whatsoever, which means that IP addresses (for example) can be anything and will still be converted into something that ktcp tries to use. The only sanity check at this time is that the first character is a number in `in_gethostbyname`.

This is still work in progress - the code is littered with odd-looking debug statements in addition to the 'regular' ones governed by `config.h`.

Finally - a more detailed discussion about the considerations and and choices made in the implementation of slow start/congestion control will be posted in the developer notes wiki.

### DIFF
--- a/tlvccmd/ktcp/arp.c
+++ b/tlvccmd/ktcp/arp.c
@@ -31,7 +31,7 @@ static void arp_prep_request(struct arp *, ipaddr_t);
 
 int arp_init (void)
 {
-	memset (arp_cache, 0, ARP_CACHE_MAX * sizeof (struct arp_cache));
+	memset(arp_cache, 0, ARP_CACHE_MAX * sizeof (struct arp_cache));
 	return 0;
 }
 
@@ -44,14 +44,15 @@ struct arp_cache *arp_cache_get(ipaddr_t ip_addr, eth_addr_t eth_addr, int flags
 		if (!entry->ip_addr)
 		    break;
 		if (entry->ip_addr == ip_addr) {
+			if (flags == -1) return entry; /* Already waiting for reply */
 			if ((flags & ARP_VALID) && !entry->valid)
 				return NULL;	/* not yet valid - awaiting ARP reply*/
 			if (flags & ARP_UPDATE) {
-				memcpy (entry->eth_addr, eth_addr, sizeof (eth_addr_t));
+				memcpy(entry->eth_addr, eth_addr, sizeof (eth_addr_t));
 				debug_arp("arp: merging cached entry for %s (%s)\n",
 					in_ntoa(ip_addr), mac_ntoa(entry->eth_addr));
 			} else {
-				memcpy (eth_addr, entry->eth_addr, sizeof (eth_addr_t));
+				memcpy(eth_addr, entry->eth_addr, sizeof (eth_addr_t));
 				debug_arp("arp: using cached entry for %s (%s)\n",
 					in_ntoa(ip_addr),mac_ntoa(entry->eth_addr));
 			}
@@ -70,7 +71,7 @@ struct arp_cache *arp_cache_update(ipaddr_t ip_addr, eth_addr_t eth_addr)
 	eth_addr_t existing_addr;
 
 	if ((entry = arp_cache_get(ip_addr, existing_addr, 0))) {
-		memcpy (entry->eth_addr, eth_addr, sizeof (eth_addr_t));
+		memcpy(entry->eth_addr, eth_addr, sizeof (eth_addr_t));
 		debug_arp("arp: updating cached entry for %s (%s)\n",
 			in_ntoa(entry->ip_addr), mac_ntoa(entry->eth_addr));
 		entry->valid = 1;
@@ -83,21 +84,26 @@ struct arp_cache *arp_cache_add(ipaddr_t ip_addr, eth_addr_t eth_addr)
 {
 	struct arp_cache *entry;
 
+	/* Check if we're already waiting for a reply for this address */
+	if ((entry = arp_cache_get(ip_addr, eth_addr, -1))) {
+		debug_arp("arp: reusing unresolved entry for %s\n", in_ntoa(ip_addr));
+		return entry;
+	}
 	/* Shift the whole cache */
 	entry = arp_cache + ARP_CACHE_MAX - 1;
 	while (entry > arp_cache) {
-		memcpy (entry, entry - 1, sizeof (struct arp_cache));
+		memcpy(entry, entry - 1, sizeof (struct arp_cache));
 		entry--;
 	}
 
 	/* Set the first entry as the more recent */
 	entry->ip_addr = ip_addr;
 	if (eth_addr) {
-		memcpy (entry->eth_addr, eth_addr, sizeof (eth_addr_t));
+		memcpy(entry->eth_addr, eth_addr, sizeof (eth_addr_t));
 		entry->valid = 1;
 	} else {
 		entry->valid = 0;	/* no MAC address yet, awaiting ARP reply*/
-		memset (entry->eth_addr, 0, sizeof(eth_addr_t));
+		memset(entry->eth_addr, 0, sizeof(eth_addr_t));
 	}
 	entry->qpacket = NULL;
 	debug_arp("arp: adding cache entry for %s, valid=%d\n", in_ntoa(ip_addr), entry->valid);
--- a/tlvccmd/ktcp/config.h
+++ b/tlvccmd/ktcp/config.h
@@ -10,7 +10,7 @@
 #define DEBUG_TCP	0	/* TCP ops*/
 #define DEBUG_TCPPKT	0	/* TCP packets info*/
 #define DEBUG_TCPDATA	0	/* TCP packet data display*/
-#define DEBUG_TUNE	1	/* tuning options*/
+#define DEBUG_TUNE	0	/* tuning options*/
 #define DEBUG_RETRANS	0	/* TCP retransmissions*/
 #define DEBUG_WINDOW	0	/* TCP window size*/
 #define DEBUG_ACCEPT	0	/* TCP accept*/
--- a/tlvccmd/ktcp/deveth.c
+++ b/tlvccmd/ktcp/deveth.c
@@ -44,7 +44,7 @@ int deveth_init(char *fdev)
     }
 
     /* read mac of nic */
-    if (ioctl (devfd, IOCTL_ETH_ADDR_GET, eth_local_addr) < 0) {
+    if (ioctl(devfd, IOCTL_ETH_ADDR_GET, eth_local_addr) < 0) {
         printf("ktcp: IOCTL_ETH_ADDR_GET fail\n");
 
         /* MAC not available is a fatal error */
@@ -64,9 +64,10 @@ int deveth_init(char *fdev)
 void eth_process(void)
 {
   eth_head_t * eth_head;
-  int len = read (devfd, sbuf, MAX_PACKET_ETH);
+  int len = read(devfd, sbuf, MAX_PACKET_ETH);
   if (len < (int)sizeof(eth_head_t)) {
-	if (len < 0) printf("ktcp: eth_process error %d (errno %d), discarding packet\n", len, errno); //FIXME
+	/* FIXME: Is this really a 'tryagain' situation? It's always non-fatal, do we need the message? */
+	if (len < 0) printf("ktcp: eth_process error %d (errno %d), discarding packet\n", len, errno); 
 	return;
   }
 
@@ -74,8 +75,8 @@ void eth_process(void)
 
 #if 0
   /* Filter on MAC addresses in case of promiscuous mode*/
-  if (memcmp (eth_head->eth_dst, broad_addr, sizeof (eth_addr_t))
-    && memcmp (eth_head->eth_dst, eth_local_addr, sizeof (eth_addr_t)))
+  if (memcmp (eth_head->eth_dst, broad_addr, sizeof(eth_addr_t))
+    && memcmp (eth_head->eth_dst, eth_local_addr, sizeof(eth_addr_t)))
       return;
 #endif
 
@@ -87,7 +88,7 @@ void eth_process(void)
 	  break;
 
   case ETH_TYPE_ARP:
-	  arp_recvpacket (sbuf, len);
+	  arp_recvpacket(sbuf, len);
 	  break;
   }
   netstats.ethrcvcnt++;
@@ -105,7 +106,7 @@ void eth_route(unsigned char *packet, int len, ipaddr_t ip_addr)
 	unsigned char *p;
 
 	/* try to get cached ethernet address and send packet*/
-	if (arp_cache_get (ip_addr, eth_addr, ARP_VALID)) {
+	if (arp_cache_get(ip_addr, eth_addr, ARP_VALID)) {
 		eth_sendpacket(packet, len, eth_addr);
 		return;
 	}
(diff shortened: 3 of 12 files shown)


ghaerr:
Interesting set of improvements and fixes to networking. I haven't had time to look at any of the code yet, but will certainly want to consider adding some of this to ELKS. Which brings me to the subject of testing - I'd like to figure out a way to somehow construct a software test rig for running ktcp between itself as well as to/from host TCP, possibly with or without ethernet drivers (and/or SLIP), that is, in an emulated environment. Have you thought much about how that might be possible?

Obviously real world testing is paramount, but it'd really nice to somehow separate out various layers of the network stack in a QEMU or other environment for simulated congestion/window management, RTT testing, etc, as well as finally introducing some regression testing. I haven't yet tried multiple-instance QEMU's talking to each other, but such a thing would seem plausible. Also some ways to introduce delays or speedups. Back when I was deeply involved fixing the ktcp internals, I'd also thought about implementing a more comprehensive loopback interface - currently it's extremely simple.

Thoughts?

Mellvik:
Interesting thoughts, @ghaerr.  The idea of adding a scripted test suite has been on my list for some time, but admittedly not a more comprehensive test environment like you propose. While the former would be immediately useful for monitoring progress, experimenting with settings and performance measurements, I'm not sure the latter is even possible with reasonable effort - while arguably very desirable. There are so many variables in this equation - I seem to be discovering new ones still, every time `ktcp` is connecting to a new peer with a different TCP implementation which behaves differently - while still within the standard.

While testing and experimenting with the enhancements in this release- using 5 quite different (speed wise) hardware variants for TLVC and several different peers, it became obvious that the goal cannot be to make `ktcp` work optimally for all of them. There are compromises to be made - which means optimize reasonably for the most common scenario, make sure the others are handled reliably. Robustness over speed. Which is where we are right now. Connecting two floppy based systems works, but is far from optimal.

There are - as the PR notes says - more optimizations pending, retransmits in particular. And there is more tuning pending. Interestingly - along the way, the behaviour of `ktcp` was found to be significantly different depending on whether traffic passed through a router or not, the router simply connecting high speed LAN segments. Timing related differences - that change the order of packet processing in the main `ktcp` loop. Visible on some systems, not all, depending on processor speed. The explanation being that when there is a router in between, it is the router rather than the host that is adapting to the physical speed differences between 10mbps and whatever is on the other side, 100M, 1G, 10G. `ktcp` is seeing the router's tcp implementation more than the peer's. Also, when connecting to a server in Amsterdam, to add latency, a new picture appeared. Unsurprising in itself but surprising in its effects. RTT is practically invariable on a LAN segment, adding these two examples expanded the 'universe' so to speak, making latency larger and variable. A 'test rig' could do that, but I'm not convinced it would be good enough because it's hardly ever one parameter alone that creates a significant difference. 

The most important tools in the process have been `ftp`, `ping` and `tcpdump` - with `nc` being a welcome addition at times.  `ping`, with all its options, is incredibly valuable, in particular the interval and packet size options. `tcpdump` invaluable in its ability to show exactly what's going on on the wire. So if I were to prioritize work to enhance the test environment for `ktcp`, it would be a TLVC/ELKS implementation of `ping`. Not having `ping` when testing and debugging ktcp<->ktcp connections is a serious drawback.

Another 'dream tool' in this setting would be the ability to live-change key parameters in `ktcp` enabling live tuning/experimentation, and thus speeding up the process significantly. 

My #3 choice is an oddity, seemingly unrelated: A `top` command - for many reasons, most of them unrelated to networking. Sticking to `ktcp` testing, the one thing I've frequently missed that `top` could provide is real time process resource usage stats, a terminal window running `top` to monitor ktcp cpu usage in real time. It's coarse but it's good enough to see if `ktcp` (or a network daemon) is looping. And low threshold I would think, with very few options needed. Like the `cons` command with the content taken from `ps` instead of 0xb800. 

Not exactly what you asked, but some loud thinking about tooling, testing and tuning.

ghaerr:
Hello @Mellvik,

Thanks for your help and explanations regarding getting these enhancements ported to ELKS! As I've been looking deeper into the code, I started looking at the enhancement of getting rid of the gettimeofday() system call along with the many kernel and user 32-bit divisions that have to take place - nice. Then I started thinking about optimizing the `Now = *jp / 6` in the main loop and it occurred to me there's a potential bug here: it might not happen much, but have you ever found ktcp timing to act strangely, even just for a moment? Looking at the ASM code generated for that statement:
```
 155:   36 c5 1e 00 00          lds    %ss:0x0,%bx
                        158: R_386_16   jp
                        158: R_386_SUB16        jp!
 15a:   8b 07                   mov    (%bx),%ax     // <--- problem here
 15c:   8b 57 02                mov    0x2(%bx),%dx  // <--- and here
 15f:   31 c9                   xor    %cx,%cx
 161:   51                      push   %cx
 162:   b9 06 00                mov    $0x6,%cx
 165:   51                      push   %cx
 166:   52                      push   %dx
 167:   50                      push   %ax
 168:   16                      push   %ss
 169:   1f                      pop    %ds
 16a:   e8 fe ff                call   16b <ktcp_run+0x115>
                        16b: R_386_PC16 __udivsi3
 16d:   83 c4 08                add    $0x8,%sp
 170:   a3 00 00                mov    %ax,0x0
                        171: R_386_16   Now
                        171: R_386_SUB16        Now!
 173:   89 16 02 00             mov    %dx,0x2
                        175: R_386_16   Now
                        175: R_386_SUB16        Now!
```
The potential bug is that kernel `jiffies` is volatile and, if the clock tick happens just between these instructions, would produce an incorrect value. Admittedly, would only be possible on 109 minute intervals, and I think would only produce a time less than the correct time, since the low word would be zero and the high word unchanged.

Now that I write this up, I'm thinking this is completely overkill on my part and probably no action needed - but I thought to point it out. I thought this might be in the precision timer library code also, but turns out that was handled by disabling interrupts while both the PIT countdown was read, *and* jiffies.

Mellvik:
Very interesting indeed, @ghaerr - thanks for the code review! 

I see the problem - it didn't occur to me that jiffies, being so heavily used all over the place, was volatile. But of course it is,  it's just a memory location. 

Now, are you also saying that a cli/sti pair wrapping the two mov instructions would fix the problem (sorry, I haven't looked at your fix yet)?

While testing, I've had runs spanning more than 30hrs, with a few letters printed per packet. I obviously haven't paid attention to the output all that time, but being visible on the side of the screen I think I would have noticed significant anomalies. I mean, jumping 64k jiffies back - if just for a brief moment - should be very visible.

Very interesting problem - I need to ponder this further in the morning.

ghaerr:
> jiffies is being used in lots of places as is
> Interrupts are indeed enabled when jiffies is incremented

You bring up a very good point - the use of kernel jiffies with the same problem of interrupted *kernel* code rendering an incorrect value - thank you, I had never considered that! Because I care about correctness (even though it may actually not matter in many or all of these cases) I am going to look further into this with each kernel usage as well as the code generated. On the 8086 there are only two single instructions, LDS and LES that load 32-bits atomically, but the 32-bit kernel jiffies loads will undoubtedly use two instructions, loading the DX:AX separately (also as shown above). I will report back.

> this is presumably happening in the IRQ0 interrupt handler. Given its priority, what are the chances it may be interrupted?

No chance (other than NMI handlers), since the PIT is the highest-priority interrupt. So guarding the jiffies increment won't help the problem.

> It seems easier to just wrap the jiffies++ in do_timer (kernel/sched.c) in a cli/sti pair,

As per above, the IRQ 0 hardware timer interrupt will interrupt any other kernel or user code, so cli/sti in the IRQ 0 handler won't have the desired effect. Instead, we need an atomic way of loading 32 bits. The LOCK prefix comes to mind, but I don't think that works for multiple instructions.

Meanwhile, the business of accessing jiffies in user mode using a far pointer is a bit different. Since it's a far pointer, it takes at least six instructions to access `*jp`. See above, basically LDS BX is used to load the far address of jiffies into DS:BX, then two instructions to load jiffies using 0(BX) and 2(BX), then another two instructions to restore DS. So the idea of using a `get_time()` wrapper to access `*jp` saves a lot of code generated every time `*jp` is seen. Also, IMO the time quantum (\*jp vs \*jp / 6) should be the same, so having the get_time() wrapper function return \*jp / 6 is superior.

That said, I could agree, depending on further analysis, that there isn't a big requirement for cli/st in get_time(), although it will be required for correctness (and the same sort of thing changed in the kernel to - which does seem like overkill for sure).
Thus just removing cli/sti in get_time for ktcp won't likely change anything one would ever notice.

It appears the only real "correctness" error possible is that in the rare case where 1) once every 109 minutes the IRQ 0 handler interrupts 2) exactly between the two instructions used to access kernel jiffies or user mode \*jp, that the lower half of jiffies will be incremented to 0 in the actual jiffies, but remain unchanged in the user mode lower 16-bits, while the upper half of jiffies will be increment up by 1 and the user mode high 16-bits will increment by 1. This will have the net effect of having the perceived time be +109 minutes for 10ms, then revert to correct time. 

The question is, what does this hurt? I will check the kernel usage and report back. For ktcp, timers would go off when they shouldn't, which, given the nature of TCP/IP, one would think they'd be recoverable, except for certain close timeouts etc. All in all, probably a non-issue. Is it worth being correct with two more instructions?

Thank you for your comments!

ghaerr:
[More discussion on 32-bit atomic reads](https://forum.vcfed.org/index.php?threads/atomic-read-of-32-bit-integer-from-memory-on-x86.59100/). Looks like LDS/LES would be the answer, but this won't fit well into code produced by ia16-elf-gcc.

Our discussion is interesting from the standpoint of what it takes to actually be completely correct, versus corrupted data and what's using that data.

ghaerr:
The use of kernel `jiffies`  is unprotected in the following areas:
- filesystem date/time get/set (filetimes could be off by 109 minutes)
- kernel busy-wait timers in ee/lance NIC (unknown risk, didn't study code - please do)
- kernel busy-wait timers DF, IDE, ATA and LP drivers (could result in long wait times in rare cases)
- kbd polling (could result in no polling for 109 minutes)
- PC98 serial xmit timeout (could result in UART xmit failure)
- TTY drive VTIME timeout (could result in TTY/serial input hang)
- select, alarm and gettimeofday system calls (potential long timeout or incorrect time returned)

The tighter the loop (e.g. busy-loops using jiffies) the more potential for error, since jiffies value is retrieved constantly while comparing to exit value.

`jiffies` is protected and handled properly in:
- kernel timeout function scheduler
- precision timer kernel library

Ultimately, all of this obviously isn't a big deal, but could at times be contributing to undefined behavior. So I went ahead and  designed some sample code to handle the problem, should this ever want to be addressed. A solution would be to use a `get_jiffies()` macro in all places where jiffies is referenced, and replaced with get_jiffies(). The following code uses exactly three more instructions than previously and protects the system completely by disabling interrupts then returning to previous interrupt state:

C code:
```
/* test program for protected access to jiffies */
volatile unsigned long jiffies;

#define get_jiffies()                       \
    __extension__ ({                        \
        unsigned long v;                    \
        asm volatile ("pushfw\n"            \
                      "cli\n"               \
                        : /* no input */    \
                        : /* no output */   \
                        : "memory");        \
        v = jiffies;                        \
        asm volatile ("popfw\n"             \
                        : /* no input */    \
                        : /* no output */   \
                        : "memory");        \
        v; })

unsigned long badway(void)
{
    return jiffies;
}

unsigned long goodway(void)
{
    return get_jiffies();
}
```
ASM output for comparison between "badway" (existing code) and "goodway" protected code:
```
0000000 <badway>:
   0:	a1 00 00             	mov    0x0,%ax
			1: R_386_16	jiffies
			1: R_386_SUB16	jiffies!
   3:	8b 16 02 00          	mov    0x2,%dx
			5: R_386_16	jiffies
			5: R_386_SUB16	jiffies!
   7:	c3                   	ret

00000008 <goodway>:
   8:	9c                   	pushf
   9:	fa                   	cli
   a:	a1 00 00             	mov    0x0,%ax
			b: R_386_16	jiffies
			b: R_386_SUB16	jiffies!
   d:	8b 16 02 00          	mov    0x2,%dx
			f: R_386_16	jiffies
			f: R_386_SUB16	jiffies!
  11:	9d                   	popf
  12:	c3                   	ret
```
No function calls are necessary - functions were just used to easily encapsulate the generated code in both cases.

ghaerr:
PR https://github.com/ghaerr/elks/pull/2458 introduces atomic access to the kernel jiffies variable, fixing the potential data corruption problems discussed here.

The same solution has already been added to ELKS user mode ktcp. I'm not ready to resume work on porting the TLVC NIC drivers over to ELKS after you've had a chance to make any required adjustments to them from the conversations over this last week. Thanks!

PR #192, opened 2025-07-07
[boot] Allow bootsector to handle soft errors

Mellvik:
The boot sector would treat any error code from disk/floppy read as an error condition. However, old style MFM drives report soft (recovered) errors as errors which means that boot will fail if such 'errors' are encountered. This update fixes that.

While in the issue, there is a related issue with early Compaqs and 360k floppies: Every 12th block read returns and error (CF bit set) and error code C1 in %ah. There is, however, no 'real' error and a retry fixes it, which is good: There is no space for handling it in the boot sector code. So this note is for the record. 

In the picture below, the '*' normally denoting a read error during boot has been replaced with the error code from `%al` plus 0x20.

IMG_0913

The presence of the 'H' before the 't' in the output from Setup.S is possibly confusing since there is no HMA, but really a reminder that the 'H' is about what's in `bootopts`, not what's actually done in Setup. The text segment address tells that story anyway.

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -453,7 +453,7 @@ pc98_1b:
 #endif
 	pop %es
 	jnc dr_cont
-	cmp $0x11,%al	// accept soft errors (MFM drives)
+	cmp $0x11,%ah	// accept soft errors (MFM drives)
 	jz dr_cont
 	//mov %ah,%al	// Print error code, comment out the '*' putc
 	//add $0x20,%al	// below to make space
@@ -464,8 +464,8 @@ pc98_1b:
 	int $0x13
 #endif
 	// PATCH: failure trace
-	// Early Compaqs return 0xC1 every 12th block when reading 360k floppies
-	// First retry is fine
+	// Note: Early Compaqs return timeout every 12th block
+	// when reading 360k floppies using FAST_READ
 
 	mov $'*',%al
 	call _putc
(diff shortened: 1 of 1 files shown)


ghaerr:
> Every 12th block read returns and error (CF bit set)

Why? What is the root cause of this error, is it a BIOS bug or boot code? Has this been tested with FAST_READ turned off?

Having a regularly-occurring "error" exactly every 12 sectors seems quite suspicious, and more likely something to do with perhaps attempting multi-track reads or DDPT issues.

> error code C1 in %ah. There is, however, no 'real' error 

What is the Compaq documentation for error 0xC1? It seems to be a specific Compaq error. I don't understand why a BIOS would return an error if there is none.

Mellvik:
Hi @ghaerr, your curiosity is appreciated (and admittedly annoying sometimes). I had written this off as some Compaq BIOS anomaly and effectively archived it, since the same floppy on a different (although faster) XT class system also with a genuine 360k drive boots just fine.

Taking another round and the picture makes more sense. 
1) With single sector reads there are no errors
2) The error code isn't 0xC1 as I stated, it's 0x80 (I misread the code page table) - which means timeout (!)
3) The error is genuine, no data gets read

So, for some reason this particular model gets a floppy read timeout regularly when reading more than one sector per transaction. A reset fixes it, a simple retry would probably do the same and save some time, but there is no space for that. It's good as it is. And 'Compaq BIOS anomaly' is still the verdict.

Issue #209, opened 2025-11-25
Boot fails on 64M minix file systems


Mellvik:
A few years back - on ELKS - the max size of minix file systems was increased from 32M to 64M. That has worked fine since, probably rely used on ELKS, while on TLVC, 64M has been the standard for HD file systems since the fork.

There have been occasional boot failures (code 4, bad kernel signature) but too rare to chase.

Recently that has changed. For whatever reason, most likely file systems filling up with whatever debugging entails (such as big dummy files, 10 or 20 variants of the kernel, whatever...) plenty of space invites to laziness as far as deleting is concerned.

So, a new kernel is more likely to end up in the upper half of the 64M file system. That - as it turns out makes ut unbootable. The boot code uses an unsigned int to hold the block to read, and shifts it left before calling `disk_read`. With obvious consequences for kernel (and indirect map) blocks above 32k.

@ghaerr, this is the reason for the problems reported in https://github.com/Mellvik/TLVC/issues/205#issuecomment-3568136056 - so the unlimited bootopts implementation is in the clear (as far as this is concerned).

I'm looking at ways to fix this right now. The extremely tight space make changes tricky. 

This presumably also being a problem on ELKS, you may want to take a look, @ghaerr.

Mellvik:
> One scenario for a solution would be that the disk_read function take a block number, not a sector number, as the first argument. Removing those four left shifts should allow for a few more instructions. Then, in the disk_read ASM function, using a few instructions to compute the sector number into a pair of registers (the upper half would always be 0 or 1), would pass that to the modified CHS function to get CHS from 32-bit sector number.

Yes, that's probably the way to go unless we decide to redo the whole thing (not tempting). I wrote up a hack this morning - works on QEMU, to be tested on HW. Proof of concept, all kinds of non-sustainable tricks to get the code to fit. Needs some real smart thinking.

Enjoy traveling.

Mellvik:
> Geez, the rabbit holes that have opened up since we started the "easy" talking about just getting the NIC drivers cleaned up for ELKS, then long interrupt cleanup discussion, and now this!! When it rains, it pours, that's for sure!

It is indeed interesting. It seems that ELKS needs more hardware exposure, TLVC needs more exposure in general - more eyes and (wider) use. Anyway, important commonalities, mutually beneficial differences.

Every time we stumble into one of these rabbit holes, we come out with both systems improved. It's inspiring!

Thanks.

Mellvik:
[...]

PR #210, opened 2025-11-27
[boot] Fix for failing boot on large minix filesystems

Mellvik:
This is a first and still 'rough' fix for the boot problem discussed in https://github.com/Mellvik/TLVC/issues/209 - boot would fail on Minix filesystems larger than 32M if the kernel was located in the upper half of the FS.

The fix is simple - pass blocks instead of sectors to the boot program's underlying `disk_read` function. But the tight space makes it really hard to fit the changes in while still making sense.

This commit, which still has quite a bit of commented out experimental code from the process, has been tested on QEMU and physical HW. It has not yet been tested on FAT file systems, but the necessary changes to accommodate FAT have been shoehorned in. IOW - the code compiles and loads - with and without FAT boot support. With one byte to spare.

@ghaerr - your eyes and comments on this would be very helpful. 

While working on this, I popped back to the problem discussed a while back about the Compaq XT which timed out on every 12th block when booting from 360k floppy. This may be the reason: The boot program has the wrong idea about the sectors per track and fails every time it tries to read a block starting at sector 9 (then accessing non-existent sector 10), then succeeds when reverting to single sector reads. I've seen the problem on other XT systems too lately (including ny 8MHz clone), so it's not purely compaq related, but not had the time and/or energy to follow it. The root cause may be in some code related to the BIOS floppy parameter table I commented out a long time ago in `boot_sector.S`. Now even more of that code is gone to make space for the current batch of fixes. IOW, something that needs to be looked at, but apparently only affects booting from 360k floppies.

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -57,14 +57,12 @@
 // Note: this may clobber BX and DS, and reset SP!  Use it only upon an error
 // or just before handing over to /linux
 .macro RESTORE_DDPT
-#ifndef BOOT_FAT	/* Experimental */
 #ifdef FAST_READ
 	mov $-8,%sp
 	pop %bx
 	pop %ds
 	popw (%bx)
 	popw 2(%bx)
-#endif
 #endif
 .endm
 
@@ -126,7 +124,6 @@ entry1:
 	rep
 	movsw
 
-#ifndef BOOT_FAT	/* EXPERIMENTAL - no space for both for now */
 #ifdef FAST_READ
 	// Copy the current BIOS floppy parameter table
 	// so that we could modify it locally
@@ -159,7 +156,6 @@ entry1:
 	mov %cx,%ds	 // CX = 0
 	popw (%bx)
 	mov %es,2(%bx)
-#endif
 #endif
 
 	// Rebase CS DS to work in the 64K segment
@@ -203,9 +199,8 @@ _next1:
 	mov %ax,%dx
 	mov $payload,%cx
 	push %ss
-	mov $ERR_PAYLOAD,%bl
 	clc
-	call _disk_read
+	call disk_read_sec
 
 	call payload
 #else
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @Mellvik,

I'd say for your first pass, this looks well done! Definitely agree using carry flag (rather than sect_oflow var) is the way to go.

Did you have to remove the FAST_READ functionality in order to find space for this? Was that for MINIX also, or just FAT? IIRC the FAT boot sector is even tighter than MINIX. Approx how much code does the FAST_READ stuff add to the boot sector?

For cleanup, I'd suggest the following, since there's (too) many [_]disk_read[*] functions:
- Rename disk_read to disk_read_blk
- Rename disk_read_s to disk_read_sec
- Can \_disk_read be eliminated, and/or perhaps renamed as disk_read_sec (and the other disk_read_sec removed)? I see we need CLC but it's not clear whether having a separate BL error code just for payload is necessary, as tight as things are. The single new routine would load ERR_DISK_READ into BL (and perform CLC).
- Rename \__disk_read to "1:", and just use `jmp 1f` above it. This eliminates a confusing 4th name for disk reading.
- Remove old sect_oflow instructions that are already commented out

It looks like the FAT boot should work. Nice work! The biggest issue is I see is the loss of FAST_READ for floppy boots, this could be a lot slower. This could possibly be mitigated by using separate floppy and HD boot sectors, especially since this change is only required for HD boots (and in particular > 32MB drives).

bootblocks/boot_sect.S
```diff
@@ -310,19 +314,26 @@ drive_reset:
...
 
-_disk_read:
-	// BL = error code in case of failure
+	shl %ax	// turn start block into sector#, use CF later
+
+disk_read_sec:	// AX is sectors (CF must be clear!)
+
+	mov $ERR_DISK_READ,%bl // BL = error code in case of failure
+
+_disk_read:	// preset error code
```
ghaerr: How many bytes do you need to reincorporate FAST_READ in to the FAT boot? I think a few more bytes can be saved by removing ERR_PAYLOAD above on 206, and calling disk_read_sec directly (this also effectively removes the need for the confusingly-named \_disk_read). I can't see any reason why a separate ERR_PAYLOAD error would ever be useful.

PR #211, opened 2025-12-06
[net, ktcp] Important bug fixes and enhancements 

Mellvik:
This PR includes the following fixes and updates to `ktcp`:
- Fix for a problem appearing when a recipient acknowledges several packets in the same ACK - originally reported and subsequently discussed in [[inet] telnetd seems to hang in a "TCP Retransmit" situation](https://github.com/ghaerr/elks/issues/2487) on ELKS. This would cause the `inflight` counter, part of the SlowStart/CongestionAvoidance mechanism, to get out of sync and eventually hang the connection. More details below.
- The code includes a commented out but easily accessible mechanism to induce outgoing packet loss in ktcp - for future debugging (in `deveth.c`).
- Fix for a problem appearing when `ktcp` is running low on memory for retrans packets. More details below.
- Add the new `get_time()` mechanism to reliably get system time (jiffies) replacing direct access to the kernel `jiffies` variable in `ktcp`, as discussed in https://github.com/Mellvik/TLVC/pull/189#issuecomment-3523171604 and https://github.com/ghaerr/elks/issues/2456.
- More elaborate comments have been added to the code at several places to help understanding the code, the choices made and not the least those NOT made.
- Revert a former 'optimization' which sent an outgoing packet to the driver level before adding it to the retrans queue. It turns out (thank you @ghaerr) that metadata updates in `add_to_retrans` are required before sending.
- Some rearrangement of (still voluminous) debug code

The first commit includes some old and/or unused code for reference. This is deleted in the second commit.

<TL;DR>

The `inflight` counter problem was discovered more or less by accident (by @swausd) and the root cause turned out to be hardware. The bug was real enough though, and brought up some interesting questions. Like 'should so-called DUP ACKs, repeated ACKs of a previously sent packet' be ignored or reacted upon? DUP ACKs appear when a recipient either discovers a missing packet or gets impatient - a receive timeout. Both are rare. Losing a packet is extremely rare in today's switched networks - even between very fast systems. Being the slowest kid on the block, losing a packet from  `ktcp` is close to impossible.

Impatience (timeouts) is a different story. Two ktcp-based systems - say, real XT systems - are likely to see timeouts every once in a while, quite often on a busy system: A recipient times out, ACKs the previous packet again (DUP ACK), while most likely the recipient - being as slow as the sender - may have received the packet but hasn't noticed yet. Or the sender hasn't had time to send it yet.
In both cases - and this is the point - doing nothing (patience) is the simple way to handle the 'problem'. In the first case, the sender's retrans timer will kick in and the not-ACKed packet will the retransmitted. In the second case, the recipient will discover the 'missing' packet in its in-queue and be happy.

So what's the problem here? A simple bug. The `inflight` counter, which keeps track of the number of 'packets in transit', packets sent but not ACKed, wrongly assumed one ack per packet. Which amazingly often is the case. When a telnet connection to a `ktcp`-based systems sends a long listing (like from the `hd` command) to the connected 'terminal', the sender often gets ahead of the recipient (many small packets), but the recipient still ACKs every single packet. However, if a packet stays missing (i.e. lost), the following happens: The recipient keeps sending DUP ACKs until the sender finally (re)sends the missing packet. By then the recipient has received many, maybe 10s of packets ahead of the missing one, so it acknowledges all of them in one single ACK. Which means our `inflight` counter is suddenly far off, which eventually will screw up the connection.
Simple problem, simple fix, serious consequences.

BTW - the 'simple fix' is simple, but not perfect. The inflight counter will still be slightly off at times, but minimally so, and it will automagically resync thanks to the initial if-statement in the fix: 
```
            if (cb->inflight == 1 || cb->send_una == cb->send_nxt)
                cb->inflight = 0;       /* all outstanding packets ACKed */
```
The 100% alternative would be to run through the retrans-list and account for packets with sequence numbers lower than the last ACK. I tried that (code included in the first commit), but even that turns into a nightmare because on a fast system, we would end up scanning the retrans buffer several times between its regular updates and thus counting the same packets more than once. In order to get this right we would have to yank every counted packet from the buffer, which would be fine except we don't really want to do this kind of work here, where fast processing of an ACK packet is the focus. It's the task of a timed retrans-loop - as it should.

Then the obvious question: Should `ktcp` do things differently, like let the DUP ACK trigger a retransmission? That would be technically simple - and fast, but there are two reasons not to do it: 1) Lost or dropped packet are extremely rare so the delay introduced by doing nothing is not really a problem. And 2) - acting immediately on a DUP ACK would disturb the most common case, the slow systems situation described above: Trigger unnecessary retransmissions, which is detrimental to efficiency. So if we were to do something, it would have to be on the 2nd or 3rd DUP ACK, which in turn would mean a per-connection counter, more complexity etc etc. Not worth it.

The 'out of retrans memory' fix is just a matter of deleting code that most likely was put in there when `ktcp` was very different than today. Every time ktcp sends a tcp packet it saves a copy and puts it into the retrans list. Memory is allocated on the fly and freed when the packet has been acked or timed out in the buffer - or the connection closed. If there isn't enough space in the retrans buffer, ktcp will hold back and tell the sending application to come back later, typically in 100ms. This way, overflowing or overextending the retrans buffer is impossible. It will grow slightly over the limit (set in `config.h`) some times, but minimally so. 

The point is that checking for overflow elsewhere is superfluous and useless because there is nothing to do. The sender(s) have already been put on hold as described before. In this particular (now deleted) case, the entire retransbuffer was purged. Admittedly, no one would notice unless a retransmission was required on an active connection. Which is what happened as I was heavy-testing the 'inflight-fix' above with artificially induced packet loss.
A bug, OK, but serious? Depends on the setting. When slow systems are communication there will be occasional retransmissions because of timeouts - without any real packet loss, and the connections would hang if the retransbuffer got slightly over the limit. Hang with a not very informative (sensible) message.

</TL;DR>

Why this long rant about something seemingly innocuous? It's educating. To me as I have to think everything through again, and for you who actually skipped the TL;DR temptation at the beginning.

### DIFF
--- a/tlvccmd/ktcp/tcp.c
+++ b/tlvccmd/ktcp/tcp.c
@@ -295,7 +295,6 @@ static void tcp_established(struct iptcp_s *iptcp, struct tcpcb_s *cb)
     if (h->flags & TF_ACK) {		/* update unacked */
 	acknum = ntohl(h->acknum);
 	if (SEQ_LT(cb->send_una, acknum)) {
-	    //long una = cb->send_una;
 	    cb->send_una = acknum;
 	    if (cb->cwnd <= cb->ssthresh && !cb->retrans_act)
 		cb->cwnd++; 		/* adjust congestion win */
@@ -304,16 +303,13 @@ static void tcp_established(struct iptcp_s *iptcp, struct tcpcb_s *cb)
 	    else {	/* We're more than 1 packet ahead of the recipient - which is normal.
 			 * The inflight counter may occasionally get slightly 
 			 * out of sync but will be autocorrected eventually. */
-		//if (cb->inflight)
-			//adj_outstanding(cb);
-		//printf("una %lu, ack %lu, inflight %d\n", una, cb->send_una, cb->inflight);
 		cb->inflight--;
 	    }
 	    //write(1,"A",1);
 	} /* An 'else clause' here would catch DUP ACKs */
 	  /* Keep in mind that a DUP ACK rarely means a lost packet nowadays, but rather
 	   * a delay that will rectify itself if we don't mess with it. */
-	  /* FIXME: We may still want to adjust the slowStart/CongestionAvoidance parameters
+	  /* TODO: We may still want to adjust the slowStart/CongestionAvoidance parameters
 	   * when this happens even if it's rare */
     }
 
--- a/tlvccmd/ktcp/tcp.h
+++ b/tlvccmd/ktcp/tcp.h
@@ -212,7 +212,6 @@ struct tcpcb_list_s *tcpcb_new(int);
 struct tcpcb_list_s *tcpcb_find(__u32, __u16, __u16);
 struct tcpcb_list_s *tcpcb_find_by_sock(void *);
 void tcpcb_update_sstimer(void);
-void adj_outstanding(struct tcpcb_s *);
 
 __u16 tcp_chksum(struct iptcp_s *);
 __u16 tcp_chksumraw(struct tcphdr_s *, __u32, __u32, __u16);
--- a/tlvccmd/ktcp/tcp_output.c
+++ b/tlvccmd/ktcp/tcp_output.c
@@ -133,7 +133,6 @@ rmv_from_retrans(struct tcp_retrans_list_s *n)
     struct tcp_retrans_list_s *next = n->next;
 
     tcp_timeruse--;
-    //n->cb->retrns_cnt--;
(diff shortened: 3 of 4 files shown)


ghaerr:
> The 'out of retrans memory' fix is just a matter of deleting code that most likely was put in there when ktcp was very different than today.

Yes, I added that code in the early days to prevent ktcp from running out of memory. Glad to hear it's no longer needed.

PR #213, opened 2025-12-17
[net, kernel] Another update of the wd/smc NIC driver

Mellvik:
This update is characterized by small changes that took a lot of effort. With 4 different NICs and 2 categories of systems (actually 3 were involved), there are countless combinations to test.

The result is first of all improved probing and auto configuration of the NIC variants. And secondly, a more reliable system. A summary:

- Extensive testing verifies that there is no reason to distinguish between 8bit and 16bit cards and buses when it comes to shared memory transfers. Word size memory moves are fine in all cases.
- A check for bus size is not useful. If the card is a 8003, we set the 8bit flag, otherwise we don't. Further, while the 16 bit cards may be able to tell with bus with, they don't sense whether the system uses the upper half of the bus or not. An XT class motherboard in a 16 bit AT type backplane, will be reported as a 16 bit system. Etc.
- That said, a warning may be appropriate at boot time when a 8003 (8bit) card is found in a 16bit ISA machine (arch_cpu > 5) because if combined with other shared memory ISA cards, the results are unpredictable. On recent (!) day systems with advanced BIOSes (shadow memory, UMB helpers an more), the 8003 cards may not work at all, regardless. On older ATs and clones, it works just fine. (more below)
- Comments in the intro has been expanded (again) to save findings that took a lot of effort to collect.
- The driver now correctly detects the 4 main card types (SMC and WD) and sets the buffer size correctly. Some 'half-compatible' clones are not supported. Buffer size may still be overridden in bootopts.
- The boot identification message has been cleaned up.

The problem with 16bit vs 8bit shared memory on the ISA bus cost a lot of effort - until I finally figured out what was going on. And instead of going through the details, this source is covering it all:
https://forum.vcfed.org/index.php?threads/wd8013ebt-xenix-what-jumper-settings-are-recommended.43789/#post-532793

### DIFF
--- a/tlvc/arch/i86/drivers/net/wd.c
+++ b/tlvc/arch/i86/drivers/net/wd.c
@@ -304,8 +304,12 @@ static int INITPROC wd_probe(void) {
 	/* The BUS SIZE detection is not completely reliable as the backplane may be 16bit
 	 * while the system is 8bit - or vise versa. Instead of adding more detection
 	 * code, the user should use the flags for borderline cases. */
+
 	/* NOTE: If the 8013 is used in an 8 bit system (i.e. 8 bit mode), buffer
-	 * size MUST be reduced to 8k. This does not apply to the 8x16 family. */
+	 * size MUST be reduced to 8k. This does not apply to the 8x16 family. 
+	 * They should be left alone, they just work as is in 8bit ISA systems. 
+	 * In fact, forced 8bit mode cause them to fail, and 16k buffer 
+	 * (8216) works fine. */
 
 	outb(tmp ^ 0x01, net_port+1 );		/* attempt to clear 16bit bit */
 	if (((type = (inb(net_port+1) & 0x01)) == 0x01)	/* A 16 bit card */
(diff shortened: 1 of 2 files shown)


Mellvik:
The most recent commit was supposed to just contain some formatting edits. Then the bugs popped up.

Like - while researching the details of the NIC registers, an old NetBSD source file revealed that we have been running the wd/smc NICs with 8 bit memory accesses since forever.

Summary:
- Formatting changes
- Renamed the cryptic 'asic_reg5' to the more commonly used (and understandable) Lan Adapter Address Register, LAAR, full name `WD_LAAR`. There is little magic about this register. The important bits in this register are named `LAAR_M16EN` and `LAAR_L16EN`.
- Fixed the mentioned bug that has prevented the NIC to use 16 bit memory accesses since forever. The speed improvement is visible if marginal (in the order of 4-5%)
- The wd8013 now works correctly in 8 bit bus systems w/o special configuration. The 8013 (but not the smc8x16 family) requires the buffer size to be reduced to 8k when used in an 8 bit bus system. Formerly this had to be done manually via `bootopts`. Now it's automatic.
- Manual buffer settings via `bootopts` now applies to 8bit mode too.

PR #215, opened 2025-12-19
[net] Add 'Interrupt Process' and output toggle to telnet + misc cleanups

Mellvik:
A collection of minor 'maintenance' edits plus and overdue edit of  function and visual 'edit' of `telnet`:
- Now supports the `Interrupt Process` telnet protocol option, used if the user tries to abort (^C) a long running output from the server side - which usually takes along time thanks to large buffers in the channel. Thanks for the inspiration @ghaerr!
- Also support the standard Unix/Linux output toggle character ^O to turn long running output off and on
- Plus a lot of simple formatting updates in the code

Really two completely unrelated commits that accidentally ended up in the same PR, and I was just too lazy/tired to fix it.

### DIFF
--- a/tlvccmd/inet/telnet/telnet.c
+++ b/tlvccmd/inet/telnet/telnet.c
@@ -21,10 +21,12 @@
 #include <netinet/in.h>
 #include <arpa/inet.h>
 
-//#define DEBUG 1
+#define debug(...)
+//#define debug     printf
 //#define RAWTELNET	/* set in telnet and telnetd for raw telnet without IAC*/
 
-#define ESCAPE      (']'&0x1f)  /* = ^] escape session, effectively terminate */
+#define ESCAPE      (']'&0x1f)  /* = ^] escape session. Since we don't have a
+				 * command mode, escape will terminate the session */
 #define BUFSIZE		1500
 
 /* telnet protocol */
@@ -54,6 +56,8 @@
 
 #define FALSE	0
 #define TRUE	(!(FALSE))
+#define CTRL_C	3
+#define CTRL_O	15
 
 #ifdef __linux__
 int DO_echo= TRUE;
@@ -77,7 +81,6 @@ struct termios def_termios;
 int escape;
 int discard;
 
-static int writeall (int fd, char *buffer, int buf_size);
 static int process_opt (char *bp, int count);
 
 void finish()
@@ -91,30 +94,41 @@ void finish()
 static void read_keyboard(void)
 {
 	int count;
-	char buffer[BUFSIZE];
+	unsigned char buffer[BUFSIZE];
 
 		count = read(0, buffer, sizeof(buffer));
-		//fprintf(stderr, "got %d (%x)\r\n", count, *buffer);
+		//fprintf(stderr, "got %d (0x%x)\n", count, *buffer);
 		if (count <= 0 || *buffer == escape) { 
-			char *msg = NULL;
-			if (count <= 0) msg = "read error, ";
-			fprintf(stderr, "\r\ntelnet: %ssession terminated\r\n", msg);
+			fprintf(stderr, "\nSession terminated\n");
 			finish();
 			return;
 		}
-		if (*buffer == 3) {		/* ^c = abort output */
+		if (*buffer == CTRL_C) {	/* ^C = abort output */
+		    if (discard < 3) {		/* ignore if we already sent an IAC_IP */
+			discard = 3;		/* discard whatever is en route */
+#define USE_IAC_IP
+#ifdef USE_IAC_IP				/* Some telnet servers behave better with ^C
+						 * than with IAC_IP (!). */
 			buffer[0] = IAC;
 			buffer[1] = IAC_IP;	/* send Interrupt Process */
 			count = 2;
-			//fprintf(stderr, "sending IAC_IP\r\n");
-			discard = 1;		/* discard whatever is en route */
+			//fprintf(stderr, "sending IAC_IP\n");
+#endif
+		    }
 		}	
-		if (*buffer == 15)		/* ^O - flush output toggle */
+		if (*buffer == CTRL_O && !(discard&2)) {	/* ^O - flush output toggle */
 			discard ^= 1;
-
+			buffer[0] = '^';
+			buffer[1] = 'O';
+			write(1, buffer, 2);
+			return;			/* Don't pass on to server */
+		}
+		if (discard == 1) {		/* let any char typed release ^O */
+			discard = 0;
+			return;
+		}
 		count = write(tcp_fd, buffer, count);
-		if (count < 0)
-		{
+		if (count < 0) {
 			perror("Connection closed");
 			finish();
 		}
@@ -129,36 +143,36 @@ static void read_network(void)
 		count = read(tcp_fd, buffer, sizeof(buffer));
 
 		if (count <= 0) {
-			printf("\r\nConnection closed\r\n");
+			printf("\nConnection closed\n");
 			finish();
 		}
-		if (discard)
-			return;
 #ifdef RAWTELNET
 		write(1, buffer, count);
 #else
 		bp = buffer;
+
 		do {
+			
 			iacptr = memchr(bp, IAC, count);
+			//fprintf(stderr, "read: %d discard %d (%x)\n", count, discard, iacptr);
 			if (!iacptr) {
-				write(1, bp, count);
-				count = 0;
-				return;
+				if (!discard) write(1, bp, count);
+				break;
 			}
-			if (iacptr && iacptr>bp) {
-				write(1, bp, iacptr-bp);
+			if (iacptr > bp) {
+				if (!discard) write(1, bp, iacptr-bp);
 				count -= (iacptr-bp);
 				bp = iacptr;
 				continue;
 			}
-			if (iacptr) {
-				optsize = process_opt(bp, count);
-				if (optsize < 0)
-					return;
-				bp += optsize;
-				count -= optsize;
-			}
+			optsize = process_opt(bp, count);
+			if (optsize < 0)
+				break;
+			bp += optsize;
+			count -= optsize;
 		} while (count);
+
+		return;
 #endif
 }
 
@@ -234,36 +248,43 @@ int main(int argc, char **argv)
 	printf("Connected\n");
 	printf("Escape character is '^%c'.\n", escape + '@');
 
-#ifdef RAWTELNET
-	{
 	struct termios termios;
 	tcgetattr(0, &termios);
+	termios.c_oflag |= (OPOST | ONLCR);
+	termios.c_lflag &= ~ISIG;       /* ISIG off to disable ^N/^O/^P */
+#ifdef RAWTELNET
 	termios.c_iflag &= ~(ICRNL|IGNCR|INLCR|IXON|IXOFF);
-	termios.c_oflag &= ~(OPOST);
-	termios.c_lflag &= ~(ECHO|ECHONL|ICANON|ISIG);
-	tcsetattr(0, TCSANOW, &termios);
-	}
+	termios.c_lflag &= ~(ECHO|ECHONL|ICANON);
 #endif
+	tcsetattr(0, TCSANOW, &termios);
 	nonblock = 1;
 	ioctl(0, FIONBIO, &nonblock);
 
 	for (;;) {
 		int n;
 		fd_set fdset;
-		struct timeval tv;
+		struct timeval tv, *ptv;
 
 		FD_ZERO(&fdset);
 		FD_SET(0, &fdset);
 		FD_SET(tcp_fd, &fdset);
 
 		tv.tv_sec = 0;
-        	tv.tv_usec = 10000;
+		if (discard == 3) {	/* If telnetd at the other end never 'acks' an IAC_IP
+					 * with a DataMark, a 3 second timeout will
+					 * release the discard - unless the user has sent a second
+					 * ^C before that. */
+		    ptv = &tv;
+        	    tv.tv_usec = 3000000;
+		} else
+        	    ptv = NULL;
 
-        	n = select(tcp_fd + 1, &fdset, NULL, NULL, &tv);
+        	n = select(tcp_fd + 1, &fdset, NULL, NULL, ptv);
         	if (n == 0) {
+		    //fprintf(stderr, "timeout!\n");
 		    discard = 0;
-		    continue;		/* TODO: The user doesn't want to see the last buffer,
-					 * purge it, leaving just the last line. */
+		    write(tcp_fd, "\r", 1);
+		    continue;	
 		}
 		if (n < 0) {
 		    perror("select");
@@ -283,13 +304,26 @@ int main(int argc, char **argv)
 	if (offset < count) { (var) = bp[offset++]; } \
 	else { if (read(tcp_fd, (char *)&(var), 1) != 1) printf("TELNET BAD READ2\n"); exit(1); }
 
-static void do_option(int optsrt)
+static int writeall(int fd, char *buffer, int buf_size)
 {
 	int result;
+
+	while (buf_size)
+	{
+		result = write(fd, buffer, buf_size);
+		if (result <= 0)
+			return -1;
+		buffer += result;
+		buf_size -= result;
+	}
+	return 0;
+}
+
+static void do_option(int optsrt)
+{
 	unsigned char reply[3];
 
-	switch (optsrt)
-	{
+	switch (optsrt) {
 	case OPT_TERMTYPE:
 		if (WILL_terminal_type)
 			return;
@@ -313,18 +347,14 @@ static void do_option(int optsrt)
 		reply[2] = optsrt;
 		break;
 	}
-	result = writeall(tcp_fd, (char *)reply, 3);
-	if (result < 0)
-		perror("write");
+	writeall(tcp_fd, (char *)reply, 3);
 }
 
 static void will_option(int optsrt)
 {
-	int result;
 	unsigned char reply[3];
 
-	switch (optsrt)
-	{
+	switch (optsrt) {
 	case OPT_ECHO:
 		if (DO_echo)
 			break;
@@ -336,19 +366,15 @@ static void will_option(int optsrt)
 			struct termios termios;
 			tcgetattr(0, &termios);
 			termios.c_iflag &= ~(ICRNL|IGNCR|INLCR|IXON|IXOFF);
-			/*termios.c_oflag &= ~(OPOST);*/	/* leave OPOST|ONLCR on*/
-			termios.c_lflag &= ~(ECHO|ECHONL|ICANON);	/* leave ISIG on for ^P*/
-			termios.c_cc[VINTR] = 0;	/* turn ^C off*/
-			termios.c_cc[VSUSP] = 0;	/* turn ^Z off*/
+			termios.c_oflag |= (OPOST | ONLCR);
+			termios.c_lflag &= ~(ECHO|ECHONL|ICANON|ISIG);	/* ISIG off for ^P*/
 			tcsetattr(0, TCSANOW, &termios);
 			DO_echo = TRUE;
 			reply[0] = IAC;
 			reply[1] = IAC_DO;
 			reply[2] = optsrt;
 		}
-		result = writeall(tcp_fd, (char *)reply, 3);
-		if (result < 0)
-			perror("write");
+		writeall(tcp_fd, (char *)reply, 3);
 		break;
 
 	case OPT_SUPP_GA:
@@ -364,41 +390,21 @@ static void will_option(int optsrt)
 			reply[1] = IAC_DO;
 			reply[2] = optsrt;
 		}
-		result = writeall(tcp_fd, (char *)reply, 3);
-		if (result < 0)
-			perror("write");
+		writeall(tcp_fd, (char *)reply, 3);
 		break;
 
 	default:
 		reply[0] = IAC;
 		reply[1] = IAC_DONT;
 		reply[2] = optsrt;
-		result = writeall(tcp_fd, (char *)reply, 3);
-		if (result < 0)
-			perror("write");
+		writeall(tcp_fd, (char *)reply, 3);
 		break;
 	}
 }
 
-static int writeall(int fd, char *buffer, int buf_size)
-{
-	int result;
-
-	while (buf_size)
-	{
-		result = write(fd, buffer, buf_size);
-		if (result <= 0)
-			return -1;
-		buffer += result;
-		buf_size -= result;
-	}
-	return 0;
-}
-
 static void dont_option(int optsrt)
 {
-	switch (optsrt)
-	{
+	switch (optsrt) {
 	default:
 		break;
 	}
@@ -406,8 +412,7 @@ static void dont_option(int optsrt)
 
 static void wont_option(int optsrt)
 {
-	switch (optsrt)
-	{
+	switch (optsrt) {
 	default:
 		break;
 	}
@@ -417,7 +422,7 @@ static int sb_termtype(char *bp, int count)
 {
 	unsigned char command, iac, optsrt;
 	unsigned char buffer[4];
-	int offset, result, ret_value;
+	int offset, result;
 
 	offset= 0;
 	next_char(command);
@@ -426,32 +431,26 @@ static int sb_termtype(char *bp, int count)
 		buffer[1] = IAC_SB;
 		buffer[2] = OPT_TERMTYPE;
 		buffer[3] = TERMTYPE_IS;
-		result = writeall(tcp_fd, (char *)buffer,4);
-		if (result < 0) {
-			ret_value = result;
-			goto ret;
-		}
+		result = writeall(tcp_fd, (char *)buffer, 4);
+		if (result < 0)
+			return result;
 		count = strlen(term_env);
 		if (!count) {
 			term_env = "unknown";
 			count= strlen(term_env);
 		}
 		result = writeall(tcp_fd, term_env, count);
-		if (result < 0) {
-			ret_value = result;
-			goto ret;
-		}
+		if (result < 0)
+			return result;
 		buffer[0]= IAC;
 		buffer[1]= IAC_SE;
-		result = writeall(tcp_fd, (char *)buffer,2);
-		if (result < 0) {
-			ret_value = result;
-			goto ret;			
-		}
-
+		result = writeall(tcp_fd, (char *)buffer, 2);
+		if (result < 0)
+			return result;
 	} else {
-		fprintf(stderr, "got an unknown command (skipping)\r\n");
+		debug("got an unknown command (skipping)\n");
 	}
+
 	for (;;) {
 		next_char(iac);
 		if (iac != IAC)
@@ -460,13 +459,11 @@ static int sb_termtype(char *bp, int count)
 		if (optsrt == IAC)
 			continue;
 		if (optsrt != IAC_SE) {
-		/*	fprintf(stderr, "got IAC %d\r\n", optsrt);*/
+			debug("got IAC %d\n", optsrt);
 		}
 		break;
 	}	
-	ret_value = offset;
-ret:
-	return ret_value;
+	return offset;
 }
 
 
@@ -478,56 +475,54 @@ static int process_opt(char *bp, int count)
 	offset = 0;
 	next_char(iac);
 	next_char(command);
-	switch(command)
-	{
+	switch(command) {
 	case IAC_NOP:
 		break;
 	case IAC_DataMark:
-		//fprintf(stderr, "got a DataMark\r\n");
+		discard = 0;
+		debug("got DataMark\n");
 		break;
 	case IAC_BRK:
-		fprintf(stderr, "got a BRK\r\n");
+		debug("got BRK\n");
 		break;
 	case IAC_IP:
-		fprintf(stderr, "got a IP\r\n");
+		debug("got IP\n");
 		break;
 	case IAC_AO:
-		fprintf(stderr, "got a AO\r\n");
+		debug("got AO\n");
 		break;
 	case IAC_AYT:
-		fprintf(stderr, "got a AYT\r\n");
+		debug("got AYT\n");
 		break;
 	case IAC_EC:
-		fprintf(stderr, "got a EC\r\n");
+		debug("got EC\n");
 		break;
 	case IAC_EL:
-		fprintf(stderr, "got a EL\r\n");
+		debug("got EL\n");
 		break;
 	case IAC_GA:
-		fprintf(stderr, "got a GA\r\n");
+		debug("got GA\n");
 		break;
 	case IAC_SB:		/* subnegotiation */
 		next_char(sb_command);
-		switch (sb_command)
-		{
+		switch (sb_command) {
 		case OPT_TERMTYPE:
 			result= sb_termtype(bp+offset, count-offset);
-			if (result<0)
+			if (result < 0)
 				return result;
 			else
 				return result+offset;
 		default:
-			fprintf(stderr, "got an unknown SB (skiping)\r\n");
-			for (;;)
-			{
+			debug("got unknown SB (skipping)\n");
+			for (;;) {
 				next_char(iac);
 				if (iac != IAC)
 					continue;
 				next_char(optsrt);
 				if (optsrt == IAC)
 					continue;
-if (optsrt != IAC_SE)
-	fprintf(stderr, "got IAC %d\r\n", optsrt);
+				if (optsrt != IAC_SE)
+					debug("got IAC %d\n", optsrt);
 				break;
 			}
 		}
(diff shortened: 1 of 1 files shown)


ghaerr:
Nice idea with ^O. I added that to ELKS in https://github.com/ghaerr/elks/pull/2543 but also added proper handling of ^O/^N/^P kernel debug commands when configured, as they otherwise take precedence.

I didn't notice any difference at all sending IAC IP instead of ^C to macOS as far as speed in halting the remote continuing to send data. And since ELKS telnetd doesn't handle IAC IP and requires ^C to stop output, I left sending ^C in.

Note also your fix sends ^O on the net, which is improper.

ghaerr:
> This was supposed to be a quick-fix, @ghaerr. Yet here I am, many days later

Well, for me it was only a quick way of dealing with the otherwise-unstoppable output from a very fast host that queued packets far faster than ELKS can process. For me, I wasn't trying to have telnet somehow clean the output up perfectly in all cases. In fact as a result of debugging an unrelated problem with `sl` via a telnet session, I've posted https://github.com/ghaerr/elks/pull/2550 which eliminates ^C discarding output at all, since 1) that is not its purpose and is non-standard behaviour in any telnet, and 2) can make the system appear hung to users when programs ignore SIGINT. Instead, only ^O is used to discard output.

> MacOS telnet does NOT turn off ^O in the terminal driver

Realiistically, this is probably where ^O processing actually belongs - in the TTY driver, not a telnet program.

> In order for the ^O (discard) function to work with our telnet client, the timeout must be off. 

> With the timeout only applying to ^C, a reasonable timeout is 40ms

That's all good to know, but IMO tuning a telnet program for various systems seems to me an approach of "perfection" that will always have the door open for the next system that doesn't quite work, and frankly doens't belong in telnet, now that you've brought up the macOS TTY behaviour. (And thank you for your testing, as I never would have thought of that!)

The ELKS telnet currently has ony ^O support built in, with no fancy timeout handling, althought it does seem to work well with very fast hosts. A user (likely only me?) now has to type ^C^O in order to stop fast streaming output. I've made a note to look into moving ^O support into the TTY driver directly, thus working to discard all unwanted output without special per-application handling.

Mellvik:
It was an admirable goal @ghaerr, one to which I immediately subscribed (obviously). The problem is real enough: When our telnet, running on a presumably slow system, is connected to any modern-day fast machine, there is going to be somewhere close to 64k bytes in the pipe when the ^C hits. That's a couple of minutes worth of output from `hd`- for example. So suppressing the unwanted output (saving time and noise) while ensuring that the user has a chance to understand what's (not) happening, is a significant usability improvement.

<TL;DR>Turns out telnet(d) has what we need to achieve reliable discard after INTR already - builtin. </TL;DR>

That said, I came to the same conclusion as you when I found that whatever timeout value I used, it would not cover all situations. The whole thing just got too complicated. Then I got the idea that packet size could be a useful metric to detect the end of whatever output was lingering on the connection at the time of ^C. Presumably, the server, when sending a long whatever text output, will use large packets except the last one. So I created a simple state machine that immediately worked with Linux as the host. On the Mac (brew telnetd) however, the size assumption turned out to be incorrect. Scheduling or whatever cause the Mac to send highly variable packet sizes. In particular, the ^C itself will cause an interruption of the network flow and cause vary variable sized packets for a little while. The state machine got more complicated and the timeout was back (although with an 'uncritical' value).

While debugging this I found something really interesting: The Mac `telnetd` 'acks' the IAC_IP command with a IAC `dataMark` 'command' which becomes part of the flow and arrives at the source immediately after the data from the interrupted process has ended. Wow. I had been experimenting with IAC_AYT and other IAC commands to find something like that and here I got it 'for free'. So the AYT (Are You There) idea wasn't so bad to begin with. If sent immediately after the ^C, we could just wait for the reply and know that when it came, the transfer had ended. Except - as it turns out - not all `telnets` implement it, it's optional. Linux `telnetd` (old raspian version) does, the Mac version does not. Also, AYT has the disadvantage that the replay comes back as implementation dependent regular text, so it would have to be searched for. 

Back to the IAC_IP ack (`dataMark`), it's actually perfect. What is returned is an IAC command and a byte (242), handled by the existing do_option code in our `telnet` client.  What we have a simple and safe way to detect the end of output when interrupting a remote process. No timeouts, no tricks. EXCEPT the Linux telnetdaemon does not support it. Maybe not others either, these are the only two I'm testing against. 

Still, this is the solution. It works for the Mac, and as mentioned above, a Linux peer sends very predictable packets. Big ones except the last. No timeouts, just sizes. So I implement both. But again there is a complication: In order to keep the simplicity, it's necessary to know whether the peer is one or the other type.

When researching this, the ultimate solution revealed itself accidentally. I was testing some telnet options on the Linux system and for convenience connected it to itself (localhost). Enabled 'options debugging' and (in the flow) hit ^C. 'RCVD IAC DMARK' replied the system. What? The Linux `telnetd` does send the dataMark reply to the IP command? Why did I miss it? Turns out I didn't. It does send the dataMark back ONLY if receiving a ^C directly, not IAC_IP. No sh..! Actually a bug in Linux `telnetd`, even possibly fixed in version newer than the one I'm using. The Mac `telnetd` doesn't care, it replies with the DataMark in both cases. 

So I'm removing the ^C til IAC_IP conversion in our `telnet` client and we're fine (you're fine already IIRC). Do you believe it? Well, there is a final caveat. While this works completely cleanly with the Mac as the peer, the Linux peer will somehow send the DataMark back, then another 4143 bytes of data. Which is annoying but not a dealbreaker. It's also puzzling. What may be happening is that the Linux `telnetd` - per telnet specs - uses the URG flag with the IAC DataMark return which then ends up ahead of the final 'window-fill' to our client. Even more likely - since I get the same behaviour from the TLVC telnetd - is that it's a scheduling 'issue'. `telnetd` is retuning the DataMark immediately upon reception, before the underlying process has received the INTR signal. So before the process has been terminated it has sent more data into the pipe, which arrives after the DataMark. I did some testing on this in the TLVC `telnetd`, and even with a `sleep(1)` before returning the DataMark, there is still data after it. 

I need to test this - out of curiosity - on more (and newer) Linux hosts to figure this out. 

Regardless, this is the solution I was looking for. It works really well - with the caveats mentioned, using minimal code. And - with this minimal functionality implemented in the TLVC `telnetd` - we have a complete working solution 'locally' (TLVC/ELKS) too.

ghaerr:
I'm curious - what happens when you telnet to macOS from ELKS and type ^C when running `sl`, with your new mods?

Mellvik:
Hmmm, what is `sl`?

Mellvik:
As it turns out, recent versions of Linux netutils `telnetd` handles ^C/IAC_IP and DataMark combination correctly, so our `telnet` client now (again) sends IAC_IP when getting a ^C from the user. In order to handle the odd case when talking to a buggy telnetd, a 3 second network read timeout has been kept. Incidentally, if talking to a telnetd that does not respond to IAC_IP with a DataMark, a second ^C (which the user is quite likely to hit when nothing seems to be happening) will also release the situation.

The ^O (discard) handing has been changed so that when set (^O has been sent), any key will reenable output, which echoes ^O handling on most unix like systems.

As to the `sl` question @ghaerr, I'm assuming this is about ANSI commands and terminal states. Randomly discarding output from programs manipulating terminal state like color, scroll regions or whatever, will leave the terminal in an undefined state. It doesn't matter whether it's the telnet client or the tty driver (^O case) or a rudely killed process or some other 'filter' doing it, the telnet client's discarding of output does not introduce anything new in that regard. It's a rare occurrence (typically with colorized output from `ls`, `xxd`, `grep` and various `git`-commands) which - if deemed a problem - can be fixed by the user adding a reset sequence to the shell prompt.

ghaerr:
Nice analysis of telnetd behaviour. Sounds like an intended purpose of the DM may be to allow a client to learn precisely when the ^C was received, in terms of the streaming output (to implement discarding or other etc), but as always, it only works when both client and server implement the "standard" whatever that may be in this case. For those servers that don't handle an IAC_IP correctly (or at all), things could get complicated when the client makes assumptions. This was my concern with `sl` (see below).

> if talking to a telnetd that does not respond to IAC_IP with a DataMark, a second ^C ...

Sounds a bit complicated. Does the second ^C send another IAC_IP also?

> The ^O (discard) handing has been changed so that when set (^O has been sent), any key will reenable output, which echoes ^O handling on most unix like systems.

Sounds like a good idea (and probably belongs in the TTY driver)?

> As to the sl question, I'm assuming this is about ANSI commands

No, `sl` happens to be a program that ignores ^C but also continues to stream output, so I was wondering how this situation  appears to an unsuspecting user, especially when the server telnetd does not implement "standard ? / special" handling. 

When I first tested my own more complex implementation of ^C on ELKS telnet, running `sl` on an unmodified ELKS telnetd and typing ^C appeared to hang the system to the user, since `sl` ignored ^C, continued to stream output, and nothing appeared on the screen. It happened at a time when I was debugging something else entirely, and I had forgotten (happens a lot these days) about this "special" behaviour programmed into telnet, and took me way too long to (re)figure out what was wrong. So I disabled it in telnet since there are times when I want to see exactly what is happening, rather than massaged data or potential system hangs that aren't accurate. I mentioned this earlier in this thread but I think it must have got lost in the shuffle.

> In order to handle the odd case when talking to a buggy telnetd, a 3 second network read timeout has been kept.

Given there seems to be no perfection and my desire to not bake assumptions in to remote ^C behaviour of both telnetd and its clients, I'll likely stick to the very simple implementation of ^C just sending ^C, and ^O toggling discarding output. In this wasy, "untrained" users won't possibly think the system has hung in certain circumstances (mostly because it tricked me!).

Thanks for your discussion and research!

Mellvik:
> 
> > if talking to a telnetd that does not respond to IAC_IP with a DataMark, a second ^C ...
> 
> Sounds a bit complicated. Does the second ^C send another IAC_IP also?

This is simply a consequence of the implementation. The telnet client doesn't send a second IAC_IP if one is already 'active' (receipt of a DM is pending, output supressed), just passes the ^C along.

> 
> > The ^O (discard) handing has been changed so that when set (^O has been sent), any key will reenable output, which echoes ^O handling on most unix like systems.
> 
> Sounds like a good idea (and probably belongs in the TTY driver)?

It does indeed belong in the tty driver. Whenever that support is in place, the ^O will never reach the telnet client so no changes required.

> `sl` happens to be a program that ignores ^C but also continues to stream output, so I was wondering how this situation  appears to an unsuspecting user, especially when the server telnetd does not implement "standard ? / special" handling. 

Interesting case. What is sl? I associate it only with the old 'reverse ls' prank.

> 
> ... It happened at a time when I was debugging something else entirely, and I had forgotten (happens a lot these days) about this "special" behaviour programmed into telnet, and took me way too long to (re)figure out what was wrong. 

Wow does that sound familiar... 

> So I disabled it in telnet since there are times when I want to see exactly what is happening, rather than massaged data or potential system hangs that aren't accurate. I mentioned this earlier in this thread but I think it must have got lost in the shuffle.

I understand. There are of course other ways to do this, such as a debug option to telnet.

> 
> > In order to handle the odd case when talking to a buggy telnetd, a 3 second network read timeout has been kept.
> 
> 
> Given there seems to be no perfection and my desire to not bake assumptions in to remote ^C behaviour of both telnetd and its clients, I'll likely stick to the very simple implementation of ^C just sending ^C, and ^O toggling discarding output. In this wasy, "untrained" users won't possibly think the system has hung in certain circumstances (mostly because it tricked me!).
> 

To me the benefits far outweigh the drawbacks. Different usage patterns etc. I believe the danger of the system appearing hung is taken care of by the adjustments in the last commit, including an extra CR to get the shell prompt or whatever, depending on the app running.

Mellvik:
Long story short, I ended up reading the `telnet` source from the GNU inetutils package, which is what Linux and presumably other systems are using these days. What caught my attention was the 'autoflush' term and the help/usage text
```
autoflush       enable flushing of output when sending interrupt characters
autosynch       enable automatic sending of interrupt characters in urgent mode
```

Autoflush - a client-local option - does what my telnet is now doing - with an important twist that brings back the discussion about ^O above: It's using the flush function in the tty driver (!) - LNOFLSH - if it exists. If not it does nothing. All clients I have looked at (Linux, Mac, Blink on the iPad) have this option, and it's not enabled by default. A very simple way to do it since there is no need to turn off the flush since the tty gets reset at program exit. I would have thought that that ensured no output from the interrupted process at all after the ^C, but even in this case there is a single buffer being output just before the new prompt.  I haven't wrapped my head around that one yet.

Anyway, this strengthens the incentive to put flushing into the tty driver I think.

BTW - autosynch seems like a rarity, and not something I've looked into.

Mellvik:
> > Autoflush - a client-local option
> 
> NIce find! Nothing like going to the source to see how things really work.
> 
> > It's using the flush function in the tty driver (!) - LNOFLSH - if it exists. If not it does nothing.
> 
> > this strengthens the incentive to put flushing into the tty driver I think.
> 
> Both statements are close to my thoughts: don't add special processing to application programs (e.g. telnet) when the solution is better performed architecturally elsewhere. 

Completely agree with that. Finding that 'someone else had this problem a long time ago and fixed it' isn't a surprise but an important reminder to dig more in history (that is, code) before band-aiding.

>It's also somewhat telling that the option is default OFF, which also mirrors my thoughts of not assuming the user has special knowledge of applications doing special processing - they usually don't. 

Well, it's not. The default (if there is one) is on - if the client at startup time finds that the discard/flush option is supported inn the tty driver.

> 
> Another thing I find interesting about your research results is all the apparent effort put in for IAC IP / DM handling, only to not do anything with it - even when autoflush is on since the TTY driver ends up handling it and presumably doesn't use the data marks! (Or does it?!)

This is not the case - or maybe I'm misunderstanding your statement. The current version of the telnet client does send IAC_IP when a ^C is encountered. Further, when the new autoflush is implemented, the test that prevents multiple IAC_IPs on top of each other will be gone - along with the timeout Likewise, the updated `telnetd` now does what the big brothers do and returns DM appropriately. When the tty driver does the flush, the DM will be ignored, but it's still there - as the spec prescribes. 

> 
> Another possible solution (for everyone) would be adding options processing to telnet (like ftp) so that users can specify the way they want the application to behave. 

I'm assuming you mean command line options. I have been considering this, not to turn on, but to turn off the autoflush mechanism. Since our telnet - as opposed to most siblings - doesn't have a command line mode (and I'd like to keep it that way to keep the size down), a command line option would be appropriate. I was thinking '-A', having caps generally mean 'off'.

> But even ftp doesn't (at this point) allow file-based options, just commands that can be typed. I'm probably going the wrong direction here... perhaps that's why GNU utils are so large!

It is, they're huge - and the capabilities are similarly endless. The price of being compatible not only across a multitude of systems but with history ... That said I find myself more often than not using single line (per file) FTPs these days, which allows for at least some file control (although rarely used): `alias t17='ftp -u ftp://xx:yy@10.0.2.17/telnet telnet'`

> > I would have thought that that ensured no output from the interrupted process at all after the ^C, but even in this case there is a single buffer being output just before the new prompt. I haven't wrapped my head around that one yet.
> 
> Just guessing here, but since the GNU telnet is just reading/writing (copying) the input/output data, we have the same potential scheduling problem you previously mentioned, where the TTY flush occurs prior to the last write being completed by telnet, or something like that?

Well, I haven't been able to get this to make sense (yet). The flush flag (more below) should be cleared at close time and there can be no output after that. hmmm... 

> 
> > since there is no need to turn off the flush since the tty gets reset at program exit.
> 
> Actually, while possibly true in Linux, in ELKS, a TTY is not reset on close (or program exit): the baud rate, ICANON, ISIG, etc values all remain the same. This is used for effect to allow `stty < /dev/ttySX` to change a baud rate and have it remain set after exit, for instance. `getty` is the one of the very few programs which initialize a TTY to known values.

yes, this is how Unix has handled it since forever and I didn't mean 'reset' literally. What I'd expect is that when the subprocess exits, the 'typically dynamic' TTY parameters are 'reset' to the values they had before the fork - or possibly to 'sane' values (unrelated thought: Aeons ago, there was a `stty` option `sane` which brought a completely screwed terminal back to something reasonable - echo on, flush off etc). This does indeed seem to be the case too, except there is this extra 'packet' getting through. And it IS the final packet of the output, not the packet after the point at which the output stopped. While writing this I did some more testing, macOS telnet client, big Linux server, 

Somewhat off topic: I spent several hours yesterday trying to figure out why Blink telnet (you may know Blink, the *nix environment for iPadOS) doesn't work at all with TLVC but (sort of) works with other systems. I figured it must be something with our `telnetd` and to fix it now that I'm active in that department. The problem as it turned out, is that Blink telnet sends LF as end-of-line, not CR (or CR-NULL or CR-LF) as the standard says. It even converts CR to LF. Fixing this in `telnetd` is technically easy but has consequences in environments that expect LF - and starting to guess context is a no-go. The cure becomes worse than the illness. And - becoming a Blink developer and submit a fix is also a no-go these days.

Then, on a hunch, I logged in as `toor` instead of 'root' to get `sash` - and voila, all is good. The 'problem' - or rather, the reason this works on other systems but not on TLVC/ELKS is that `ash` doesn't accept LF as a 'return' character which all the other shells (Mac, Linux, ...) do. Effectively making the Blink telnet bug invisible. IOW I'm reverting `telnetd` to its previous state and fixing `ash`. 
I figured you may like to know - our `ash` versions have diverged quite a bit so there is no moving ash-code back and forth at this time.

ghaerr:
> I figured you may like to know - our ash versions have diverged quite a bit 

Thanks, I've made a note of it. Not sure whether a "fix" is worth it considering apparently only Blink absuses the standard. `ash` is internally quite complicated, especially when it comes to script execution, I have learned. I'm not sure if the "bug" is in linenoise or internally. BTW, most of my ash mods have been relating to fixing the `jobs` subcommand which is implemented outside of job control. It appears our version of ash was written during the very early days of job control when `wait` and a few other system calls were changing. This is all cleaned up in a version of ash about a year after ours but I haven't ported it since I think it will turn out much larger, which isnt worth it for a variety of reasons.

> All clients I have looked at (Linux, Mac, Blink on the iPad) have this option, and it's not enabled by default. 

> Well, it's not. The default (if there is one) is on 

I was confused between these two statement of default operations and thought we were talking about autoflush.

> What I'd expect is that when the subprocess exits, the 'typically dynamic' TTY parameters are 'reset' to the values they had before the fork 

I was just pointing out that none of the TTY parameters, typically dynamic or otherwise, are changed on any program exit, subprocess or not. This must be done by a waiting parent, or an unrelated program. When running `ash`, of course, the terminal parameters are changed both before and after a command line is typed - raw when typing and sane when finished, such that the executed command always runs with sane TTY values. `sash` doesn't change TTY settings and just uses `fgets`. Note that the special "hacked" version of fgets for ^C handling prevously used in sash and basic is no longer needed and has been removed (in ELKS).

ghaerr:
> Does this mean that job control is now working? Like including the bg and fg commands?

No, full shell-based job control is not (yet) implemented, as this requires SIGTTIN, SIGTTOUT, and various process group and controlling tty ioctl's which are not present. Also, we've run out of signals using 16-bit signal mask bits. There's quite a bit of code in ash for full jobs support - I felt a partial implemntation (see below) would be better suited to keeping ash small, while still allowing moving a command to the background using ^Z and restarting it using `kill -CONT` when desired.

However, the ash `jobs` command is implemented, which allows full tracking of jobs started with &, stopped with ^Z (SIGTSTP) and starting/stopping jobs using `kill -CONT` / `kill -STOP`. See the screenshot in https://github.com/ghaerr/elks/pull/2523 for more details. The `wait4` system call has to be enhanced for this support, along with changes to ash and sash, for proper operation. These changes can be seen in https://github.com/ghaerr/elks/pull/2509, https://github.com/ghaerr/elks/pull/2511, https://github.com/ghaerr/elks/pull/2516 and https://github.com/ghaerr/elks/pull/2520.

> Does that mean that the buffered IO library has been updated (maybe even speeded up)?

No, the problem was that if characters were typed before ^C and CR, `fgets` would return the characters received rather than -EINTR, which had the effect of discarding the ^C SIGINT. That was originally fixed via a hack to `fread` included in sash and basic, but then the TTY driver was corrected in https://github.com/ghaerr/elks/pull/2001 which returns -EINTR regardless of previous input, allowing the stdio library to work correctly without the hack code in sash/basic.

I've not seen a way to speed up stdio very much except by using getc/putc instead of fgetc/fputc, do you have specific issues you've run into? Also, using buffer sizes > 1K probably slows things down, but this would need to be measured with the precision timer routines to be sure.

> add command line history (simplified) to ash which was really easy since the history was there in Linenoise anyway.

The linenoise functionality has been moved to the C library in ELKS, so it's super easy to add history/autocompletion functionality to any program, like `basic`, with two function calls. In this way changes to `ash` have remained minimal, except for internal enhancements like getting jobs working, etc.

ghaerr:
> It's using the flush function in the tty driver (!) - LNOFLSH - if it exists. If not it does nothing. 

I decided to look further into the LNOFLSH option you've brought up to see what Linux is actually doing in the TTY driver. Long story short, it seems that this option is *not* causing the TTY output to be flushed, and might in fact be doing the exact opposite of what you are thinking it does. Further, the option is not one that is "set" on to flush, then requiring "reset" after process exit, etc in order to re-enable output.

Looking at the [definition](https://elixir.bootlin.com/glibc/glibc-2.1.96/source/sysdeps/unix/sysv/aix/bits/ioctls.h) of LNOFLSH, it's basically a synonym for NOFLUSH, a termios bit which is explained as "No output flush on signal". Looking further into this, including Linux 2.0 where this is implemented, it appears that Linux will normally always "flush" the TTY output queue when a signal is received. What we're talking about is that when ISIG is set and SIGINT received (via ^C typically or otherwise), the TTY driver will flush any output, rather than displaying it. I would guess the idea is that when a user types ^C, that user doesn't want to see pending output, from any application that might be otherwise writing TTY data.

Keep in mind that Linux uses interrupt-driven TTY output, so there is such a thing as an "unflushed" TTY output queue. In ELKS and TLVC, it is a bit more complicated, since all processes wait in the `write` system call until tty_write finishes output to the actual device (e.g. system console or serial port). For instance, kernel busy-loops inside the serial driver waiting for UART TBE before sending the next character. The writing application process fills the output queue until its limit, then calls the subdriver (rs_write in this case) to output each character using the TBE busy-loop, then the write loop repeats in another area of the kernel.

What ioctl(..., LNOFLSH) does is tell the system to NOT flush the TTY output buffer on any signal - that is, to always keep any queued contents in the TTY output buffer when a signal occurs. This can be seen in the following Linux 2.0 tty driver code isig, which is called on any signal received:
```
static void isig(int sig, struct tty_struct *tty)
{
    if (tty->pgrp > 0)
        kill_pg(tty->pgrp, sig, 1);
    if (!L_NOFLSH(tty)) {
        n_tty_flush_buffer(tty);
        if (tty->driver.flush_buffer)
            tty->driver.flush_buffer(tty);
    }
}
```
Such a thing could be implemented on ELKS or TLVC, but this would serve the opposite purpose of what you're looking for - which is to flush buffers on ^C only in telnet, even though the GNU telnet actually seemingly intends to do the opposite, unless the ioctl is actually turning OFF LNOFLSH. 

In summary, the normal Linux implementation always flushes TTY output on signals, unless LNOFLSH is set. In ELKS/TLVC, this could be made to empty any output queue currently in the middle of a `write` system call, but far easier might be to just terminate the `write` inside tty_write when any signal is received. This is further complicated by our PTY driver requiring special processing - see tty_write.

Getting this working on ELKS would require serious testing, since output would *always* be flushed by default, and decision(s) would have to be made regarding flushing just the TTY queue, and/or also flushing the applications write buffer as well, in TTY, PTY and serial output queue cases. In addition, it seems this isn't even what was discussed, which is performing this activity *only* while in telnet, versus all the time.

I'm a bit confused by all this, and am guessing perhaps GNU telnet actually intended to turn OFF LNOFLSH, which would mean always flush on signal. Not sure? But this doesn't solve the problem of flushing output for a period outside that of a signal received and the current contents of the telnet application and its associated TTY output queue's worth of data; this would have to be done inside telnet.

ghaerr:
> What do you think?

I think I became quite confused from this discussion how LNOFLSH could be made to work flushing incoming telnet data after a ^C was typed. So I took a much deeper dive, including studying the GNU inetutils telnet source. And what I learned is that the way that supposedly works from this discussion, versus how I see it actually working from reading the GNU source code, is almost entirely different. 

> The Linux telnet flush output on interrupt function works exactly as we expected.

Well - kind of. It may act the way we expected, but the implementation is not as discussed. Keep reading.

Lots to cover, so I'll break down my findings into Special input character handling, Interrupt request, Output flushing, and Autoflush.

- Special Input character handling: telnet uses termios to get the current TTY characters for VINTR, VDISCARD, VSTART, VSTOP, and other special characters. Telnet has "commands" to enable the local interpretation of each of these characters in which case the TTY driver special character is disabled. This appears to be the default, e.g. ^C, ^O, ^S ^Q etc are all handled locally instead of using any TTY driver code when the character is typed. Because of this, signal processing is usually disabled in the TTY driver, in which case it doesn't matter whether the kernel flushes output on received signals (ELKS does not).

- Interrupt request: when VINTR is typed (^C), IAC IP is sent, and output flushed only if auto flush is set. (details below on autoflush).
- Flush request: when VDISCARD is typed (^O), IAC AO (About Output) is sent, and output flushed only if auto flush is set. 

> apologies for not being more specific when mentioning LNOFLSH. I know what it does, and was surprised to find that this is the capability Linux is checking for existence before allowing autoflush. 

First, telnet doesn't actually use LNOFLSH in an ioctl. It uses ioctl(TIOCLGET,...) to get the local flags from the TTY termios.c_lflag bits, and then checks whether LNOFLSH is set. This is different than setting LNOFLSH. As previously described, Linux (but not ELKS) flushes a TTY output buffer when any signal is generated from the TTY controlling terminal. 

Telnet, on the other hand, never sets LNOFLSH in termios.c_lflag and does not use this method to disable flushing TTY output on signals received. Since telnet by default is all Special Input characters locally, no signals are usually ever generated, and output is never flushed or not flushed (disabled) using this method.

Instead, the LNOFLSH bit is queried to only to disable auto flush. Thus it follows since if a TTY has auto-flushing disabled for signals, then telnet ought to disable flushing output in its own local processing also.

Output Flushing
- When auto flush is set, which is set when termios LNOFLSH is not set, telnet uses a fairly complex method consisting of a while loop of ioctl(TCSADRAIN,...) followed by ioctl(TIOCFLUSH,...) and then repeating that until an internal output ring buffer is depleted. Unlike our telnet, GNU seems to use both input and output ring buffers for queuing receive and transmit characters, much more complicated than ELKS. I suspect this is equivalent to our telnet `discard` variable but in a much more complicated way.
- [EDIT: It appears the reason for the complicated telnet output ring buffer is to collect received data and hold it for processing for a small period of time *before* writing to stdout. The reason for this is that a subsequent DM may be receved (see below), for which the action is to discard prior received but not yet displayed data. Complicated but now I understand I think. Each block of received data is also time-stamped in the output ring buffer for subsequent processing.]
- DM received: whenever a telnet Data Mark is received, the output is flushed as described above, regardless of whether auto flush is set. After that, a timer is set, which is then used to handle determining whether Out-Of-Band data has been received or not (which ELKS TCP doesn't handle), and then uses all sorts of hack code to handle the many different kernel variations of TCP/IP OOB processing over the years.

Thus, it seems that output flushing requires lots of complexity to handle correctly, since a subsequent DM apparently means "discard previous data". That won't work if telnet has already written the data to the terminal! Thus the output ring buffer. For ELKS, far simpler to keep a small timeout timer, or perhaps (with more risk) discard output until a DM is received. If a timer is set for waiting either for DM or no incoming data, then this might easily work for ELKS/TLVC without keeping a ring buffer.

In summary, while the overall idea is the same, GNU telnet handles ^O abort output and output flushing differently than discussed. ELKS doesn't support any programmatic output flushing via either ioctl mentioned, while both attempt handling of all special characters internally by default. Given that ^O and ^C need special processing for output flushing entirely different than a TTY driver VFLUSH implementation, I think it better to keep ELKS and TLVC special processing within telnet.

I'll leave it to you to piece together this new information with the outbound flow of ^C, IAC IP, ^O and IAC AO (separate or combined) and the subsequent receipt of IAC DM from telnetd as you have researched. This should then give us sufficient information as to exactly when to stop discarding input data and how receipt of DM could be connected to it.

Overall, given that ELKS telnet hangs in the TTY driver write routine already and does not currently ever return early, adding a driver flush function seems pretty unnecessary (other than for future general ^O processing, which won't work for telnet anyways). When telnet returns from write, any ^C or ^O input special character can be acted upon immediately, and the discard variable set just like we currently do, and the user won't be able to tell that ^C processing actually happened after the last write, which will always have been < 1500 characters anyways (e.g. displayed very quickly).

ghaerr:
As an add-on to the above, from reading about the telnet specification, it seems that whenever IAC IP or IAC AO is sent, the sender is also supposed to set the TCP Urgent bit to specify OOB data, and then follow either command with an IAC DM within the urgent data. The reason for this is the that receiver doesn't know when to end "special processing" of IP or AO requests until a DM is seen. This is called the telnet Sync Function and is used to guarantee immediate processing of special requests even though the receive window may be full. Apparently for speed reasons the receiver (telnetd) doesn't in fact scan for IAC across all data, but only during a "Sync Function". Inr ELKS / TLVC is there is not OOB capability, but a DM may have to be sent anyways. Also, our client and server scan for IAC at all times.

Because of all this added complexity, it might be good if our simple telnet client can just send ^C which you've reported works the same as IAC IP, correct? I'm not sure whether a ^C will cause a DM to be returned however. If not, then we probably have to send IAC IP IAC DM, then wait for DM or timeout. (Sounds like only waiting for a timeout would still work just fine). For our telnetd server, if it can be made to work without having to handle IAC IP but just an incoming ^C from a Linux or macOS telnet client, things will likely be much simpler. In other words, the simple ELKS implementation of telnet and telnetd apparently already work well with all known servers, despite not knowing about or ever sending IAC DM or IAC IP.

My conclusion is that telnet is an ancient protocol that was designed long ago when urgent / OOB data was generally required for performance reasons, but nowadays, none of the later protocols (RSH, RLOGIN, etc) use anything special, they just transfer data as input and output, without any IAC or DM handling. Thus, I'm hoping to find a solution that will enable minimal changes to telnet and telnetd (for ELKS, at least), since the further one jumps into the telnet protocol, the uglier it gets it seems, to actually conform to the spec.

ghaerr:
Thanks for your summary. It seems we now know enough to make an informed choice about our respective implementations.    Mine will be skinny, because I enjoy keeping things absolutely as small as possible yet highly functional in ELKS. And thanks to this thread, I've realized that adding support for DM really is mandatory in order to get interrupt process to work well, in both telnet and telnetd. More comments follow.

> SYNC is not implemented in any of the servers, neither is AO. There is no OOB usage.

I havne't studied Urgent Data much at all, but it seems the problem is on the receiving end is that there is no way to know that any data is urgent, unless special historically incompatible socket options are used. So the data just appears ahead of the normal data stream, that's all. Which brings up my point: SYNCH is in fact performed in the server through the mandatory use of DM being sent at the end of urgent data. This tells the server to "look backwards" (which actually means look at the data just prior to the DM) and interpret IAC commands. If a server always performs IAC search, then there is no meaning in SYNCHing.

Separately, not having AO support means to me there's no reason to ever send IAC AO. Local ^O just starts the auto-flush mechanism in the telnet client, which will require the use of a timer, *unless* IAC AO causes the server to respond with DM.

> All telnet servers I have come across support IAC_IP and reply with DM. They also accept ^C directly and reply similarly.

That's great that ^C also works. I haven't decided whether this means our telnet needs to convert ^C to IAC IP, but it sounds like our telnetd must specially signal(SIGINT) the controlling TTY or else ^C signaling won't work telnet outside ELKS. It would be interesting to test this to make sure.

> OTOH, this is emergency, should never happen - reception of DM is reliable, so it's better to make it too long than too short.

I have some ideas about how the timeout can be changed pre- and post-DM which would both solve the problem of missing DM (e.g. older ELKS telnetd's) as well as getting rid of that last bit of output resulting from scheduling issues within the telnetd system and the remote shell sending data just after the DM.

>  adding LNOFLSH is not worth the effort unless it is required in the context of (for example) job control.

I think there may still be some confusion on this: the issue is whether the TTY driver should flush the output queue when a signal is received. This doesn't need LNOFLSH and is by far a bigger issue then "LNOFLSH turning off a feature we don't even implement".

> adding flush (^O) and LNEXT (^V) would be welcome from my perspective. 

What exactly would you use either of ^O or ^V for? I can't think of any ELKS application that keeps writing such that an ^O would be useful - sl? What would one do while the output was discarded anyways, especially since any character typed would reenable output?

ghaerr:
> After deleting the debug printfs, connecting to the telnet server while another connection is receiving data will hang 

What makes you think this is data corruption? Is the entire system hung, i.e. no login available elsewhere, or is it just the group of PTY tasks that are hung? There is a very long explanation for exactly why our PTY implementation needs to allow intermittent rescheduling to avoid a deadlock between the reader/writer processes. I need more info to know whether that's what's actually happening. The following code in ntty.c:tty_write() is just the tip of the iceberg on this phenomenon:
```
    while (i < len) {
        s = chq_wait_wr(&tty->outq, (file->f_flags & O_NONBLOCK) | i);
        if (s < 0) {
            /* FIXME EAGAIN not returned, cycle required on telnet nonblocking terminal */   
            if (s == -EINTR || s == -EAGAIN) {
                tty->ops->write(tty);
                wake_up(&tty->outq.wait);
                schedule();
                continue;
            }
```

PR #238, opened 2026-04-02
[kernel,net] ne2k driver rewrite, now completely buffered

Mellvik:
With reference to the discussion in #237, this represents the first cut of a new generation of network drivers: Completely buffered and interrupt driven, taking advantage of the new TH/BH interrupt handler structure and simplifying the interface between kernel IO and device IO.

The buffer mechanism itself is unchanged from the original implementation (2023) but the optional static buffers and the 'no buffers' option have been removed. The interrupt handler and the read/write routines have been simplified and optimized, as has the assembler code. The low level read from the NIC now always reads 256 bytes in the first read operation - instead of only 4 to get the header with the packet length. If the packet is smaller than 252 bytes including headers, this is the only operation required - and obvious optimization compared to always doing two read operations when packed are small, which typically is the case in `telnet` connections. When packets are larger the difference is nil, if smaller it's negligible. This is the way the read operation worked from approximately 2018 until direct reads into user space were implemented in ~2020.

A nasty race condition in write/xmit led to a long and thorough rabbit hunt through most of the driver which uncovered some esoteric (and not-so-esoteric) bugs along the way and overall contributed to the driver now being at least as robust as before the structural change. Error handling in particular has improved and delays when buffers are temporarily unavailable have been minimized. Testing/rabbit hunting entailed 3 hardware platforms and 3 different ne?k interfaces plus QEMU, literally thousands of test runs across all of them.

This driver is the first to take full advantage of the TH/BH structure and the experiences have been very encouraging. It's now a valuable role model for repeating the structural changes in other net drivers, starting with the SMC/WD driver. Alongside this the new buffer structure mentioned in #237 is coming along nicely, and with it the option to place the buffers anywhere in memory.

EDIT: About performance - see next post.

### DIFF
--- a/tlvc/arch/i86/drivers/net/ne2k-asm.S
+++ b/tlvc/arch/i86/drivers/net/ne2k-asm.S
@@ -193,19 +193,13 @@ dma_init:
 //-----------------------------------------------------------------------------
 //
 // BX    : NIC memory address (to write to)
-// CX    : byte count
+// CX    : byte count (must leave unharmed)
 // DS:SI : host memory address (to read from)
-// Also using the 3rd arg to the caller (local/user flag) from the stack
+// Returns bytes written in ax
 //-------------------------------------
 
 dma_write:
 
-	push    %cx
-	push	%ds
-
-	inc     %cx     // make byte count even
-	and     $0xfffe,%cx
-	//cli		// Mandatory
 	call    dma_init
 
 	// start DMA write
@@ -218,15 +212,10 @@ dma_write:
 
 	//mov	net_port,%dx
 	add	$io_ne2k_data_io,%dx
-	mov	ne2k_flags,%ax		// Get this before changing the data segment
-	mov	8(%bp),%bx		// get the 3rd arg from the call
-	cmp	$0,%bx			// zero means kernel (buffer)
-	jz	1f			// use kernel data seg
-	mov	current,%bx		// setup for user data seg
-	mov	TASK_USER_DS(%bx),%ds
-1:
+	mov	ne2k_flags,%ax
+	push	%cx		// caller expects CX to be preserved
 	cld
-	test	$ISA_8B,%ax		// checking ne2k_flags
+	test	$ISA_8B,%ax	// check ne2k_flags: 8 or 16 bit
 	jz	wr_loop_w
 
 	// Byte loop
@@ -234,10 +223,9 @@ wr_loop_b:
 	lodsb
 	outb	%al,%dx
 	loop	wr_loop_b
-
 	jmp	wr_loop_done
 
-	// word loop
+	// Word loop
 wr_loop_w:
 	shr	%cx
 3:	lodsw
@@ -245,22 +233,7 @@ wr_loop_w:
 	loop	3b
 
 wr_loop_done:
-	
-	pop	%ds	// get the data segment back
-			// otherwise the net_port reference below won't work
-	// wait for DMA completed
-check_dma_w:
-	mov	net_port,%dx
-	add	$io_ne2k_int_stat,%dx
-	in      %dx,%al
-	and	$0x40,%al	// make sure we're done
-	jz	check_dma_w
-
-	//mov	$0x40,%al       // clear DMA intr bit in ISR
-	out	%al,%dx
-
-	//sti			// Mandatory
-	pop     %cx
+	pop	%cx
 	ret
 
 //-----------------------------------------------------------------------------
@@ -270,28 +243,20 @@ check_dma_w:
 // BX    : NIC memory to read from
 // CX    : byte count
 // ES:DI : host memory to write to
-// AL:	 : 0: buffer is local (kernel), <>0: buffer is far (process)
 
 dma_read:
 
 	push    %di
 	push    %es
 	push	%bx
-	push	%ax
 
 	inc     %cx     // make byte count even
 	and     $0xfffe,%cx
 
-	//cli		// Experimental - disable INTR
 	call    dma_init
 
 	mov     %ds,%bx
-	mov     %bx,%es
-	pop	%ax
-	cmp	$0,%al		// Use local (kernel DS) buffer if zero
-	jz	rbuf_local
-	mov	current,%bx	// use task DS buffer
-	mov	TASK_USER_DS(%bx),%es
+	mov     %bx,%es		// using kernel data segment (heap)
 
 rbuf_local:
 	mov	net_port,%dx	// command register
@@ -318,19 +283,6 @@ word_loop:
 	stosw
 	loop	word_loop
 3:
-	// PROABLY NOT NEEDED
-	//mov	net_port,%dx
-	//add	$io_ne2k_int_stat,%dx
-check_dma_r:
-	//in      %dx,%al
-	//test    $0x40,%al       // dma done?
-	//jz      check_dma_r
-	
-
-	mov     $0x40,%al       // clear ISR (RDC bit only)
-	out     %al,%dx
-	//sti			//Experimental - Enable INTR
-
 	pop	%bx
 	pop	%es
 	pop     %di
@@ -427,8 +379,6 @@ nrs_exit:
 //-----------------------------------------------------------------------------
 // arg1: buffer to receive the data
 // arg2: int - requested read size (max buffer)
-// arg3: int array [2] (return) containing the NIC packet header.
-// arg4: flag whether dest is kernel buffer or proc data
 //
 // returns:
 // AX : < 0 if error, >0 is length read
@@ -440,78 +390,50 @@ ne2k_pack_get:
 	push    %bp
 	mov     %sp,%bp
 	push    %di
-	//push	%es
 
-	//sub 	$4,%sp		// temp space
-	//mov	%sp,%di
-
-	// get the 4 byte header first -> arg3
-	mov	8(%bp),%di	// arg3
 	mov	ne2k_next_pk,%bh
 	xor	%bl,%bl		// Next pkt to read in BX
 
-	mov	$4,%cx		// Bytes to read
-	//mov     %ds,%ax
-	//mov     %ax,%es	// local address space
-	xor	%al,%al		// indicate kernel DS
+	// get the first page - which includes the metadata (header) in front
+	mov	4(%bp),%di	// Buffer address to receive data.
+	mov	$256,%cx
 	call	dma_read
-
 	mov	0(%di),%ax	// AH : next record, AL : status
 	mov	2(%di),%cx	// packet size (without CRC)
 
-	// get the actual data
-
-	//add	$4,%sp
-	mov	4(%bp),%di	// Buffer address to receive data.
-	mov	6(%bp),%dx	// read len
+	mov	6(%bp),%dx	// Adjust read length if necessary,
 	cmp	%cx,%dx		// choose the shorter
-	jnb	npg_cont0
+	jnb	1f		// FIXME: a shortened packet is a damaged packet
+				// of no use, should abort right here
 	mov	%dx,%cx		
 
-#if 0
 	// -------------------------------------------------------------
 	// Packet size check not required since the NIC will not
 	// accept such packets per our initialization. Note, when handling
-	// erroneous packets manually, the rx_get (BOUNDARY) pointer must subsequently
+	// erroneous packets manually, the rx_get (BOUNDARY) pointer must 
 	// be updated do point to the next packet.
 	// If oversized packets still occur, it's a driver problem (most likely
 	// reading the wrong buffer page from the NIC).
 	// -------------------------------------------------------------
-	or      %cx,%cx		// zero length
-	jz      npg_err2
-
-	cmp     $1528,%cx	// max - head - crc
-	jnc     npg_err
-#endif
-	// -------------------------------------------------------------
-	// This section did the smart thing when reading the NIC packet header:
-	// Got the entire block (256b) instead of the 4 first bytes. Which was great 
-	// for small packets (telnet, command packets etc.): One read instead of 2.
-	//
-	// Removed when read was changed to put data directly into process address space,
-	// there is no longer anywhere to put the first 4 bytes.
-
-#if 0
+1:
 	sub	$252,%cx	// Got entire packet?
 	jle	npg_cont
 				// If not, get rest.
 	inc	%bh		// Point to next page
-	cmp	$rx_last,%bh	// check wraparound
+	cmp	ne2k_rx_last,%bh	// check wraparound
 	jnz	npg_cont0
 	mov	$rx_head,%bh
-#endif
 
 npg_cont0:
-	//add	$256,%di	// Update destination memory address 
-				// (keep the 4 byte NIC header)
-	push	%cx		// save length
 	push	%ax
-	add	$4,%bx		// Skip the 4 bytes already read
-
-	//mov	$1,%al		// use far transfer
-	mov	10(%bp),%ax	// buffer type flag
+	mov	net_port,%dx
+	add	$io_ne2k_int_stat,%dx
+	mov	$0x40,%al	// Clear RDC interrupt bit
+	out	%al,%dx
+	add	$256,%di	// Update destination memory address 
+				// (keep the 4 byte NIC header)
 	call    dma_read
-	pop     %ax
+	pop	%ax
 
 	// update RX_get pointer (BOUNDARY, end of ring)
 npg_cont:
@@ -525,7 +447,6 @@ npg_cont:
 	dec	%al
 
 npg_next:
-
 	mov	net_port,%dx
 	add	$io_ne2k_rx_get,%dx	// update RX_get (BOUNDARY)
 	out     %al,%dx
@@ -542,10 +463,7 @@ npg_exit:
 
 npg_exit_ok:
 	sti
-	pop	%ax	// return byte count (from %cx)
-	//pop	%es
 	pop     %di
-	//mov	%bp,%sp	// restore stack pointer
 	pop     %bp
 	ret
 
@@ -582,9 +500,8 @@ nts_exit:
 //-----------------------------------------------------------------------------
 // arg1 : packet buffer to transfer
 // arg2 : size in bytes
-// arg3 : 1 if data is far, 0 if kernel data seg
 // returns:
-//	AX : error code
+//	AX : bytes written (no error conditions)
 
 	.global ne2k_pack_put
 
@@ -593,12 +510,15 @@ ne2k_pack_put:
 	push    %bp
 	mov     %sp,%bp
 	push    %si
+	//cli			// DEBUG
 
-	// write packet to chip memory
+	// write packet to NIC memory
 
-	mov     6(%bp),%cx	// arg2 - count
+	mov     6(%bp),%cx	// arg2 - byte count
+	inc	%cx
+	and	$0xfffe,%cx	// make even
 	xor     %bl,%bl
-	mov     $tx_head,%bh
+	mov     $tx_head,%bh	// always the same start address
 	mov     4(%bp),%si	// arg1 - buffer
 	call    dma_write	// copy the data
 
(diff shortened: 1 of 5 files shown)


Mellvik:
In addition to the architectural issues and reliability requirements, performance is an obvious factor in this rather big change. And while not having received much focus during development - after all most of the basics are the same, testing is a different story altogether. 

TL;DR: The performance of the new `ne2k` driver is generally slightly lower than the old (unbuffered) one. The difference in file transfers is in the minus 5-10% range (~55kb/s) (I don't have the old numbers easily at hand) after some tuning - inspired by the new (buffered)  `wd` driver which turned out to be faster than the ne2k driver (the buffered `wd` driver is being submitted in a separate PR).

Interestingly (although not surprisingly), the ping numbers are very different from before. While they used to be stable at (say, depending on the host CPU) ~1.3ms for a standard ping, they now vary constantly between 10ms and 1.5ms. The variation of not its cyclic pattern makes sense and brings understanding to how the new TH/BH interrupt structure works in practice. It also adds inspiration to further work on optimizing asynchronous IO on TLVC.

More numbers on performance are coming as more drivers are added to the new interrupt/buffer regime.

Having two drivers converted, it becomes apparent that some routines are identical or almost identical across - as discussed in #217. Sharing them instead of keeping separate identical copies becomes a natural progression. The `netbuf.h` and `netbuf.c` files have been renamed `netlib.*` in preparation for becoming 'host' to such shared routines - `write`, `read` and `select` being immediate candidates, `open` and `release` optional.

ghaerr:
Very nice getting the NIC driver rewrites moving forward and removing lots of older cruft!!

> The performance of the new ne2k driver is generally slightly lower than the old (unbuffered) one. 

Do you think the reason is because of the copying of every packet? I think previous testing showed that, at least for the 8088, the buffer copying (of any kind) definitely slows down overall system throughput, while at the same time, 386+ systems have no issues. Systems in between will likely also show performance degradation.

Given this potential issue, what was the reason for removing buffering completely, was it to simplify the driver design, so that the bottom half always runs identically, regardless of buffer location? It would seem that at least read operations might be able to skip an intermediate buffer.

I'm traveling and haven't had time to study the new code much, but, given that only a single application ever talks to the NIC at any time, and that application always reads/writes serially, I'm wondering a bit about how much performance effect forced buffering has (or hasn't).

> it becomes apparent that some routines are identical or almost identical across 
> Sharing them instead of keeping separate identical copies becomes a natural progression. 

That'll be a nice addition, for sure. I'm thinking though - instead of moving nic_read/write/select() etc into a library, perhaps those three (or more) common entry points should stay in one upper-level "master" NIC driver file, and then call out to lower-level "sub drivers" using a function table? Other routines not used in a master NIC driver would fit well into a netlib library. [EDIT: thinking a bit more, perhaps its no big deal to keep shared routines in a library and skip the overhead of a "master" NIC driver].

> the ping numbers are very different from before. While they used to be stable at (say, depending on the host CPU) ~1.3ms for a standard ping, they now vary constantly between 10ms and 1.5ms.

Interesting. I would guess that the fast executions (1.5ms) would be all the bottom halves executing directly after a hardware interrupt, where the BH can be scheduled immediately, but that the slow executions (max 10ms, the timer tick) may be the result of a bottom half being requested, but not scheduled, until the next hardware timer tick, due to restrictions on the way that the kernel can or can't schedule BH executions. We should think about this in more detail, since read/write system calls should also allow scheduling a BH just before returning from the syscall (which shouldn't cause a delay, unless perhaps a task was switched). I'm happy to help look further into this if desired.

I'll comment further on code and the WD driver when I can.

Mellvik:
> > The performance of the new ne2k driver is generally slightly lower than the old (unbuffered) one.
> 
> Do you think the reason is because of the copying of every packet? I think previous testing showed that, at least for the 8088, the buffer copying (of any kind) definitely slows down overall system throughput, while at the same time, 386+ systems have no issues. Systems in between will likely also show performance degradation.

First, the difference is small and the tests so far non-empiric, what I'm reporting is the first readings - 'off the bat'. Previous testing with buffers (before the new interrupt regime) showed no speed difference on fast systems because the systems were generally fast enough to keep up with the NIC, and most IO still went direct, bypassing the buffers. That was not the case on slow systems and the presence of buffers improved general performance noticeably.

I don't think buffer copying is having any impact at all, it's just a too small part in the big picture, possibly with the exception of the slowest XT systems. But even in those cases, I believe it's negligible overall. I think the difference is the delay between write-to-buffer and write-to-NIC, ie. the BH delay. 

> 
> Given this potential issue, what was the reason for removing buffering completely, was it to simplify the driver design, so that the bottom half always runs identically, regardless of buffer location? It would seem that at least read operations might be able to skip an intermediate buffer.

First, we may be talking about different things here. When comparing performance - however non-empiric - 'before' means 'before TH/BH regime, old buffering scheme' (ne2k), 'new' means 'TH/BH regime, fully buffered'. Second, I'm assuming 'removing buffering completely' is a typo since we're doing the opposite.  Third, the motivation for making network IO completely buffered is 1) asynchronicity, 2) general performance - enabled in part by the new interrupt structure, 3) cleaner (more reliable code) and 4) smaller. E.g. there are now literally no interrupt blocking in the network drivers (the minimal blocks remaining in the code are going away). Exceptions may come as more drivers are added, in particular the Lance case, which uses DMA and system RAM for NIC buffers. 
Anyway, the clean-ness of the new scheme, which is not fully exploited yet, is very attractive, as is the ascynchronicity, which will deliver improved performance on a running system.

> 
> > the ping numbers are very different from before. While they used to be stable at (say, depending on the host CPU) ~1.3ms for a standard ping, they now vary constantly between 10ms and 1.5ms.
> 
> Interesting. I would guess that the fast executions (1.5ms) would be all the bottom halves executing directly after a hardware interrupt, where the BH can be scheduled immediately, but that the slow executions (max 10ms, the timer tick) may be the result of a bottom half being requested, but not scheduled, until the next hardware timer tick, due to restrictions on the way that the kernel can or can't schedule BH executions. We should think about this in more detail, since read/write system calls should also allow scheduling a BH just before returning from the syscall (which shouldn't cause a delay, unless perhaps a task was switched). I'm happy to help look further into this if desired.
> 

Yes, this is an interesting discussion with potentially significant performance implications. The pings (on an otherwise idle system) are visualizing how this works in a great way, and - combined with debug `kputchar`s in the code show exactly the sequence of things: 
```
helge@raspi2b:~ $ ping 10.0.2.17
PING 10.0.2.17 (10.0.2.17) 56(84) bytes of data.
64 bytes from 10.0.2.17: icmp_seq=1 ttl=64 time=25.8 ms <---- includes ARP exchange
64 bytes from 10.0.2.17: icmp_seq=2 ttl=64 time=4.27 ms
64 bytes from 10.0.2.17: icmp_seq=3 ttl=64 time=2.44 ms
64 bytes from 10.0.2.17: icmp_seq=4 ttl=64 time=10.5 ms
64 bytes from 10.0.2.17: icmp_seq=5 ttl=64 time=8.66 ms
64 bytes from 10.0.2.17: icmp_seq=6 ttl=64 time=6.53 ms
64 bytes from 10.0.2.17: icmp_seq=7 ttl=64 time=4.70 ms
64 bytes from 10.0.2.17: icmp_seq=8 ttl=64 time=2.46 ms
64 bytes from 10.0.2.17: icmp_seq=9 ttl=64 time=10.5 ms
64 bytes from 10.0.2.17: icmp_seq=10 ttl=64 time=8.67 ms
64 bytes from 10.0.2.17: icmp_seq=11 ttl=64 time=7.46 ms
64 bytes from 10.0.2.17: icmp_seq=12 ttl=64 time=5.65 ms
64 bytes from 10.0.2.17: icmp_seq=13 ttl=64 time=3.57 ms
64 bytes from 10.0.2.17: icmp_seq=14 ttl=64 time=11.6 ms
64 bytes from 10.0.2.17: icmp_seq=15 ttl=64 time=9.66 ms
64 bytes from 10.0.2.17: icmp_seq=16 ttl=64 time=7.62 ms
64 bytes from 10.0.2.17: icmp_seq=17 ttl=64 time=5.56 ms
64 bytes from 10.0.2.17: icmp_seq=18 ttl=64 time=4.39 ms
64 bytes from 10.0.2.17: icmp_seq=19 ttl=64 time=3.19 ms
^C
--- 10.0.2.17 ping statistics ---
19 packets transmitted, 19 received, 0% packet loss, time 18031ms
rtt min/avg/max/mdev = 2.448/7.552/25.831/5.130 ms
helge@raspi2b:~ $ 
```
With 1s intervals between `pings` there are 5 steps per cycle, 2ms apart. Reducing the interval to 0.5s doubles the number of steps and reduces the steps to 1ms. Further interval reduction makes no difference. Given the various clock intervals in the system I'm sure this 'staircase' behaviour can be timelined to the 1ms level if desired.

If a write system call scheduled a BH run after returning, I suspect this behaviour would be all but gone - and the general (file transfer) performance indistinguishable or better than the 'direct' NIC-to-userspace regime. Again, keep in mind that as-is, the file transfer performance on a loaded system seems identical to the old scheme, which I think is pretty good. Improvements over that are pure 'profits'.

I'm working on (testing) additional changes to the `wd` driver type-detection/autoconfig code before completing that PR. 

BTW - Another small change coming is parameterization of the buffer segment, enabling a disconnect of  the buffers from the HEAP, thus allowing experimentation with high buffer counts instead of max ~6 as is the case now (the new buffer scheme will contribute to that too of course). Or just put them all in XMS if available.

Finally, and sort of off topic, I've been - off and on - coming back to the idea of moving TCP back into the kernel. With the fartext segment being available (for (ab)use) and the ability to release unused code space (INITPROC style), the arguments for keeping tcp in user space are fading. We should be able to 'drop' the tcp code from the kernel at boot time if desired. AnFaster, cleaner - and the total code space for networking should be smaller too. (except it would be more tempting to add functionality etc. ...)

ghaerr:
>  I don't think buffer copying is having any impact at all, it's just a too small part in the big picture, possibly with the exception of the slowest XT systems. But even in those cases, I believe it's negligible overall.

Yes its not like the system is continuously reading/writing networking packets like the system accesses floppy or hard disk file streams, so agreed buffering likely doesn't add much delay.

> I think the difference is the delay between write-to-buffer and write-to-NIC, ie. the BH delay.

> If a write system call scheduled a BH run after returning, I suspect this behaviour would be all but gone

I had thought I had coded a check for BH execution after syscalls, but just looked and see that it doesn't. While some delays in BH scheduling are unavoidable, I think it will be safe to allow BH scheduling directly after a syscall, since the execution environment is very similar to that of just prior to a hardware interrupt interrupting user mode execution, where BH scheduling is allowed (i.e. intr_count == 0 at \_irqit entry). We can look into adding this, which should eliminate the up to ~10ms delay you've found.

You might try the following code to test:
```
...
        call    syscall
        push    %ax             // syscall return value in ax

#ifdef CONFIG_TRACE
        // strace.c must be compiled with tail optimization off to protect top of stack
        call    trace_end       // syscall return value is top of stack
#endif

// new code here
        cmpw    $0,bh_active    // Any active bottom halfs?
        je      1f              // No
        call    do_bottom_half  // Run bottom halves
        //call schedule         // experimental
// end of new code

1:      call    do_signal       // process signals
        cli
        jmp     restore_regs
...
```
Additionally, a `call schedule` may want to be made after calling do_bottom_half, but that may cause problem, need to think about that a bit more. Doing so may starve ktcp of execution, prior to having it waking up the writer app, etc. 

> I've been - off and on - coming back to the idea of moving TCP back into the kernel. 

I'm not sure TCP was ever 'in' the kernel, but could be interesting to think about - the tcpdev code is messy with lots of buffer copying, semaphores and task switches. The cons are that ktcp would be possibly bringing back a bunch of bugs to the networking system, and lots of (re)testing. Also would need to think about the issue(s) of where the TCP connection data (and buffers) would be stored, and the problem that the kernel can't use ktcp (C) style local variable declarations on the kernel stack easily. 'Far' TCP connection buffers would work, but also complicate things if placed arbitrarily in system memory. In general, we don't have an extra 32k of ktcp heap to allocate within the kernel.

Mellvik:
Thanks @ghaerr. 
In order to ensure the correctness of the conclusions, I ran a comprehensive set of benchmarks with the 'old' drivers and the pre-new-interrupt-scheme in the kernel. That put the new performance readings into perspective, in particular outgoing transfers - which clocked in at about 50% (!), even lower when going through a router (which changes the packet timing slightly) and notably, much faster (but still slower than the 'old') when using 1 transmit buffer vs 3.

Good news: As expected - after your suggested change in `irqtab.S` the negative difference between old and new is gone. The differences are 'noise level' and insignificant on an idle system. Also there is no difference between 1 and 3 output buffers (these transfers, on an idle system, hardly ever use more than one transmit buffer).
I'm expecting, but don't have a good test case yet, that the difference between old and new would become much more visible on a loaded system - in favour of the buffered version, and also on a slower system, which I haven't tested yet.

Interestingly - and something we've seen since forever - outgoing file transfers are typically more than 50% faster than incoming even though incoming packets are full size, outgoing limited to 512 bytes payload. IOW incoming is ⅓ the number of packets and half the speed. The reason of course being `ktcp` scheduling - but its size keep surprising me.

> I'm not sure TCP was ever 'in' the kernel, but could be interesting to think about - 

The 'back into the kernel' was a historical reference, no connection to ELKS. And I don't (at least didn't) think that using `ktcp` as a base in such a setting would be smart. But this is a long shot. A shorter one would be to handle ICMP and ARP in the kernel, which would ease the system load when connected to a busy network with a lot of ARP and broadcast traffic (switches and other devices are really chatty these days, which is totally insignificant on a Gigabit/10gig ethernet, not so on 10M ethernet).

Thanks for the `irqtab` fix. During the testing I found that the last (?) race condition in the drivers (in the now common `_write()` routine) was still there so the previous fix didn't work. The fix before that (which disabled irqs for a few instructions) worked though, so it's fixed but I still haven't figured out why.

ghaerr:
> That put the new performance readings into perspective, in particular outgoing transfers - which clocked in at about 50% (!)

> after your suggested change in irqtab.S the negative difference between old and new is gone. 

Are you saying that the new performance is 50% slower than the old performance, but that adding the BH scheduling fix after syscalls changes the performance to the same speed? I'm not quite following what your testing has shown.

It would be nice if networking was faster using the BH scheduling, but if not, then that likely just means that the network driver code isn't a speed bottleneck.

> there is no difference between 1 and 3 output buffers (these transfers, on an idle system, hardly ever use more than one transmit buffer).

Are you thinking that, then, there may be no need to introduce the overhead of buffered transmit packets?

> so the previous fix didn't work. The fix before that (which disabled irqs for a few instructions) worked though, so it's fixed 

You had previously mentioned that it might be a bad idea to disable interrupts in the new BH-style drivers, right? Actually, I don't think necessarily so. If interrupts are disabled just around a small critical section where a single "head" pointer is updated, that shouldn't affect system operation speed- or resource- wise, and IMO would be preferable to coding up a more complicated head/tail buffer system (which sounds like didn't work due a race condition?).

> IOW incoming is ⅓ the number of packets and half the speed. The reason of course being ktcp scheduling

Are you sure the reason is ktcp being scheduled, not something else? Is this one of the reasons you're thinking it would be nice to get the kernel running TCP/IP instead of ktcp? I don't actually know where most of the time is being taken up in the networking subsystem(s). It might be interesting to profile that; I've been thinking about implementing kernel profiling for a while now, which would tell us a lot when combined with the kernel system.map file.

> I don't (at least didn't) think that using ktcp as a base in such a setting would be smart. 

Actually, I think keeping ktcp as-is as much as possible would be the only way to move TCP into the kernel. This is because 1) the ktcp code is known 16-bit working, which very few other TCP's are, and 2) moving the code over with as few changes as possible would greatly simplify getting/keeping things working, versus the high workload required for getting an entirely new TCP/IP up and running from scratch (inside the kernel or not, for that matter).

> But this is a long shot. A shorter one would be to handle ICMP and ARP in the kernel, which would ease the system load 

That could be a good idea to start with the only the low level - NIC drivers / network layers 1-3 into the kernel first, then followed by layer 4 and up. We could discuss more in a new issue if you get serious about it.

Mellvik:
> > That put the new performance readings into perspective, in particular outgoing transfers - which clocked in at about 50% (!)
> 
> > after your suggested change in irqtab.S the negative difference between old and new is gone.
> 
> Are you saying that the new performance is 50% slower than the old performance, but that adding the BH scheduling fix after syscalls changes the performance to the same speed?

Yes.

> 
> It would be nice if networking was faster using the BH scheduling, but if not, then that likely just means that the network driver code isn't a speed bottleneck.

Agree, but having worked and experimented with the buffered version of the `ne2k` driver for several years now, significant performance improvement was not on my list of expectations. A small one, yes, and it's there - sometimes. The drivers are and have been very efficient, the 'direct user space-to/from-NIC' contributing to that. When fully buffered clocks in at a similar speed, sometimes slightly better, is actually quite good. 
As pointed out previously in this thread, to me this change is about structure (and code size), not speed. (That said, the speed factor comes in when the system is loaded instead of mostly idle, but the benefits will not materialize until an interrupt driven HD driver is up and running, and never with most XT_IDE/XT_CF interfaces which  don't support interrupts).
Another element is of course that the slower the system, the more important receive buffers become, reducing NIC buffer overruns.

> 
> > there is no difference between 1 and 3 output buffers (these transfers, on an idle system, hardly ever use more than one transmit buffer).
> 
> Are you thinking that, then, there may be no need to introduce the overhead of buffered transmit packets?

Not for performance reasons, only for structural reasons. The balance point is system speed vs NIC speed, really Ethernet speed (disregarding the effect of a busy network). [I really need to get `ping` running on TLVC for testing and benchmarking.] Not even my 40MHz 386SX can saturate the 10Mbps Ethernet - my guess is a 486/66 is needed to do that. So 1 - or 2 for good measure - TX buffer(s) is just fine.

> 
> > so the previous fix didn't work. The fix before that (which disabled irqs for a few instructions) worked though, so it's fixed
> 
> You had previously mentioned that it might be a bad idea to disable interrupts in the new BH-style drivers, right? Actually, I don't think necessarily so. If interrupts are disabled just around a small critical section where a single "head" pointer is updated, that shouldn't affect system operation speed- or resource- wise, and IMO would be preferable to coding up a more complicated head/tail buffer system (which sounds like didn't work due a race condition?).

I did indeed mention that, not necessarily as a 'bad idea', more like undesirable, a matter of principle. And - if it does turn out to be necessary, disabling just NIC interrupts is easy - and preferable to blocking all interrupts. I agree with you - avoiding IRQ blocking does not qualify code complexity. OTOH, the case at hand is really simple - which is why I've found it worthwhile to spend extra time on it: I really need to understand the problem before accepting the fix, even if it is IRQ blocking. The symptoms don't (yet) make sense, and as I write I can (again) no longer recreate the problem(!), not even with the exact same setting/setup as before. It seems an interrupt has to happen at a very specific point in the driver `_write` routine (data gets corrupted). The code is extremely simple and no logical (!) risk of contention (race conditions). A mystery exacerbated by the volatility of the symptoms.

> 
> > IOW incoming is ⅓ the number of packets and half the speed. The reason of course being ktcp scheduling
> 
> Are you sure the reason is ktcp being scheduled, not something else? Is this one of the reasons you're thinking it would be nice to get the kernel running TCP/IP instead of ktcp? I don't actually know where most of the time is being taken up in the networking subsystem(s). It might be interesting to profile that; I've been thinking about implementing kernel profiling for a while now, which would tell us a lot when combined with the kernel system.map file.

Indeed - that would be something - and right up your alley I would think. And yes I'm pretty sure this is about ktcp and scheduling simply because it makes sense (to me) the way I understand the system. Output is 'simple' - push the data and continue, read is less predictable and involves more waits (including select). It is indeed my primary motivation for pondering the TCP_IN_KERNEL idea.

> 
> > I don't (at least didn't) think that using ktcp as a base in such a setting would be smart.
> 
> Actually, I think keeping ktcp as-is as much as possible would be the only way to move TCP into the kernel. This is because 1) the ktcp code is known 16-bit working, which very few other TCP's are, and 2) moving the code over with as few changes as possible would greatly simplify getting/keeping things working, versus the high workload required for getting an entirely new TCP/IP up and running from scratch (inside the kernel or not, for that matter).

Good points! Very far in the back of my head I had a kernel TCP implementation you mentioned many years ago, created by a somewhat know person in the community whose name I didn't pay much attention to at the time (may have sounded something like David Kern or Kersch or probably something entirely different). Then there is of course the BSD2.11 TCP implementation which is basically 16 bit but bloated by our standards. Anyway I haven't even started to look, and your point may be the best shot: Don't.

> 
> > But this is a long shot. A shorter one would be to handle ICMP and ARP in the kernel, which would ease the system load
> 
> That could be a good idea to start with the only the low level - NIC drivers / network layers 1-3 into the kernel first, then followed by layer 4 and up. We could discuss more in a new issue if you get serious about it.

Yes, let's do that! That will be interesting regardless. 

BTW: 3 out of 5 NIC drives are now fully buffered. That includes the Lance driver - which was always buffered (no NIC buffers) and can use buffers in the heap, main memory or xms. That's where we're heading with `ne` and `wd` too 'real soon now'. And eventually 3com and Intel.

